PR #77, opened 2019-09-20
Optimization example script

blakecaldwell:
Adapt dipole optimization from HNN GUI to script form with hnn-core. Only runs simulations on a single core. Ran into errors with multiple trials, so num_trials=1 in the script, but once that is fixed, the number of trials can be configured at will. No other changes are necessary in this code.

Further details on optimization and weighted RMSE measure:
https://www.biorxiv.org/content/10.1101/740597v1

Tests remain for functions added to dipole.py


blakecaldwell:
I’m open to ideas for better places to store some of the functions in the script. Eventually, I think the functions dealing with parameters (and calculating the weight function for RMSE) should go into params.py. However, the run_optimization function that imports nlopt doesn’t clearly fit within hnn-core. It’s essentially a utility function that would be shared between the command-line interface and the HNN GUI.

Also the globals in optrun, unfortunately, are necessary because to my knowledge the interface is fixed by nlopt.

examples/optimize_evoked.py
```diff
@@ -0,0 +1,433 @@
...
+    opt_results: List
+        List with the values of the best parameters found by
+        optimization engine. Order is consistent with
+        opt_params['ranges']
+    """
+    import nlopt
+    global opt_params
+
+    nlopt.srand(seed)
+    algorithm = nlopt.LN_COBYLA
```
jasmainak: could we use scipy instead: https://docs.scipy.org/doc/scipy/reference/optimize.minimize-cobyla.html ?

Does nlopt give something that scipy does not?

examples/optimize_evoked.py
```diff
@@ -0,0 +1,433 @@
...
+    Returns
+    -------
+    avg_rmse: float
+        Weighted RMSE between data in dpl and exp_dpl
+    """
+    global opt_params, params, exp_dpl, best_dpl, cur_step, optiter, stepminopterr
+    print("Optimization step %d, iteration %d"%(cur_step+1, optiter+1))
+
+    for param_name, test_value in zip(opt_params['ranges'].keys(), new_params): # set parameters
+        if test_value >= opt_params['ranges'][param_name]['minval'] and test_value <= opt_params['ranges'][param_name]['maxval']:
```
jasmainak: @blakecaldwell would you mind running a flake8 linter on the code? It makes it a bit easier to read the code

blakecaldwell: Yeah, no problem for the future. I have limited time to finish this request that you made before the MNE sprint. Figured you'd still be able to do something with this code as-is.

examples/optimize_evoked.py
blakecaldwell: Oh, interesting! Feel free to change out for scipy. NLopt implemented PRAXIS for comparison and had better user documentation (scipy docs are complete but somewhat pedantic). Now that I've settled on using COBYLA, scipy seems preferable. I might make the change when the HNN optimization code gets rewritten.

blakecaldwell: I looked into the change. Would take a bit more time to define constraint functions. Also not clear if the results are reproducible like nlopt (specifying seed).

jasmainak: You can set the seed by providing `x0`

blakecaldwell:
Once #83 has been merged, I will rebase this and make some updates, including using `scipy` instead of `nlopt`

jasmainak:
okay I just rebased this branch. I'm going to try to squeeze some time to work on this in the coming days if someone wants to team up.

Here's the old branch: https://github.com/jasmainak/hnn-core/pull/new/optim_old

in case something went wrong, we can use that to revert to it

blakecaldwell:
@jasmainak happy to help with reviews. A thought is that this can be broken into pieces with RMSE first, and HNN GUI can use that immediately. It might be a good time to improve the original choices for trimming dipole at start/stop, which will impact the RMSE as reported in the GUI. I can help with tests for the weighted RMSE.

jasmainak:
@blakecaldwell I updated this script to use the latest API but the MPIBackend doesn't quite work for me. I get this:

```py
In [1]: %run examples/optimize_evoked.py
Running simulation with initial parameters
MPI will run over 2 processes
Running 1 trials...
/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/parallel_backends.py:180: UserWarning: Received BrokenPipeError exception. Child process failed unexpectedly
  warn("Received BrokenPipeError exception. "
[proxy:0:0@meghnn.nmr.mgh.harvard.edu] HYD_pmcd_pmip_control_cmd_cb (./pm/pmiserv/pmip_cb.c:939): process reading stdin too slowly; can't keep up
[proxy:0:0@meghnn.nmr.mgh.harvard.edu] HYDT_dmxu_poll_wait_for_event (./tools/demux/demux_poll.c:77): callback returned error status
[proxy:0:0@meghnn.nmr.mgh.harvard.edu] main (./pm/pmiserv/pmip.c:206): demux engine error waiting for event
/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/parallel_backends.py:168: UserWarning: Child process failed unexpectedly
  warn("Child process failed unexpectedly")
RuntimeError                              Traceback (most recent call last)
/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/examples/optimize_evoked.py in <module>
    403 net = Network(params, add_drives_from_params=True)
    404 with MPIBackend(n_procs=2):
--> 405     dpls = simulate_dipole(net, n_trials=1)
    406 # initial_dpl = average_dipoles(dpls)
    407 initial_dpl = dpls.copy()

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/dipole.py in simulate_dipole(net, n_trials, record_vsoma, record_isoma, postproc)
    108                         % type(record_isoma).__name__)
    109 
--> 110     dpls = _BACKEND.simulate(net, n_trials, postproc)
    111 
    112     return dpls

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/parallel_backends.py in simulate(self, net, n_trials, postproc)
    681             command=self.mpi_cmd, obj=net, timeout=30,
    682             proc_queue=self.proc_queue, env=env, cwd=os.getcwd(),
--> 683             universal_newlines=True)
    684 
    685         dpls = _gather_trial_data(sim_data, net, n_trials, postproc)

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/parallel_backends.py in run_subprocess(command, obj, timeout, proc_queue, *args, **kwargs)
    233         # simulation failed with a numeric return code
    234         raise RuntimeError("MPI simulation failed. Return code: %d" %
--> 235                            proc.returncode)
    236 
    237     child_data = _process_child_data(proc_data_bytes, data_len)

RuntimeError: MPI simulation failed. Return code: -13
```

any clue what's going on?

jasmainak:
I'm on CentOS 7

blakecaldwell:
It looks like the `mpich` MPI distribution is running. We have done all of our testing with `openmpi`. So while I think `MPIBackend` should be able to work regardless of the MPI distribution, can you verify which is being used, and if needed, switch to `openmpi`?

jasmainak:
alright, thanks for the help @blakecaldwell ! I am now able to make the example run locally in 5 minutes with `n_procs = 10` and an installation of OpenMPI that I compiled. Now I can keep my sanity while refactoring and debugging :)

examples/optimize_evoked.py
```diff
@@ -0,0 +1,85 @@
...
+net = Network(params, add_drives_from_params=True)
+with MPIBackend(n_procs=n_procs, mpi_cmd=mpi_cmd):
+    initial_dpl = simulate_dipole(net, n_trials=1)
+
+###############################################################################
+# Start the optimization!
+
+from hnn_core.optimization import optimize_evoked
+
+# From HNN, will take hours without parallelism
```
ntolley, line 58:
```suggestion
# Optimizing HNN simulations without parallelism may take several hours
```

jasmainak: nuked this sentence for now :)

jasmainak:
@rythorpe do you want to fetch this PR and give it a go in one of your existing scripts and see how you like the API? I am still working on improving the internal code though ...

hnn_core/dipole.py
```diff
@@ -172,6 +172,73 @@ def average_dipoles(dpls):
...
+        # normal RMSE
+        weights = np.ones(len(sim_times[0:sim_end_index]))
+    weights = weights[sim_start_index:sim_end_index]
+
+    dpl1 = dpl.data['agg'][sim_start_index:sim_end_index]
+    dpl2 = exp_dpl.data['agg'][exp_start_index:exp_end_index]
+
+    if (sim_length > exp_length):
+        # downsample simulation timeseries to match exp data
+        dpl1 = signal.resample(dpl1, exp_length)
```
rythorpe, line 231: For data time courses that end far from zero, I've usually had better luck with `scipy.signal.resample_poly()` to reduce edge effects. For instance, 

```python
dpl1 = signal.resample_poly(dpl1, exp_length, sim_length)
```
might be a better option.

hnn_core/optimization.py
```diff
@@ -0,0 +1,427 @@
...
+        The maximum number of of simulations to run for optimizing
+        one "chunk".
+    timing_range_multiplier : float
+        The scale of timing values to sweep over.
+    sigma_range_multiplier : float
+        The scale of sigma values to sweep over.
+    synweight_range_multiplier : float
+        The scale of input synaptic weights to sweep over.
+    decay_multiplier : float
+        The decay multiplier.
```
rythorpe, line 327: What is the `decay_multiplier` referring to here?

rythorpe:
This is great @jasmainak and @blakecaldwell! I just gave `optimize_evoked.py` a run and it worked well.

examples/optimize_evoked.py
```diff
@@ -0,0 +1,79 @@
...
+###############################################################################
+# Let us import hnn_core
+
+import hnn_core
+from hnn_core.dipole import simulate_dipole
+from hnn_core import read_params, Dipole, MPIBackend, Network
+
+hnn_core_root = op.join(op.dirname(hnn_core.__file__))
+
+mpi_cmd = '/autofs/space/meghnn_001/users/mjas/opt/openmpi/bin/mpirun'
```
rythorpe, line 27: Is there a way to automatically set this?

jasmainak: this is just set like this for my environment because I compiled it on the server. Couldn't figure out how to get it to find my compiled version automatically ... I'm guessing it works for you without having to set this?

hnn_core/optimization.py
blakecaldwell: It's part of the calculation for the time windows to optimize over. I set it empirically so it should be an optional parameter if the user knows better.

examples/optimize_evoked.py
rythorpe: Oh okay. Yes, I can just enter `mpiexec` or `mpirun` and everything works fine.

jasmainak:
okay I am done here. Some things were left were future PRs:

- The function could take `net` as input so it uses the drives from the API as input
- Tests could be improved

But I'd rather let people complain and ask for features as they use the functions rather than trying to bake in everything right now and delay folks to start using it. That way we can really move on from using legacy GUI

cc @rythorpe @ntolley @cjayb reviews are welcome! (hopefully not too much work ...)

examples/optimize_evoked.py
```diff
@@ -0,0 +1,81 @@
...
+                 extdata[:, 1], extdata[:, 1]])
+
+###############################################################################
+# Read the base parameters from a file
+params_fname = op.join(hnn_core_root, 'param', 'default.json')
+params = read_params(params_fname)
+
+###############################################################################
+# Let's first simulate the dipole with the initial parameters.
+net = Network(params, add_drives_from_params=True)
```
ntolley, line 51:
```suggestion
net = jones_2009_model(params, add_drives_from_params=True)
```
Is this the reason the example is failing after rebasing?

examples/optimize_evoked.py
```diff
@@ -0,0 +1,81 @@
...
+
+###############################################################################
+# Let us import hnn_core
+
+import hnn_core
+from hnn_core.dipole import simulate_dipole
+from hnn_core import read_params, Dipole, MPIBackend, Network
+
+hnn_core_root = op.join(op.dirname(hnn_core.__file__))
+
```
ntolley, line 26:
```suggestion
# The number of cores may need modifying depending on your current machine.
```

examples/optimize_evoked.py
```diff
@@ -0,0 +1,81 @@
...
+with MPIBackend(n_procs=n_procs):
+    initial_dpl = simulate_dipole(net, n_trials=1)
+
+###############################################################################
+# Start the optimization!
+
+from hnn_core.optimization import optimize_evoked
+
+with MPIBackend(n_procs=n_procs):
+    params_optim = optimize_evoked(params, exp_dpl)
```
ntolley, line 61: In the spirit of moving away from working with the `params` dictionary correctly, perhaps you could pass a copy of the `Network` object that gets modified in place?

examples/optimize_evoked.py
```diff
@@ -0,0 +1,81 @@
...
+
+###############################################################################
+# Now, let's simulate the dipole with the optimized parameters.
+net = Network(params, add_drives_from_params=True)
+with MPIBackend(n_procs=n_procs):
+    best_dpl = simulate_dipole(net, n_trials=1)
+
+###############################################################################
+# Now plot the results against experimental data:
```
ntolley, line 70:
```suggestion
# Finally, we can plot the results against experimental data:
```

examples/optimize_evoked.py
```diff
@@ -0,0 +1,81 @@
...
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+###############################################################################
+# Let us import hnn_core
+
+import hnn_core
+from hnn_core.dipole import simulate_dipole
+from hnn_core import read_params, Dipole, MPIBackend, Network
```
ntolley, line 23:
```suggestion
from hnn_core import read_params, Dipole, MPIBackend, jones_2009_model
```

hnn_core/optimization.py
```diff
@@ -0,0 +1,427 @@
...
+
+import numpy as np
+import scipy.stats as stats
+from scipy.optimize import fmin_cobyla
+
+from .network import Network
+from .dipole import _rmse
+
+
+def _get_range(val, multiplier):
```
ntolley, line 17: oneliner docstring?

hnn_core/optimization.py
```diff
@@ -0,0 +1,427 @@
...
+    range_min = max(0, val - val * multiplier / 100.)
```
ntolley, line 18: where does the 100 come from?

hnn_core/optimization.py
```diff
@@ -0,0 +1,427 @@
...
+    Parameters
+    ----------
+    new_params: Array
+        List or numpy array with the parameters chosen by
+        optimization engine. Order is consistent with
+        opt_params['ranges'].
+    opt_params : dict
+        The optimization parameters
+    params : dict
+        The params dictionary.
```
ntolley, line 243:
```suggestion
    params : an instance of Params
        Params containing paramter values from file
```
I think this applies to a couple more places (docstring pulled from `params.py`)

doc/whats_new.rst
```diff
@@ -36,6 +36,9 @@ Changelog
 
 - Add ``calcium_model`` with a distance dependent calcium channel conductivity, by `Nick Tolley`_ in `#348 <https://github.com/jonescompneurolab/hnn-core/pull/333>`_
 
+- Add :func:`~hnn_core.optimize_evoked` to optimize the timing and weights of driving inputs for simulating
+  evoked responses, by `Blake Caldwell`_ and `Mainak Jas`_ in `#77 <https://github.com/jonescompneurolab/hnn-core/pull/333>`_
```
rythorpe, line 40:
```suggestion
  evoked responses, by `Blake Caldwell`_ and `Mainak Jas`_ in `#77 <https://github.com/jonescompneurolab/hnn-core/pull/77>`_
```

rythorpe:
Should 'optimize_evoked.py' go in 'hnn-core/examples/howto'?

examples/optimize_evoked.py
ntolley: Maybe if no connections, throw error/warning

examples/optimize_evoked.py
rythorpe: This is tricky because if you're going to optimized a bunch of parameters, it's nice to be able to find them all in one place in order to see what changed.

examples/optimize_evoked.py
jasmainak: ugh ... didn't seem to have fixed it :(

jasmainak: Note to self: I wonder if there is some mismatch in the units of dipole that is read in vs the simulate dipole

hnn_core/optimization.py
```diff
@@ -0,0 +1,427 @@
...
+    params: an instance of Params
+        Full set of simulation parameters
+
+    Returns
+    -------
+    sorted_evinput_params: dict
+        Dictionary with parameters grouped by evoked inputs.
+        Keys for each evoked input are
+        'mean', 'sigma', 'ranges', 'start',
+        'end'.
```
ntolley, line 39:
```suggestion
        'mean', 'sigma', 'ranges', 'start', and 'end'.
```

hnn_core/optimization.py
```diff
@@ -0,0 +1,427 @@
...
+
+def _split_by_evinput(params, sigma_range_multiplier, timing_range_multiplier,
+                      synweight_range_multiplier):
+    """ Sorts parameter ranges by evoked inputs into a dictionary
+
+    Parameters
+    ----------
+    params: an instance of Params
+        Full set of simulation parameters
+
```
ntolley, line 32: Missing docstring for multipliers

hnn_core/optimization.py
```diff
@@ -0,0 +1,427 @@
...
+    new_params: Array
+        List or numpy array with the parameters chosen by
+        optimization engine. Order is consistent with
+        opt_params['ranges'].
+    opt_params : dict
+        The optimization parameters
+    params : dict
+        The params dictionary.
+    opt_dpls : dict
+        Dictionary with keys exp_dpl and best for
```
ntolley, line 245:
```suggestion
        Dictionary with keys 'exp_dpl' and 'best' for
```

hnn_core/optimization.py
```diff
@@ -0,0 +1,427 @@
...
+    synweight_range_multiplier : float
+        The scale of input synaptic weights to sweep over.
+    decay_multiplier : float
+        The decay multiplier.
+
+    Returns
+    -------
+    params : dict
+        The optimized params dictionary.
+    """
```
ntolley, line 333:
```suggestion
    params : dict
        The optimized params dictionary.
    
    Notes
    -----
    This optimization protocol utilizes the Constrained Optimization
    By Linear Approximation (COBYLA) method:
    https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.fmin_cobyla.html  #noqa
    """
```

ntolley:
@jasmainak your offhand comment about the scaling possible being messed up was spot on. Checkout these plots from `exp_dpl` vs. `initial_dpl` respectively:
image
image

The y-axis is orders of magnitude off, just to confirm this was not the case before the rebase?

examples/optimize_evoked.py
```diff
@@ -0,0 +1,82 @@
...
+###############################################################################
+# Read the base parameters from a file
+params_fname = op.join(hnn_core_root, 'param', 'default.json')
+params = read_params(params_fname)
+
+###############################################################################
+# Let's first simulate the dipole with the initial parameters.
+net = jones_2009_model(params, add_drives_from_params=True)
+with MPIBackend(n_procs=n_procs):
+    initial_dpl = simulate_dipole(net, n_trials=1)
```
ntolley, line 54:
```suggestion
    initial_dpl = simulate_dipole(net, n_trials=1, postproc=True)
```
Sneaky bugger :wink

ntolley:
#372 is most likely the culprit. However, this introduces a good question, how should scaling be introduced into the optimization? Perhaps a `scale` argument to `optimize_evoked`? On second thought we probably shouldn't reintroduce `postproc` as a hacky workaround.

examples/optimize_evoked.py
ntolley: Don't actually commit this, I'm just pointing out a line that can fix the initial comparison. Scaling still needs to be fixed inside `optimization.py1

jasmainak: indeed, this is what I was suspecting. Let me give this a shot. Need to get it to work first ... after that we can think of the optimal API :)

jasmainak:
@ntolley I can confirm that setting `postproc=True` does the job. Yayyyy!

So, now what's the right solution? Is it enough to have `scaling_factor` as an option? What about smoothing? I'm guessing the smoothing acts like a regularization and is necessary for getting reasonable weights

examples/optimize_evoked.py
jasmainak: Even though I agree in principle with @ntolley I'd rather defer this to a future PR

hnn_core/optimization.py
jasmainak: honestly I don't know ... maybe @blakecaldwell can shed some light

hnn_core/optimization.py
jasmainak: Is this relevant for the user to know? You could replace Cobyla with any other solver and it should still work ...

ntolley: I agree that it should still work, but considering the ability to change the solver is not exposed to the user I think it's important to document what is actually doing the optimizing in this function.

ntolley:
Instead of having scale be an argument, is it possible to set it as a bool such that the scaling factor is one of the parameters that is optimized? When `optimize_scale=False` the user can rely on `Dipole.scale()` to get their waveform into a reasonable range.

As for smoothing, I suppose the same logic could apply?

hnn_core/optimization.py
blakecaldwell: multiplier is a percentage, probably borrowed from when a value was read from the GUI.

ntolley: Thanks Blake! That makes sense, pretty easy to update in the docstring

jasmainak:
I don't think we should be optimizing `scale` and/or smoothing because that was not part of the original algorithm in the paper. I'd like to get the GUI functionality first and then any enhancements can be added after validation/published papers on the topic.

hnn_core/optimization.py
```diff
@@ -0,0 +1,439 @@
...
+            lambda x: x[idx] <= param_ranges[param_name]['maxval'])
+        cons.append(
+            lambda x: x[idx] >= param_ranges[param_name]['minval'])
+
+    result = fmin_cobyla(optrun, cons=cons, rhoend=1e-4,
+                         x0=x0, maxfun=maxiter)
+    return result
+
+
+def optimize_evoked(params, exp_dpl, maxiter=50,
```
jasmainak, line 312: note for self, add `initial_dpl` as a parameter

jasmainak:
@rythorpe do you mind trying to run the example? I hit an error that's related to your last PR (https://github.com/jonescompneurolab/hnn-core/commit/abc7ee8e2716e66de3a1d337801f7c03f88004c8) and am not sure how to fix it. I get:

```py
ValueError                                Traceback (most recent call last)
/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/examples/optimize_evoked.py in <module>
     69 with MPIBackend(n_procs=n_procs):
     70     params_optim = optimize_evoked(params, exp_dpl, scale_factor=scale_factor,
---> 71                                    smooth_window_len=smooth_window_len)
     72 
     73 ###############################################################################

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/optimization.py in optimize_evoked(params, exp_dpl, maxiter, timing_range_multiplier, sigma_range_multiplier, synweight_range_multiplier, decay_multiplier, scale_factor, smooth_window_len)
    457         opt_results = _run_optimization(maxiter=maxiter,
    458                                         param_ranges=opt_params['ranges'],
--> 459                                         optrun=_myoptrun)
    460 
    461         # update opt_params for the next round

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/optimization.py in _run_optimization(maxiter, param_ranges, optrun)
    319 
    320     result = fmin_cobyla(optrun, cons=cons, rhoend=1e-4,
--> 321                          x0=x0, maxfun=maxiter)
    322     return result
    323 

/autofs/space/meghnn_001/users/mjas/anaconda3/envs/hnn_pip/lib/python3.7/site-packages/scipy/optimize/cobyla.py in wrapper(*args, **kwargs)
     32     def wrapper(*args, **kwargs):
     33         with _module_lock:
---> 34             return func(*args, **kwargs)
     35     return wrapper
     36 

/autofs/space/meghnn_001/users/mjas/anaconda3/envs/hnn_pip/lib/python3.7/site-packages/scipy/optimize/cobyla.py in fmin_cobyla(func, x0, cons, args, consargs, rhobeg, rhoend, maxfun, disp, catol)
    175 
    176     sol = _minimize_cobyla(func, x0, args, constraints=con,
--> 177                            **opts)
    178     if disp and not sol['success']:
    179         print("COBYLA failed to find a solution: %s" % (sol.message,))

/autofs/space/meghnn_001/users/mjas/anaconda3/envs/hnn_pip/lib/python3.7/site-packages/scipy/optimize/cobyla.py in wrapper(*args, **kwargs)
     32     def wrapper(*args, **kwargs):
     33         with _module_lock:
---> 34             return func(*args, **kwargs)
     35     return wrapper
     36 

/autofs/space/meghnn_001/users/mjas/anaconda3/envs/hnn_pip/lib/python3.7/site-packages/scipy/optimize/cobyla.py in _minimize_cobyla(fun, x0, args, constraints, rhobeg, tol, maxiter, disp, catol, **unknown_options)
    260     xopt, info = _cobyla.minimize(calcfc, m=m, x=np.copy(x0), rhobeg=rhobeg,
    261                                   rhoend=rhoend, iprint=iprint, maxfun=maxfun,
--> 262                                   dinfo=info)
    263 
    264     if info[3] > catol:

/autofs/space/meghnn_001/users/mjas/anaconda3/envs/hnn_pip/lib/python3.7/site-packages/scipy/optimize/cobyla.py in calcfc(x, con)
    250 
    251     def calcfc(x, con):
--> 252         f = fun(x, *args)
    253         i = 0
    254         for size, c in izip(cons_lengths, constraints):

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/optimization.py in _myoptrun(new_params)
    451                            params, opt_dpls=opt_dpls,
    452                            scale_factor=scale_factor,
--> 453                            smooth_window_len=smooth_window_len)
    454 
    455         print('Optimizing from [%3.3f-%3.3f] ms' % (opt_params['opt_start'],

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/optimization.py in _optrun(new_params, opt_params, params, opt_dpls, scale_factor, smooth_window_len)
    276 
    277     # run the simulation, but stop early if possible
--> 278     net = jones_2009_model(params, add_drives_from_params=True)
    279     tstop = params['tstop'] = opt_params['opt_end']
    280     net._instantiate_drives(n_trials=1, tstop=tstop)

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/network_models.py in jones_2009_model(params, add_drives_from_params)
     40         params = read_params(params_fname)
     41 
---> 42     net = Network(params, add_drives_from_params=add_drives_from_params)
     43 
     44     delay = net.delay

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/network.py in __init__(self, params, add_drives_from_params, legacy_mode)
    390 
    391         if add_drives_from_params:
--> 392             _add_drives_from_params(self)
    393 
    394     def __repr__(self):

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/drives.py in _add_drives_from_params(net)
    131                 location=specs['location'], seedcore=specs['seedcore'],
    132                 synaptic_delays=specs['synaptic_delays'],
--> 133                 space_constant=specs['space_constant'])
    134         elif specs['type'] == 'poisson':
    135             net.add_poisson_drive(

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/network.py in add_evoked_drive(self, name, mu, sigma, numspikes, location, n_drive_cells, cell_specific, weights_ampa, weights_nmda, space_constant, synaptic_delays, seedcore)
    504         self._attach_drive(name, drive, weights_ampa, weights_nmda, location,
    505                            space_constant, synaptic_delays,
--> 506                            n_drive_cells, cell_specific)
    507 
    508     def add_poisson_drive(self, name, *, tstart=0, tstop=None, rate_constant,

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/network.py in _attach_drive(self, name, drive, weights_ampa, weights_nmda, location, space_constant, synaptic_delays, n_drive_cells, cell_specific)
    726         target_populations, weights_by_type, delays_by_type = \
    727             _get_target_properties(weights_ampa, weights_nmda, synaptic_delays,
--> 728                                    location)
    729 
    730         # weights passed must correspond to cells in the network

/autofs/space/meghnn_001/users/mjas/github_repos/hnn-core/hnn_core/drives.py in _get_target_properties(weights_ampa, weights_nmda, synaptic_delays, location)
     50 
     51     if set(delays_by_type.keys()) != target_populations:
---> 52         raise ValueError('synaptic_delays is either a common float or needs '
     53                          'to be specified as a dict for each of the cell '
     54                          'types defined in weights_ampa and weights_nmda '

ValueError: synaptic_delays is either a common float or needs to be specified as a dict for each of the cell types defined in weights_ampa and weights_nmda ({'L5_basket', 'L2_pyramidal', 'L2_basket'})

```

rythorpe:
@jasmainak just wanted to let you know that I'm working on it! Somehow `fmin_cobyla` is assigning a negative synaptic weight value which makes the drives API silently ignore certain values when `_add_drives_from_params()` is called. I'm still trying to figure out how/why `fmin_cobyla` is exploring parameters outside the bounds of our constraints.

jasmainak:
Ahhh ... I see, thanks for the explanation. I think `fmin_cobyla` can go outside the constraints a bit. There is the `catol` parameter for that. Maybe try changing that?

Other option is to use LBFGS: https://docs.scipy.org/doc/scipy/reference/optimize.minimize-lbfgsb.html

it will honor the bounds more strictly I believe

jasmainak:
> negative synaptic weight value which makes the drives API silently ignore certain values

shouldn't this be fixed? or is this intended behavior?

rythorpe:
> Ahhh ... I see, thanks for the explanation. I think `fmin_cobyla` can go outside the constraints a bit. There is the `catol` parameter for that. Maybe try changing that?

I tried that, but I think there is something wrong with the way we have setup our constraint functions. First of all, they need to output a float or int, not logical. `fmin_cobyla` then ensures that the output of each constraint never falls below zero (with a specified tolerance). This is an easy fix. What I can't figure out is why each all minimum-specifying lambda functions in the `cons` appear to be the same, and why all maximum-specifying lambda functions appear to be the same.

jasmainak:
Maybe I screwed up somewhere in the constraint functions. The nlopt implementation of COBYLA has mechanism to add bound constraints, it's a modified algorithm. But I'm pretty sure we can replace it with any of the other solvers (LBFGS etc) and get nearly identical results. Probably, instead of boolean the constraint functions should return "upper bound - x" and "x - lower bound"

rythorpe:
Do you have any lingering commits floating around @jasmainak or am I clear to push?

jasmainak:
you can push!

rythorpe:
The example should work now, however, we might want to consider using an alternative to `fmin_cobyla` as it still allows initial steps within parameter space to exceed the boundary constraints. Also, we should probably add a check somewhere that throws an error if/when a negative synaptic weight is entered, however, this can be addressed in a different PR.

jasmainak:
> Also, we should probably add a check somewhere that throws an error if/when a negative synaptic weight is entered, however, this can be addressed in a different PR.

would you mind taking the lead on this?

I will look into other solvers. Maybe we can even have an argument `'solver'` (like in scikit-learn) that allows switching between different options

PR #83, opened 2019-09-27
RMSE and dipole averaging

blakecaldwell:
New PR to be considered for merging in support of optimization in HNN

### DIFF
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -145,6 +145,101 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
     return dpl
 
 
+def average_dipoles(dpls):
+    """Compute average over a list of Dipole objects.
+    Parameters
+    ----------
+    dpls: list of Dipole objects
+        Contains list of dipole results to be averaged
+
+    Returns
+    -------
+    dpl: instance of Dipole
+        A dipole object with averages of the dipole data
+    """
+    # need at least on Dipole to get times
+    assert (len(dpls) > 0)
+
+    agg_avg = np.mean(np.array([dpl.dpl['agg'] for dpl in dpls]), axis=0)
+    L5_avg = np.mean(np.array([dpl.dpl['L5'] for dpl in dpls]), axis=0)
+    L2_avg = np.mean(np.array([dpl.dpl['L2'] for dpl in dpls]), axis=0)
+
+    avg_dpl_data = np.c_[agg_avg,
+                         L2_avg,
+                         L5_avg]
+
+    avg_dpl = Dipole(dpls[0].t, avg_dpl_data)
+
+    return avg_dpl
+
+
+def rmse(dpl, exp_dpl, tstart=0.0, tstop=0.0, weights=None):
+    """ Calculates RMSE between data in dpl and exp_dpl
+    Parameters
+    ----------
+    dpl: instance of Dipole
+        A dipole object with simulated data
+    exp_dpl: instance of Dipole
+        A dipole object with experimental data
+    tstart | None: float
+        Time at beginning of range over which to calculate RMSE
+    tstop | None: float
+        Time at end of range over which to calculate RMSE
+    weights | None: array
+        An array of weights to be applied to each point in
+        simulated dpl. Must have length >= dpl.dpl
+        If None, weights will be replaced with 1's for typical RMSE
+        calculation.
+
+    Returns
+    -------
+    err: float
+        Weighted RMSE between data in dpl and exp_dpl
+    """
+    from scipy import signal
+
+    exp_times = exp_dpl.t
+    sim_times = dpl.t
+
+    # do tstart and tstop fall within both datasets?
+    # if not, use the closest data point as the new tstop/tstart
+    for tseries in [exp_times, sim_times]:
+        if tstart < tseries[0]:
+            tstart = tseries[0]
+        if tstop > tseries[-1]:
+            tstop = tseries[-1]
+
+    # make sure start and end times are valid for both dipoles
+    exp_start_index = (np.abs(exp_times - tstart)).argmin()
+    exp_end_index = (np.abs(exp_times - tstop)).argmin()
+    exp_length = exp_end_index - exp_start_index
+
+    sim_start_index = (np.abs(sim_times - tstart)).argmin()
+    sim_end_index = (np.abs(sim_times - tstop)).argmin()
+    sim_length = sim_end_index - sim_start_index
+
+    if weights is None:
+        # weighted RMSE with weights of all 1's is equivalent to
+        # normal RMSE
+        weights = np.ones(len(sim_times[0:sim_end_index]))
+    weights = weights[sim_start_index:sim_end_index]
+
+    dpl1 = dpl.dpl['agg'][sim_start_index:sim_end_index]
+    dpl2 = exp_dpl.dpl['agg'][exp_start_index:exp_end_index]
+
+    if (sim_length > exp_length):
+        # downsample simulation timeseries to match exp data
+        dpl1 = signal.resample(dpl1, exp_length)
+        weights = signal.resample(weights, exp_length)
+        indices = np.where(weights < 1e-4)
+        weights[indices] = 0
+    elif (sim_length < exp_length):
+        # downsample exp timeseries to match simulation data
+        dpl2 = signal.resample(dpl2, sim_length)
+
+    return np.sqrt((weights * ((dpl1 - dpl2) ** 2)).sum() / weights.sum())
+
+
 class Dipole(object):
     """Dipole class.
 
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -256,14 +256,19 @@ class Dipole(object):
     t : array
         The time vector
     dpl : dict of array
-        The dipole with keys 'agg', 'L2' and 'L5'
+        The dipole with keys 'agg' 'L2' (optional) and 'L5' (optional)
     """
 
     def __init__(self, times, data):  # noqa: D102
         self.units = 'fAm'
         self.N = data.shape[0]
         self.t = times
-        self.dpl = {'agg': data[:, 0], 'L2': data[:, 1], 'L5': data[:, 2]}
+        self.dpl = {}
+
+        shp = data.shape
+        keys = ['agg', 'L2', 'L5']
+        for key, tseries in zip(keys, range(shp[1])):
+            self.dpl[key] = data[:, tseries]
 
     def convert_fAm_to_nAm(self):
         """ must be run after baseline_renormalization()
--- a/hnn_core/tests/test_rmse.py
+++ b/hnn_core/tests/test_rmse.py
@@ -1,0 +1,31 @@
+import os.path as op
+
+from numpy import loadtxt, c_
+
+from mne.utils import _fetch_file
+import hnn_core
+from hnn_core import simulate_dipole, average_dipoles, rmse
+from hnn_core import Params, Network, Dipole
+
+
+def test_rmse():
+    """Test to check RMSE calculation"""
+    data_url = ('https://raw.githubusercontent.com/jonescompneurolab/hnn/'
+                'master/data/MEG_detection_data/yes_trial_S1_ERP_all_avg.txt')
+    if not op.exists('yes_trial_S1_ERP_all_avg.txt'):
+        _fetch_file(data_url, 'yes_trial_S1_ERP_all_avg.txt')
+    extdata = loadtxt('yes_trial_S1_ERP_all_avg.txt')
+
+    exp_dpl = Dipole(extdata[:, 0], c_[extdata[:, 1]])
+
+    hnn_core_root = op.join(op.dirname(hnn_core.__file__), '..')
+    params_fname = op.join(hnn_core_root, 'param', 'default.json')
+    params = Params(params_fname)
+
+    net = Network(params)
+    dpls = simulate_dipole(net)
+    avg_dpl = average_dipoles(dpls)
+    avg_rmse = rmse(avg_dpl, exp_dpl, tstop=params['tstop'])
+    expected_rmse = 4.533252902006792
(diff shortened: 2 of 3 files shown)


hnn_core/dipole.py
```diff
@@ -145,6 +145,101 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
     return dpl
 
 
+def average_dipoles(dpls):
+    """Compute average over a list of Dipole objects.
+    Parameters
+    ----------
+    dpls: list of Dipole objects
+        Contains list of dipole results to be averaged
```
jasmainak: perhaps you can add that the average will be done layerwise?

hnn_core/dipole.py
```diff
@@ -145,6 +145,101 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
...
+
+    Returns
+    -------
+    dpl: instance of Dipole
+        A dipole object with averages of the dipole data
+    """
+    # need at least on Dipole to get times
+    assert (len(dpls) > 0)
```
jasmainak: can you raise a proper user friendly error?

hnn_core/dipole.py
```diff
@@ -145,6 +145,101 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
...
+    exp_times = exp_dpl.t
+    sim_times = dpl.t
+
+    # do tstart and tstop fall within both datasets?
+    # if not, use the closest data point as the new tstop/tstart
+    for tseries in [exp_times, sim_times]:
+        if tstart < tseries[0]:
+            tstart = tseries[0]
+        if tstop > tseries[-1]:
+            tstop = tseries[-1]
```
jasmainak: This is dangerous, you should raise an error here

hnn_core/dipole.py
```diff
@@ -145,6 +145,101 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
     return dpl
 
 
+def average_dipoles(dpls):
```
jasmainak: I am wondering if we should have an `__add__` method in dipoles. That can be used for this and also for the rmse.

hnn_core/tests/test_rmse.py
```diff
@@ -0,0 +1,31 @@
...
+    params_fname = op.join(hnn_core_root, 'param', 'default.json')
+    params = Params(params_fname)
+
+    net = Network(params)
+    dpls = simulate_dipole(net)
+    avg_dpl = average_dipoles(dpls)
+    avg_rmse = rmse(avg_dpl, exp_dpl, tstop=params['tstop'])
+    expected_rmse = 4.533252902006792
+
+    assert(round(avg_rmse, 12) == round(expected_rmse, 12))
```
jasmainak: use `assert_allclose`: https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_allclose.html

hnn_core/dipole.py
blakecaldwell: Yeah, it could be surprising behavior, but it's not an error to HNN. If a user adds an arbitrary experimental dipole file, that's fine, but RMSE should only be calculated over the common region. I think warnings.warn might be a better way to communicate to the user.

hnn_core/dipole.py
blakecaldwell: Maybe later

hnn_core/dipole.py
jasmainak: can you describe a use case where it's actually helpful?

hnn_core/dipole.py
```diff
@@ -145,6 +145,101 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
...
+            tstop = tseries[-1]
+
+    # make sure start and end times are valid for both dipoles
+    exp_start_index = (np.abs(exp_times - tstart)).argmin()
+    exp_end_index = (np.abs(exp_times - tstop)).argmin()
+    exp_length = exp_end_index - exp_start_index
+
+    sim_start_index = (np.abs(sim_times - tstart)).argmin()
+    sim_end_index = (np.abs(sim_times - tstop)).argmin()
+    sim_length = sim_end_index - sim_start_index
```
blakecaldwell: These lines define the common region.

hnn_core/dipole.py
blakecaldwell: I had a different example in mind above. Let me start over. Say the user loads in a short experimental data set. Simulation tstop is way beyond the end of the exp data. So this updates tstop be the last data point. The closest data point is still identified below. Once the region of interest has been changed to the smallest common region, then downsampling is performed.

blakecaldwell: Credit to @rythorpe for identifying this case

jasmainak: Then you should crop to tstop and then compute rmse.

```py
exp_dipole.crop(tmax=5.0)
rmse(exp_dipole, dipole)
```

hnn_core/dipole.py
```diff
@@ -145,6 +145,101 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
...
+        # weighted RMSE with weights of all 1's is equivalent to
+        # normal RMSE
+        weights = np.ones(len(sim_times[0:sim_end_index]))
+    weights = weights[sim_start_index:sim_end_index]
+
+    dpl1 = dpl.dpl['agg'][sim_start_index:sim_end_index]
+    dpl2 = exp_dpl.dpl['agg'][exp_start_index:exp_end_index]
+
+    if (sim_length > exp_length):
+        # downsample simulation timeseries to match exp data
```
jasmainak: even this is dangerous. Why not simulate with the correct `sfreq` in the first place?

rythorpe: @jasmainak We usually simulate at time steps of around 0.01 ms to maintain stability during numerical integration of the HH model. Since 100,000 Hz is very impractical for experimental measurements, we're forced to downsample the simulation time series. What is dangerous about it?

jasmainak: If you don't have force the two signals to have equal number of samples, the differences in rmse could be because of: a) Different sampling frequency, b) Different length of signal. Now because all this is happening within one function, you will have no control over it. To me the `rmse` function should only accept signals of equal number of samples. Then if you want to crop or resample, there should be different methods ``dipole.crop`` and ``dipole.resample`` for it. And the time points are stored in ``dipole.times``. See how we do it in MNE: https://mne.tools/dev/generated/mne.Evoked.html#mne.Evoked.

Also, it's weird that you resample such a short signal, wouldn't that introduce ringing artifacts?

hnn_core/tests/test_rmse.py
```diff
@@ -0,0 +1,31 @@
+import os.path as op
+
+from numpy import loadtxt, c_
```
jasmainak:
```suggestion
import numpy as np
```

that's the canonical way to import `numpy`

hnn_core/dipole.py
blakecaldwell: I'd like to pick this back up with a rebase. Though, there are some discussions to follow up on.

> If you don't have force the two signals to have equal number of samples, the differences in rmse could be because of: a) Different sampling frequency, b) Different length of signal. Now because all this is happening within one function, you will have no control over it. To me the `rmse` function should only accept signals of equal number of samples. Then if you want to crop or resample, there should be different methods `dipole.crop` and `dipole.resample` for it. And the time points are stored in `dipole.times`. See how we do it in MNE: https://mne.tools/dev/generated/mne.Evoked.html#mne.Evoked.
> 
> Also, it's weird that you resample such a short signal, wouldn't that introduce ringing artifacts?

This is helpful advice. I think in most cases the simulation is going to get downsampled, but it's at a relatively high frequency, so why do you call it a `short signal`. Do you agree that downsampling is still necessary? I think the goal is how to get two signal of the same number of samples while introducing minimal noise.

hnn_core/dipole.py
blakecaldwell: @jasmainak can you describe whay would happen in `__add__`? would RMSE get calculated automatically?

jasmainak: ummm ... I am not sure any more what I meant, I think it's fine to resolve this comment.

hnn_core/dipole.py
```diff
@@ -145,6 +145,101 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
     return dpl
 
 
+def average_dipoles(dpls):
+    """Compute average over a list of Dipole objects.
```
jasmainak, line 149: need a blank line after this

hnn_core/dipole.py
jasmainak: I need to ponder over this whole procedure, get back to you in a couple of hours :)

jasmainak:
@blakecaldwell @rythorpe 

here is an example to demo the problem:

```py
import numpy as np

from scipy import signal

sfreq = 100000

t = np.linspace(0, 1, sfreq, endpoint=False)
y = signal.square(2 * np.pi * 2 * t)

sfreq_new = 100
y_resampled = signal.resample(y, sfreq_new)
t_resampled = np.linspace(0, 1, sfreq_new, endpoint=False)

plt.plot(t_resampled, y_resampled, 'r')
plt.plot(t, y)

plt.ylim(-2, 2)
plt.show()
```

which gives

Screen Shot 2020-08-06 at 9 07 59 PM

now it really depends what you want to do with the `rmse` function. Do you want to make it a public function or just use it inside the optimization? If you want to make it a public function, I would suggest putting in appropriate checks. E.g., add a flag `resample` which is `False` by default but can be set to `True` when used from the optimization functions. We can work on the API, that's not an issue. Just wanted to share this first

rythorpe:
> now it really depends what you want to do with the `rmse` function. Do you want to make it a public function or just use it inside the optimization?

I think the `rmse` function should be public because it is a useful tool for conducting manual model parameter tuning.

> If you want to make it a public function, I would suggest putting in appropriate checks. E.g., add a flag `resample` which is `False` by default but can be set to `True` when used from the optimization functions. We can work on the API, that's not an issue. Just wanted to share this first

The most common case is that the simulation time-series will need to be resampled, so I think default behavior should be true (with appropriate checks of course). That said, it might be worth separating the resampling function from the optimization function in order to clearly contain and test for potential errors.

rythorpe:
Also, I've had better luck resampling the types of signals we use (i.e., EEG inverse solutions) using `scipy.signal.resample_poly()`. Anecdotally, I've found it to do a much better job at limiting ringing due to edge-effects. For example, here's a snippet of code I wrote for my own analysis.

```python
def _resampled_rmse(sig, ref, plot_resamp_sig=False):
    '''Compute the root mean squared error (RMSE)
    between a signal and its reference following
    signal resampling. First, the signal is resampled
    to match the timecourse of the reference; then,
    RMSE is computed.

    Parameters
    ----------
    sig : array (2D)
        2 x n array of time points and samples
        where n is the number of samples.
    ref : array (2D)
        2 x m array of time points and samples
        where m is the number of samples.
    plot_resamp_sig : boolean
        If True, plot a comparison of the original
        and resampled signal for sanity check.

    Returns
    -------
    rmse : float
        Root mean squared error
    '''

    sig_length = len(sig[0])
    ref_length = len(ref[0])
    sig_resamp = resample_poly(sig[1], ref_length-1, sig_length-1)
    sig_resamp[0] = sig[1][0] # Control edge effects
    sig_resamp[-1] = sig[1][-1] # Control edge effects
    dt_new = (sig[0][1]-sig[0][0])*(sig_length-1)/(ref_length-1)
    t_resamp = np.arange(sig[0][0], sig[0][-1]+dt_new, dt_new)
    rmse = np.sqrt(np.mean((sig_resamp - ref[1])**2))

    if plot_resamp_sig==True:
        plt.figure()
        plt.plot(sig[0], sig[1], label='original signal')
        plt.scatter(t_resamp, sig_resamp, marker='x', label='resampled signal')
        plt.scatter(ref[0], ref[1], marker='.', label='reference')
        plt.xlabel('time')
        plt.legend()
        plt.show()

    return rmse
```

blakecaldwell:
@jasmainak very helpful example to think about this a little bit. Does upsampling the experimental data come with its own set of problems? The reason I like this is because it would be a one-time transformation to be used in RMSE calculation for all simulated dipoles. Any introduced artifacts will be present for all comparisons. So as a relative measure RMSE is still valid.

Thanks @rythorpe, I'm happy to implement your suggested method as an alternative and compare the results.

jasmainak:
@blakecaldwell don't think there would be issues related to ringing etc. But you do need to think carefully what the upsampling means. It reminds me of these multi-resolution approaches in image processing. If I have developed a new segmentation algorithm and I downsample the image, it could be that with sufficient downsampling I could make it appear that my algorithm is just as good as the other ones. If you ran the algorithm at full resolution, the competing algorithms might be more accurate but the differences disappear when you do a downsample-run algorithms-upsample procedure.

regarding the ringing, it depends where you use it in the optimization algorithm. Do you use it as a cost function? are tstart and tstop somehow dynamically chosen? Because what if the time span across which rmse is computed becomes really small? I don't remember the optimization algorithm very well (sorry!) but just some things to keep in mind and also put in adequate checks in the software to avoid users doing the wrong thing.

hnn_core/dipole.py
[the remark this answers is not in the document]
blakecaldwell: Okay, but this might be a bigger issue since `n_trials` would need to be saved with the dipole object to prevent such an error. Right now `Dipole.write()` doesn't do that. Is that what you're suggesting?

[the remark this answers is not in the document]
jasmainak: we could save it too actually. `np.loadtxt` for e.g., allows you to skip few rows in the beginning. You can save this information there. I have seen this kind of thing in other ephys stuff, I think [when saving MNE annotations](https://github.com/mne-tools/mne-python/blob/maint/0.20/mne/annotations.py#L574) ...

blakecaldwell:
Abandoning this

PR #96, opened 2020-03-27
Fix dipole and network (spiking) class write methods

rythorpe:
Two changes have been made: 1) `dipole.write()` has been fixed to save files according to the user-specified name, and 2) `network.write_spikes()` has been added so that spiking data for selected trials can be saved to a single .txt file. See the attached example (a modified version of the plot_simulate_evoked.py example) and its corresponding output .txt and .png files that demonstrates reconstruction of the simulated spike event plot using the spike .txt files.
[demo.zip](https://github.com/jonescompneurolab/hnn-core/files/4391152/demo.zip)

### DIFF
--- a/examples/plot_simulate_evoked.py
+++ b/examples/plot_simulate_evoked.py
@@ -16,7 +16,7 @@ import os.path as op
 # Let us import hnn_core
 
 import hnn_core
-from hnn_core import simulate_dipole, read_params, Network
+from hnn_core import simulate_dipole, read_params, Network, read_spikes
 
 hnn_core_root = op.join(op.dirname(hnn_core.__file__), '..')
 
@@ -42,12 +42,18 @@ dpls = simulate_dipole(net, n_jobs=1, n_trials=2)
 import matplotlib.pyplot as plt
 fig, axes = plt.subplots(2, 1, sharex=True, figsize=(6, 6))
 for dpl in dpls:
-    dpl.plot(ax=axes[0], layer='agg')
+    dpl.plot(ax=axes[0], layer='agg', show=False)
 net.plot_input(ax=axes[1])
 
 ###############################################################################
-# Finally, we can also plot the spikes.
-net.plot_spikes()
+# Also, we can plot the spikes and write them to txt files.
+# Note that we can use formatting syntax to specify the filename pattern
+# with which each trial will be written. To read spikes back in, we can use
+# wildcard expressions.
+net.spikes.plot()
+net.spikes.write('spk_%d.txt')
+spikes = read_spikes('spk_*.txt')
+spikes.plot()
 
 ###############################################################################
 # Now, let us try to make the exogenous driving inputs to the cells
--- a/examples/plot_simulate_evoked.py
+++ b/examples/plot_simulate_evoked.py
@@ -51,6 +51,7 @@ net.plot_input(ax=axes[1])
 # Note that we can use formatting syntax to specify the filename pattern
 # with which each trial will be written. To read spikes back in, we can use
 # wildcard expressions.
+net.spikes.plot()
 with tempfile.TemporaryDirectory() as tmp_dir_name:
     print(tmp_dir_name)
     net.spikes.write(op.join(tmp_dir_name, 'spk_%d.txt'))
--- a/examples/plot_simulate_evoked.py
+++ b/examples/plot_simulate_evoked.py
@@ -11,6 +11,7 @@ waveforms using HNN-core.
 #          Sam Neymotin <samnemo@gmail.com>
 
 import os.path as op
+import tempfile
 
 ###############################################################################
 # Let us import hnn_core
@@ -50,8 +51,10 @@ net.plot_input(ax=axes[1])
 # Note that we can use formatting syntax to specify the filename pattern
 # with which each trial will be written. To read spikes back in, we can use
 # wildcard expressions.
-net.spikes.write('spk_%d.txt')
-spikes = read_spikes('spk_*.txt')
+with tempfile.TemporaryDirectory() as tmp_dir_name:
+    print(tmp_dir_name)
+    net.spikes.write(op.join(tmp_dir_name, 'spk_%d.txt'))
+    spikes = read_spikes(op.join(tmp_dir_name, 'spk_*.txt'))
 spikes.plot()
 
 ###############################################################################
--- a/examples/plot_simulate_evoked.py
+++ b/examples/plot_simulate_evoked.py
@@ -50,7 +50,6 @@ net.plot_input(ax=axes[1])
 # Note that we can use formatting syntax to specify the filename pattern
 # with which each trial will be written. To read spikes back in, we can use
 # wildcard expressions.
-net.spikes.plot()
 net.spikes.write('spk_%d.txt')
 spikes = read_spikes('spk_*.txt')
 spikes.plot()
--- a/hnn_core/__init__.py
+++ b/hnn_core/__init__.py
@@ -2,7 +2,7 @@ from .utils import load_custom_mechanisms
 
 load_custom_mechanisms()
 
-from .dipole import simulate_dipole
+from .dipole import simulate_dipole, import_dipole
 from .feed import ExtFeed
 from .params import Params, read_params
 from .network import Network
--- a/hnn_core/__init__.py
+++ b/hnn_core/__init__.py
@@ -2,7 +2,7 @@ from .utils import load_custom_mechanisms
 
 load_custom_mechanisms()
 
-from .dipole import simulate_dipole, import_dipole
+from .dipole import simulate_dipole, read_dipole
 from .feed import ExtFeed
 from .params import Params, read_params
 from .network import Network
--- a/hnn_core/__init__.py
+++ b/hnn_core/__init__.py
@@ -5,6 +5,6 @@ load_custom_mechanisms()
 from .dipole import simulate_dipole, read_dipole
 from .feed import ExtFeed
 from .params import Params, read_params
-from .network import Network
+from .network import Network, Spikes, read_spikes
 from .pyramidal import L2Pyr, L5Pyr
 from .basket import L2Basket, L5Basket
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -134,6 +134,26 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
     return dpl
 
 
+def import_dipole(fname, units='nAm'):
+    """Read dipole values from a file and create a Dipole instance.
+
+    Parameters
+    ----------
+    fname : str
+        Full path to the input file (.txt)
+
+    Returns
+    -------
+    dpl : Dipole
+        The instance of Dipole class
+    """
+    dpl_data = np.loadtxt(fname, dtype=float)
+    dpl = Dipole(dpl_data[:, 0], dpl_data[:, 1:4])
+    if units == 'nAm':
+        dpl.units = units
+    return dpl
+
+
 class Dipole(object):
     """Dipole class.
 
@@ -277,7 +297,7 @@ class Dipole(object):
         Outputs
         -------
         txt file at fname where rows correspond to samples
-            and columns, delimited by \\t, correspond to 
+            and columns, delimited by \\t, correspond to
             1) time (s),
             2) aggregate current dipole (scaled nAm),
             3) L2/3 current dipole (scaled nAm), and
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -276,10 +276,12 @@ class Dipole(object):
 
         Outputs
         -------
-        txt file at fname where rows correspond to samples and columns,
-            delimited by \\t, correspond to 1) time (s), 2) aggregate current
-            dipole (scaled nAm), 3) L2/3 current dipole (scaled nAm), and 4) L5
-            current dipole (scaled nAm)
+        txt file at fname where rows correspond to samples
+            and columns, delimited by \\t, correspond to 
+            1) time (s),
+            2) aggregate current dipole (scaled nAm),
+            3) L2/3 current dipole (scaled nAm), and
+            4) L5 current dipole (scaled nAm)
         """
         X = np.r_[[self.t, self.dpl['agg'], self.dpl['L2'], self.dpl['L5']]].T
         np.savetxt(fname, X, fmt=['%3.3f', '%5.4f', '%5.4f', '%5.4f'],
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -273,6 +273,13 @@ class Dipole(object):
         ----------
         fname : str
             Full path to the output file (.txt)
+
+        Outputs
+        -------
+        txt file at fname where rows correspond to samples and columns,
+            delimited by \\t, correspond to 1) time (s), 2) aggregate current
+            dipole (scaled nAm), 3) L2/3 current dipole (scaled nAm), and 4) L5
+            current dipole (scaled nAm)
         """
         X = np.r_[[self.t, self.dpl['agg'], self.dpl['L2'], self.dpl['L5']]].T
         np.savetxt(fname, X, fmt=['%3.3f', '%5.4f', '%5.4f', '%5.4f'],
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -275,5 +275,5 @@ class Dipole(object):
             Full path to the output file (.txt)
         """
         X = np.r_[[self.t, self.dpl['agg'], self.dpl['L2'], self.dpl['L5']]].T
-        np.savetxt('dpl2.txt', X, fmt=['%3.3f', '%5.4f', '%5.4f', '%5.4f'],
+        np.savetxt(fname, X, fmt=['%3.3f', '%5.4f', '%5.4f', '%5.4f'],
                    delimiter='\t')
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -105,7 +105,7 @@ def _simulate_single_trial(net):
         dpl.convert_fAm_to_nAm()
         dpl.scale(net.params['dipole_scalefctr'])
         dpl.smooth(net.params['dipole_smooth_win'] / h.dt)
-    return dpl, net.spikes.times.to_python(), net.spikes.gids.to_python()
+    return dpl, net.spikes._times.to_python(), net.spikes._gids.to_python()
 
 
 def simulate_dipole(net, n_trials=1, n_jobs=1):
@@ -129,8 +129,8 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
     parallel, myfunc = _parallel_func(_clone_and_simulate, n_jobs=n_jobs)
     out = parallel(myfunc(net.params, idx) for idx in range(n_trials))
     dpl, spiketimes, spikegids = zip(*out)
-    net.spikes.times = spiketimes
-    net.spikes.gids = spikegids
+    net.spikes._times = list(spiketimes)
+    net.spikes._gids = list(spikegids)
     net.spikes.update_types(net.gid_dict)
     return dpl
 
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -128,9 +128,9 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
     """
     parallel, myfunc = _parallel_func(_clone_and_simulate, n_jobs=n_jobs)
     out = parallel(myfunc(net.params, idx) for idx in range(n_trials))
-    dpl, spiketimes, spikegids = zip(*out)
-    net.spikes._times = list(spiketimes)
-    net.spikes._gids = list(spikegids)
+    dpl, spike_times, spike_gids = zip(*out)
+    net.spikes._times = list(spike_times)
+    net.spikes._gids = list(spike_gids)
     net.spikes.update_types(net.gid_dict)
     return dpl
 
@@ -297,8 +297,8 @@ class Dipole(object):
 
         Outputs
         -------
-        txt file at fname where rows correspond to samples
-            and columns, delimited by \\t, correspond to
+        A tab separatd txt file where rows correspond
+            to samples and columns correspond to
             1) time (s),
             2) aggregate current dipole (scaled nAm),
             3) L2/3 current dipole (scaled nAm), and
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -105,7 +105,7 @@ def _simulate_single_trial(net):
         dpl.convert_fAm_to_nAm()
         dpl.scale(net.params['dipole_scalefctr'])
         dpl.smooth(net.params['dipole_smooth_win'] / h.dt)
-    return dpl, net.spiketimes.to_python(), net.spikegids.to_python()
+    return dpl, net.spikes.times.to_python(), net.spikes.gids.to_python()
 
 
 def simulate_dipole(net, n_trials=1, n_jobs=1):
@@ -129,8 +129,9 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
     parallel, myfunc = _parallel_func(_clone_and_simulate, n_jobs=n_jobs)
     out = parallel(myfunc(net.params, idx) for idx in range(n_trials))
     dpl, spiketimes, spikegids = zip(*out)
-    net.spiketimes = spiketimes
-    net.spikegids = spikegids
+    net.spikes.times = spiketimes
+    net.spikes.gids = spikegids
+    net.spikes.update_types(net.gid_dict)
     return dpl
 
 
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -134,7 +134,7 @@ def simulate_dipole(net, n_trials=1, n_jobs=1):
     return dpl
 
 
-def import_dipole(fname, units='nAm'):
+def read_dipole(fname, units='nAm'):
     """Read dipole values from a file and create a Dipole instance.
 
     Parameters
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -558,6 +558,12 @@ class Network(object):
         trial_idx : list of int
             Indices of selected trials. If None,
             all trials are selected.
+
+        Outputs
+        -------
+        txt file at fname where rows correspond to spikes and columns, delimited
+            by '\\t', correspond to 1) spike time (s), 2) spike gid, and 3) gid
+            type
         """
         if trial_idx is None:
             trial_idx = range(len(self.spiketimes))
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -13,6 +13,51 @@ from .pyramidal import L2Pyr, L5Pyr
 from .basket import L2Basket, L5Basket
 from .params import create_pext
 
+def read_spikes(self, fname, gid_dict=None):
+    """Read spike times from a file.
+
+    Parameters
+    ----------
+    fname : str
+        String format (e.g., '<pathname>/spk_*.txt') of the full
+        path to the spike file(s).
+    gid_dict : dict | None
+        Dictionary with keys 'evprox1', 'evdist1' etc.
+        containing the range of Cell or input IDs of different
+        cell or input types. If None, each spike file must contain
+        a 3rd column for spike type.
+
+    Returns
+    ----------
+    spikes : Spikes object
+        An instance of the Spikes object.
+    """
+    from glob import glob
+
+    spiketimes = ();
+    spikegids = ();
+    spiketypes = ();
+    for file in glob(fname):
+        spike_trial = np.loadtxt(file, dtype=str)
+        spiketimes = spiketimes + (list(spike_trial[:, 0].astype(float)),)
+        spikegids = spikegids + (list(spike_trial[:, 1].astype(float)),)
+
+        # Note that legacy HNN 'spk.txt' files don't contain a 3rd column for
+        # spike type. If reading a legacy version, validate that a gid_dict is
+        # provided.
+        if spike_data.shape[0] == 3:
+            spiketypes = spiketypes + (list(spike_trial[:, 2].astype(str)),)
+        else:
+            assert (gid_dict is not None, "Error: gid_dict must be provided \
+                if spike types are unspecified in 'spk.txt' file")
+            spiketypes_trial = np.empty((spike_data.shape[1], 1), dtype=str)
+            for gidtype, gids in self.gid_dict.items():
+                spikegids_mask = np.in1d(spike_trial[:, 1].astype(float), gids)
+                spiketypes_trial[spikegids_mask] = gidtype
+            spiketypes = spiketypes + (spiketypes_trial,)
+
+    return Spikes(times=spiketimes, gids=spikegids, types=spiketypes)
+
 
 class Network(object):
     """The Network class.
@@ -30,7 +75,8 @@ class Network(object):
         The list of cells
     gid_dict : dict
         Dictionary with keys 'evprox1', 'evdist1' etc.
-        containing the range of Cell IDs of different cell types.
+        containing the range of Cell IDs of different cell
+        (or input) types.
     extfeed_list : dictionary of list of ExtFeed.
         Keys are:
             'evprox1', 'evprox2', etc.
@@ -513,8 +559,62 @@ class Network(object):
             plt.show()
         return ax.get_figure()
 
-    def plot_spikes(self, ax=None, show=True):
-        """Plot the spiking activity for each cell type.
+
+# spikes.write()
+# fname = 'myexperiment_%d.txt'
+# net.spikes.write(fname)
+# from hnn_core import read_spikes
+# spikes = read_spikes(fname)
+# spikes.plot()
+
+class Spikes(object):
+    '''The Spikes class.
+
+    Parameters
+    ----------
+    times : tuple (n_trials, ) of list of float | None
+        Each element of the tuple is a trial.
+        The list contains the time stamps of spikes.
+    gids : tuple (n_trials, ) of list of float | None
+        Each element of the tuple is a trial.
+        The list contains the cell IDs of neurons that spiked.
+    types : tuple (n_trials, ) of list of float | None
+        Each element of the tuple is a trial.
+        The list contains the type of spike (e.g., evprox1 or
+        L2_pyramidal) that occured at the corresonding time stamp.
+        Each gid corresponds to a type via Network::gid_dict.
+
+    Attributes
+    ----------
+    times : tuple (n_trials, ) of list of float
+        Each element of the tuple is a trial.
+        The list contains the time stamps of spikes.
+    gids : tuple (n_trials, ) of list of float
+        Each element of the tuple is a trial.
+        The list contains the cell IDs of neurons that spiked.
+    types : tuple (n_trials, ) of list of float
+        Each element of the tuple is a trial.
+        The list contains the type of spike (e.g., evprox1 or
+        L2_pyramidal) that occured at the corresonding time stamp.
+        Each gid corresponds to a type via Network::gid_dict.
+
+    Methods
+    -------
+    plot : plot and return a matplotlib Figure object showing the
+        aggregate network spiking activity according to cell type
+    write : write spike times to a file
+    '''
+
+    def __init__(self, times=None, gids=None, types=None):
+        self.times = times
+        if times is None:
+            self.times = h.Vector()
+        self.gids = gids
+        self.types = types
+
+    def plot(self, ax=None, show=True):
+        """Plot the aggregate spiking activity according to cell
+        type.
 
         Parameters
         ----------
@@ -530,18 +630,15 @@ class Network(object):
             The matplotlib figure object
         """
         import matplotlib.pyplot as plt
-        spikes = np.array(sum(self.spiketimes, []))
-        gids = np.array(sum(self.spikegids, []))
-        spike_times = np.zeros((4, spikes.shape[0]))
+        spiketimes = np.array(sum(self.times, []))
+        spiketypes = np.array(sum(self.types, []))
         cell_types = ['L5_pyramidal', 'L5_basket', 'L2_pyramidal', 'L2_basket']
-        for idx, key in enumerate(cell_types):
-            mask = np.in1d(gids, self.gid_dict[key])
-            spike_times[idx, mask] = spikes[mask]
+        spiketimes_cell = [spiketimes[spiketypes==cell_type] for cell_type in cell_types]
 
         if ax is None:
             fig, ax = plt.subplots(1, 1)
 
-        ax.eventplot(spike_times, colors=['r', 'b', 'g', 'w'])
+        ax.eventplot(spiketimes_cell, colors=['r', 'b', 'g', 'w'])
         ax.legend(cell_types, ncol=2)
         ax.set_facecolor('k')
         ax.set_xlabel('Time (ms)')
@@ -552,75 +649,31 @@ class Network(object):
             plt.show()
         return ax.get_figure()
 
-    def write_spikes(self, fname, trial_idx=None):
+    def write(self, fname):
         """Write spike times to a file.
 
         Parameters
         ----------
         fname : str
-            Full path to the output file (.txt)
+            String format (e.g., '<pathname>/spk_%d.txt') of the full
+            path to the output spike file(s).
         trial_idx : list of int
             Indices of selected trials. If None,
             all trials are selected.
 
         Outputs
         -------
-        txt file at fname where rows correspond to
-            spikes and columns, delimited by '\\t',
-            correspond to 1) spike time (s),
-            2) spike gid, and 3) gid type
+        A numerically labelled txt file for each trial
+            where rows correspond to spikes, and columns,
+            delimited by '\\t', correspond to
+            1) spike time (s),
+            2) spike gid, and
+            3) gid type
         """
-        if trial_idx is None:
-            trial_idx = range(len(self.spiketimes))
-
-        spiketimes = []
-        spikegids = []
-        for idx in trial_idx:
-            spiketimes += self.spiketimes[idx]
-            spikegids += self.spikegids[idx]
-
-        gidtypes = np.empty_like(spikegids, dtype='<U36')
-        for spike_type, gid_range in self.gid_dict.items():
-            gidtypes[np.in1d(spikegids, gid_range)] = spike_type
-
-        with open(fname, 'w') as f:
-            for spk_idx in range(len(spiketimes)):
-                f.write('{:.3f}\t{}\t{}\n'.format(spiketimes[spk_idx],
-                        int(spikegids[spk_idx]), gidtypes[spk_idx]))
-
-    def read_spikes(self, fname, append_trial=True):
-        """Read spike times from a file.
-
-        Parameters
-        ----------
-        fname : str
-            Full path to the input file (.txt)
-        append_trials : bool
-            If True, append the contents of fname
-            (i.e., as a trial) to the spike-related
-            attributes of Network instance. If False,
-            all spikes of the Network instance will
-            be overwritten.
-        """
-        spike_data = np.loadtxt(fname, dtype=str)
-        spiketimes = spike_data[:, 0].astype(float)
-        spikegids = spike_data[:, 1].astype(float)
-
-        # Note that legacy HNN 'spk.txt' files contain only 2 columns
-        # Handle 3-column version with gid validation
-        if spike_data.shape[0] == 3:
-            gid_types = spike_data[:, 2].astype(str)
-            for gid_type in np.unique(gid_types):
-                gid_type_max = np.max(spikegids[gid_types == gid_type])
-                gid_type_min = np.min(spikegids[gid_types == gid_type])
-                assert (gid_type_max in self.gid_dict.get(gid_type, [None]) and
-                        gid_type_min in self.gid_dict.get(gid_type, [None])), \
-                    "Error: gids in %s are incompatible with the current \
-                    network" % (fname,)
-
-        if append_trial is True:
-            self.spiketimes = self.spiketimes + (list(spiketimes),)
-            self.spikegids = self.spikegids + (list(spikegids),)
-        else:
-            self.spiketimes = (list(spiketimes),)
-            self.spikegids = (list(spikegids),)
+        for trial_idx in range(len(self.spikes.times)):
+            with open(fname % (trial_idx,), 'w') as f:
+                for spk_idx in range(len(spiketimes)):
+                    f.write('{:.3f}\t{}\t{}\n'.format(
+                        self.times[trial_idx][spk_idx],
+                        int(self.gids[trial_idx][spk_idx]),
+                        self.types[trial_idx][spk_idx]))
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -556,7 +556,7 @@ class Network(object):
         fname : str
             Full path to the output file (.txt)
         trial_idx : list of int
-            Indices of selected trials. If 'all',
+            Indices of selected trials. If None,
             all trials are selected.
         """
         if trial_idx==None:
@@ -573,6 +573,6 @@ class Network(object):
             gidtypes[np.in1d(spikegids,gid_range)] = spike_type
 
         with open(fname,'w') as f:
-            for i in range(len(spiketimes)):
-                f.write('{:.3f}\t{}\t{}\n'.format(spiketimes[i],
-                    int(spikegids[i]),gidtypes[i]))
+            for spk_idx in range(len(spiketimes)):
+                f.write('{:.3f}\t{}\t{}\n'.format(spiketimes[spk_idx],
+                    int(spikegids[spk_idx]),gidtypes[spk_idx]))
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -547,3 +547,30 @@ class Network(object):
         if show:
             plt.show()
         return ax.get_figure()
+
+    def write_spikes(self, fname, trial_idx='all'):
+        """Write spike times to a file.
+
+        Parameters
+        ----------
+        fname : str
+            Full path to the output file (.txt)
+        trial_idx : list of int
+            Indices of selected trials. If 'all',
+            all trials are selected.
+        """
+        if trial_idx=='all':
+            spiketimes = sum(self.spiketimes,[])
+            spikegids = sum(self.spikegids,[])
+        else:
+            spiketimes = sum(self.spiketimes[trial_idx],[])
+            spikegids = sum(self.spikegids[trial_idx],[])
+
+        gidtypes = np.empty_like(spikegids,dtype='<U36')
+        for spike_type,gid_range in self.gid_dict.items():
+            gidtypes[np.in1d(spikegids,gid_range)] = spike_type
+
+        with open(fname,'w') as f:
+            for i in range(len(spiketimes)):
+                f.write('{:.3f}\t{}\t{}\n'.format(spiketimes[i],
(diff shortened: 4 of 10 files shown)


hnn_core/network.py
```diff
@@ -518,3 +518,32 @@ def plot_spikes(self, ax=None, show=True):
...
+        spikegids = []
+        for i in trial_idx:
+            spiketimes += self.spiketimes[i]
+            spikegids += self.spikegids[i]
+
+        gidtypes = np.empty_like(spikegids,dtype='<U36')
+        for spike_type,gid_range in self.gid_dict.items():
+            gidtypes[np.in1d(spikegids,gid_range)] = spike_type
+
+        with open(fname,'w') as f:
```
jasmainak: you should use standard functions whenever possible. In this case, I think you could use [np.savetxt](https://docs.scipy.org/doc/numpy/reference/generated/numpy.savetxt.html)

hnn_core/network.py
```diff
@@ -518,3 +518,32 @@ def plot_spikes(self, ax=None, show=True):
...
             plt.show()
         return ax.get_figure()
+
+    def write_spikes(self, fname, trial_idx='all'):
+        """Write spike times to a file.
+
+        Parameters
+        ----------
+        fname : str
+            Full path to the output file (.txt)
```
jasmainak: What motivated the use of a `.txt` file instead of say an `.npz` or `.mat` file?

hnn_core/network.py
```diff
@@ -518,3 +518,32 @@ def plot_spikes(self, ax=None, show=True):
...
+        trial_idx : list of int
+            Indices of selected trials. If 'all',
+            all trials are selected.
+        """
+        if trial_idx=='all':
```
jasmainak: use `None`. The problem with `'all'` is that string comparison may fail if it's not the right case

hnn_core/network.py
```diff
@@ -518,3 +518,32 @@ def plot_spikes(self, ax=None, show=True):
         if show:
             plt.show()
         return ax.get_figure()
+
+    def write_spikes(self, fname, trial_idx='all'):
+        """Write spike times to a file.
```
jasmainak: how do you plan to read them back in?

hnn_core/network.py
```diff
@@ -518,3 +518,32 @@ def plot_spikes(self, ax=None, show=True):
...
+
+        Parameters
+        ----------
+        fname : str
+            Full path to the output file (.txt)
+        trial_idx : list of int
+            Indices of selected trials. If 'all',
+            all trials are selected.
```
jasmainak: can you also document what the file actually looks like?

hnn_core/network.py
```diff
@@ -518,3 +518,32 @@ def plot_spikes(self, ax=None, show=True):
...
+        """
+        if trial_idx=='all':
+            trial_idx = range(len(self.spiketimes))
+
+        spiketimes = []
+        spikegids = []
+        for i in trial_idx:
```
jasmainak: avoid use of single letter variable names when possible. They are hard to Ctrl + F

hnn_core/network.py
rythorpe: Because I combined data types (i.e., list of floats and list of strings) which don't easily combine into a single array, the most succinct way to handle the formatting was to use low-level I/O functions. I could potentially convert each list to a formatted string, then convert it to a numpy array, and then save it if you think that'd be better?

Note that HNN originally stored spike.txt files along side param.txt files that stored the gid ranges. I tried to retain the HNN format but also combine all the relevant spiking info into a single file.

hnn_core/network.py
rythorpe: I just wanted to be consistent with the `dipole.write()` function and previous HNN I/O convention.

hnn_core/network.py
rythorpe: The file is formatted in three columns where each row is <spike_time>\t<spike_gid>\t<gid_type>. There is an example in the demo.zip folder. Are you looking for a formal description in the function documentation?

hnn_core/network.py
rythorpe: Check out the plot_simulate_evoked.py script in the attached demo.zip folder. Are you thinking we should have a built-in read function as well?

hnn_core/network.py
blakecaldwell: Since testing new hnn-core output against old hnn output will be important for the foreseeable future, I plead that we leave files the same (.txt). We will have test functions to compare array values after loading from the file, but also being able to do a `diff` from the command line is useful.

hnn_core/network.py
blakecaldwell: HNN has the functionality to take a given parameter file, find the spike output in an expected directory and load it. If this functionality is part of hnn-core, HNN can use it. Otherwise, the function will still be part of HNN.

hnn_core/network.py
blakecaldwell: Since we are proceeding with using hnn-core for HNN. I'd prefer to keep spikes for different trials in separate files, as with HNN. If it's not, I will have to write a function for HNN that reads the single file and then splits into different files. :(

blakecaldwell:
Thanks @rythorpe!

hnn_core/network.py
rythorpe: It is currently setup to take an argument (i.e., `trial_idx`) that specifies which trials are selected to write to the file, so the write function can be looped over to save all trials separately. The reason for this was to mimic the low-level functionality of the `dipole.write()` function which also only writes one file at a time. Would it be more helpful if we changed both the `dipole.write()` and `network.write_spikes()` functions to output all files with the same naming convention as HNN?

hnn_core/network.py
rythorpe: So are you in favor of adding a read function in hnn-core or would you rather that be handled by an HNN wrapper?

hnn_core/network.py
jasmainak: yep, a formal description in the function docstring

hnn_core/network.py
jasmainak: I think you should add a reader if you're going to have a way to write it :) It's also easier to test that way. You can do a round trip and check if you have what you started with.

hnn_core/dipole.py
```diff
@@ -275,5 +275,5 @@ def write(self, fname):
             Full path to the output file (.txt)
         """
         X = np.r_[[self.t, self.dpl['agg'], self.dpl['L2'], self.dpl['L5']]].T
-        np.savetxt('dpl2.txt', X, fmt=['%3.3f', '%5.4f', '%5.4f', '%5.4f'],
+        np.savetxt(fname, X, fmt=['%3.3f', '%5.4f', '%5.4f', '%5.4f'],
```
jasmainak: can you also add a test for this? Something like:

```py
assert op.exists(fname)
```

This way we won't have a regression of the bug

jasmainak:
@rythorpe also don't forget to update `whats_new.rst` and `api.rst` once you have added the reader function.

hnn_core/network.py
rythorpe: Since spikes currently exist only as a `network` class attribute, they either need to be read in through a `network` class method at/after network initialization or handled as their own object outside of the `network` class (as with params). In keeping with our desire to make HNN data structures more transparent and tractable (e.g., using .json files for params, json schema for param validation), what do you think of turning spikes into their own object with the following example structure:
```
{
    <gid_type>: {
        <gid>: [
            <spiketime[0]>,
            <spiketime[1]>,
            ...
            ]
        }
}
```
We could still maintain the .txt file format convention of HNN output for now, but this would make the spike data more intuitive, modular, and readable into a `network` instance with proper tests and validation.

jasmainak: Humm ... I would keep it simple for now. You can read in a dictionary of lists. The dictionary would contain three keys `gid`, `gid_type` and `spiketimes`. It sounds like this stuff should belong to the `_Cell` class but I'm sure it's nasty to save it directly and read it back. Let's stick to python built-ins until it becomes apparent if something should go to a separate class.

hnn_core/network.py
```diff
@@ -518,3 +518,32 @@ def plot_spikes(self, ax=None, show=True):
...
+
+        Parameters
+        ----------
+        fname : str
+            Full path to the output file (.txt)
+        trial_idx : list of int
+            Indices of selected trials. If None,
+            all trials are selected.
+        """
+        if trial_idx==None:
```
jasmainak:
```suggestion
        if trial_idx is None:
```

jasmainak: https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html

hnn_core/network.py
```diff
@@ -518,3 +518,32 @@ def plot_spikes(self, ax=None, show=True):
...
+            trial_idx = range(len(self.spiketimes))
+
+        spiketimes = []
+        spikegids = []
+        for idx in trial_idx:
+            spiketimes += self.spiketimes[idx]
+            spikegids += self.spikegids[idx]
+
+        gidtypes = np.empty_like(spikegids,dtype='<U36')
+        for spike_type,gid_range in self.gid_dict.items():
```
jasmainak: you need a PEP8 checker. For some reason Travis didn't run on your branch and PEP8 didn't run. Let me rebase your branch so you see the problem.

hnn_core/network.py
jasmainak: thinking more, you could probably use a list of [NamedTuple](https://docs.python.org/2/library/collections.html#collections.namedtuple). Maybe that simplifies some code, I don't know ... would keep it for another PR.

hnn_core/network.py
jasmainak: That's weird your branch is up to date. @blakecaldwell any idea why Travis is not running? I saw you do some recent experiments on HNN with Travis.

blakecaldwell: I'm not sure... Do you have this repository on your Travis dashboard? I haven't linked it to my account. I've found the Travis dashboard to have useful hints under the "requests" section. If it fails because of a bad .travis.yml, for example, it will indicate it there (doesn't appear to the case here, though).

hnn_core/network.py
blakecaldwell: Yeah, a reader would be helpful. In the future, I like the idea of moving spiketime and spikegids out of the network class. I'm in favor of either keeping things as close to HNN for now or the other extreme of making a new Spike or Spikedata class.

hnn_core/network.py
jasmainak: I do have it under the dashboard. Maybe @rythorpe just try pushing another commit and it might work ...

hnn_core/network.py
blakecaldwell: Oh, yes. This is where a separate Spike (Spikedata) class would be helpful for parity with Dipole: Spikedata.write() and Spikedata.write_all_trials(). That way you can enforce some standard naming (like HNN for better or worse) and don't have the ambiguous case of naming a file with some, but not all trials.

There's more than one way to do it. Just my thoughts. Thanks for your work on this.

hnn_core/network.py
rythorpe: The namedtuple class does look super convenient for this - good call. In the short term, I'll just implement a simple read function that validates gids prior to them overwriting the spikes of a `network` instance (to be compatible with the ranges in `network.gid_dict`).

jasmainak:
@rythorpe do you see an error when pushing? what's the issue?

rythorpe:
> @rythorpe do you see an error when pushing? what's the issue?

My latest [commit](https://github.com/rythorpe/hnn-core/commit/a1d421988ac0ac773f8c4fa8801e89dd031f8d73) doesn't appear in this thread. Am I missing something?

jasmainak:
@rythorpe that was so strange. I'm guessing it was related to [this incident](https://www.githubstatus.com/incidents/80d0cs6kpsps). I fixed it for you now by making a new commit, pushing again, and then removing that commit.

rythorpe:
I logged the following error traceback when trying to update the documentation with `$ make html` from within the hnn-core/doc/ directory. Any idea what is going on?

```# Sphinx version: 2.4.4
# Python version: 3.7.7 (CPython)
# Docutils version: 0.16 release
# Jinja2 version: 2.11.1
# Last messages:

# Loaded extensions:
Traceback (most recent call last):
  File "/home/ryan/anaconda3/envs/hnn_core/lib/python3.7/site-packages/sphinx/cmd/build.py", line 275, in build_main
    args.tags, args.verbosity, args.jobs, args.keep_going)
  File "/home/ryan/anaconda3/envs/hnn_core/lib/python3.7/site-packages/sphinx/application.py", line 278, in __init__
    self._init_builder()
  File "/home/ryan/anaconda3/envs/hnn_core/lib/python3.7/site-packages/sphinx/application.py", line 334, in _init_builder
    self.events.emit('builder-inited')
  File "/home/ryan/anaconda3/envs/hnn_core/lib/python3.7/site-packages/sphinx/events.py", line 99, in emit
    results.append(callback(self.app, *args))
  File "/home/ryan/anaconda3/envs/hnn_core/lib/python3.7/site-packages/sphinx_gallery/gen_gallery.py", line 291, in generate_gallery_rst
    gallery_conf = parse_config(app)
  File "/home/ryan/anaconda3/envs/hnn_core/lib/python3.7/site-packages/sphinx_gallery/gen_gallery.py", line 91, in parse_config
    abort_on_example_error, lang, app.builder.name, app)
  File "/home/ryan/anaconda3/envs/hnn_core/lib/python3.7/site-packages/sphinx_gallery/gen_gallery.py", line 215, in _complete_gallery_conf
    "found type %s" % type(backref))
ValueError: The 'backreferences_dir' parameter must be of type str, pathlib.Path or None, found type <class 'bool'>
```

jasmainak:
I think we need to update that in `conf.py`. There have been some changes in `sphinx-gallery` ...

jasmainak:
You need to rebase @rythorpe ? Are you able to do it yourself?

Also -- I think we need to discuss this PR in person. Maybe on Thursday if you have time?

rythorpe:
> You need to rebase @rythorpe ? Are you able to do it yourself?
> 
> Also -- I think we need to discuss this PR in person. Maybe on Thursday if you have time?

I can rebase, though I don't understand why? Sure, Thursday works for me.

jasmainak:
You need to rebase because there are conflicts now. You edited the same file in both the pull requests.

jasmainak:
Please make a back up of your branch before you begin rebasing. It might be a mess if you are trying the first time

rythorpe:
@jasmainak Any idea what is going wrong with the circleci error?

jasmainak:
Ignore it for now, I have to configure CircleCI ...

hnn_core/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 load_custom_mechanisms()
 
-from .dipole import simulate_dipole
+from .dipole import simulate_dipole, import_dipole
```
jasmainak, line 5:
```suggestion
from .dipole import simulate_dipole, read_dipole
```

jasmainak: also update `api.rst`

hnn_core/network.py
```diff
@@ -518,3 +522,76 @@ def plot_spikes(self, ax=None, show=True):
...
+        Parameters
+        ----------
+        fname : str
+            Full path to the input file (.txt)
+        append_trials : bool
+            If True, append the contents of fname
+            (i.e., as a trial) to the spike-related
+            attributes of Network instance. If False,
+            all spikes of the Network instance will
+            be overwritten.
```
jasmainak, line 574: Returns?

jasmainak:
@rythorpe looks like you need to rebase

hnn_core/network.py
```diff
@@ -518,3 +522,76 @@ def plot_spikes(self, ax=None, show=True):
...
+        gidtypes = np.empty_like(spikegids, dtype='<U36')
+        for spike_type, gid_range in self.gid_dict.items():
+            gidtypes[np.in1d(spikegids, gid_range)] = spike_type
+
+        with open(fname, 'w') as f:
+            for spk_idx in range(len(spiketimes)):
+                f.write('{:.3f}\t{}\t{}\n'.format(spiketimes[spk_idx],
+                        int(spikegids[spk_idx]), gidtypes[spk_idx]))
+
+    def read_spikes(self, fname, append_trial=True):
```
jasmainak, line 562:
```suggestion
    def read_spikes(self, fname):
```

just allow `fname` to have a wildcard expression and loop over all the files that match that expression and return a list.

jasmainak:
An example also needs to be updated as we discussed!

rythorpe:
> An example also needs to be updated as we discussed!

@jasmainak yes, sorry - I wanted to run a few tests before pushing my most recent commits where I address the remaining items we discussed. I'll ping you when I push this afternoon.

jasmainak:
okay sounds good!

rythorpe:
@jasmainak I've created the `Spikes` class within network.py and integrated its functionality into `Network`, the examples (specifically, look at plot_simulate_evoked.py), and test_network.py. Let me know what you think.

hnn_core/network.py
```diff
@@ -14,6 +14,54 @@
...
+        containing the range of Cell or input IDs of different
+        cell or input types. If None, each spike file must contain
+        a 3rd column for spike type.
+
+    Returns
+    ----------
+    spikes : Spikes object
+        An instance of the Spikes object.
+    """
+    from glob import glob
```
jasmainak, line 37: You should always prefer global imports when writing a package. The reason is that the import will happen every time you call this function. So imagine if there is a function which calls this function 100 times, you will have `glob` imported 100 times and each of those imports will take some time. The only exceptions to this rule are:

1) matplotlib imports. We do these nested inside functions because of some weird bug. I don't exactly know the reason.
2) circular imports. If you file A containing function a1 calls file B containing function b, and then function b calls a function a2 in file A. Then you would need to nest the function. Usually this is an indicator of bad code structure though.
3) partial dependencies: if you want to have a partial dependency on a package so that it's imported only by users who want to use an exotic function. This should be used sparingly.

hnn_core/network.py
```diff
@@ -14,6 +14,54 @@
 from .params import create_pext
 
 
+def read_spikes(fname, gid_dict=None):
```
jasmainak, line 17: This function should be added to the `api.rst` page: https://45-168215891-gh.circle-artifacts.com/0/html/api.html. Maybe add a subsection called input/output?

hnn_core/network.py
```diff
@@ -509,8 +557,99 @@ def plot_input(self, ax=None, show=True):
             plt.show()
         return ax.get_figure()
 
-    def plot_spikes(self, ax=None, show=True):
-        """Plot the spiking activity for each cell type.
+
+class Spikes(object):
```
jasmainak, line 561: can you add it to the public API here: https://45-168215891-gh.circle-artifacts.com/0/html/api.html# ? We are not yet finalized as we haven't made a release but I think it's nice to have it there so we have an overview of the documentation

hnn_core/network.py
```diff
@@ -525,25 +664,53 @@ def plot_spikes(self, ax=None, show=True):
...
         ax.set_ylim((-1, 4.5))
+        ax.set_xlim(left=0)
 
         if show:
             plt.show()
         return ax.get_figure()
+
+    def write(self, fname):
+        """Write spiking activity to a collection of spike trial
+        files.
```
jasmainak, line 692: This should be a single line. Quoting [PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings):

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

hnn_core/network.py
```diff
@@ -509,8 +557,99 @@ def plot_input(self, ax=None, show=True):
...
 
-    def plot_spikes(self, ax=None, show=True):
-        """Plot the spiking activity for each cell type.
+
+class Spikes(object):
+    '''The Spikes class.
+
+    Parameters
+    ----------
+    times : tuple (n_trials, ) of list of float | None
```
jasmainak, line 566: I think the code below will simplify if we use `np.ndarray` instead of tuples for `times`, `gids`, and `types`.

hnn_core/network.py
jasmainak: Starting from Python 3, you don't need to inherit from `object`. But I guess we have done this elsewhere, so let's remove it in another pull request

hnn_core/network.py
```diff
@@ -509,8 +557,99 @@ def plot_input(self, ax=None, show=True):
...
+
+    def __init__(self, times=None, gids=None, types=None):
+        from neuron import h
+
+        self.times = times
+        if times is None:
+            self.times = h.Vector()
+        self.gids = gids
+        if gids is None:
+            self.gids = h.Vector()
```
jasmainak, line 610:
```suggestion
        if gids is None:
            gids = h.Vector()
        self.gids = gids
```

if this works, I feel it's a bit easier to read. Can you also do the same for `times`

hnn_core/tests/test_network.py
```diff
@@ -28,3 +28,11 @@ def test_network():
     for ev_input in params['t_ev*']:
         type_key = ev_input[2: -2] + ev_input[-1]
         assert len(net.gid_dict[type_key]) == net.n_cells
+
+    # Test read/write spikes consistency
```
jasmainak, line 32: I think you need to add a function called `test_spikes` and check each of the methods. Here you haven't tested the `__repr__` for instance. See [how we test repr in MNE](https://github.com/mne-tools/mne-python/blob/ddc007f44c5eb840fcfd16dd3987b6535bed9f44/mne/forward/tests/test_forward.py#L55-L59)

examples/plot_simulate_evoked.py
```diff
@@ -42,12 +42,18 @@
...
 net.plot_input(ax=axes[1])
 
 ###############################################################################
-# Finally, we can also plot the spikes.
-net.plot_spikes()
+# Also, we can plot the spikes and write them to txt files.
+# Note that we can use formatting syntax to specify the filename pattern
+# with which each trial will be written. To read spikes back in, we can use
+# wildcard expressions.
+net.spikes.plot()
```
jasmainak, line 53:
```suggestion
```
I think it's fine to have just one plot

hnn_core/network.py
```diff
@@ -509,8 +557,99 @@ def plot_input(self, ax=None, show=True):
...
+        Each element of the tuple is a trial.
+        The list contains the cell IDs of neurons that spiked.
+    types : tuple (n_trials, ) of list of float
+        Each element of the tuple is a trial.
+        The list contains the type of spike (e.g., evprox1 or
+        L2_pyramidal) that occured at the corresonding time stamp.
+        Each gid corresponds to a type via Network::gid_dict.
+
+    Methods
+    -------
```
jasmainak, line 593: I have never seen this before. I looked up the [numpy docstring convention](https://numpydoc.readthedocs.io/en/latest/format.html). They say it's not really necessary but it's fine to have it. Could you format it to closely resemble it though?

hnn_core/network.py
```diff
@@ -509,8 +557,99 @@ def plot_input(self, ax=None, show=True):
...
+    plot : plot and return a matplotlib Figure object showing the
+        aggregate network spiking activity according to cell type
+    write : write spiking activity to a collection of spike trial
+        files
+    '''
+
+    def __init__(self, times=None, gids=None, types=None):
+        from neuron import h
+
+        self.times = times
```
jasmainak, line 605: you need to do some input validation. Check that the shapes are consistent etc. Something like [this](https://github.com/mne-tools/mne-python/blob/master/mne/epochs.py#L403-L406)

hnn_core/network.py
```diff
@@ -429,7 +477,7 @@ def _record_spikes(self):
         # agnostic to type of source, will sort that out later
         for gid in self._gid_list:
             if pc.gid_exists(gid):
-                pc.spike_record(gid, self.spiketimes, self.spikegids)
+                pc.spike_record(gid, self.spikes.times, self.spikes.gids)
```
jasmainak, line 480: how about `spikes.record(pc)`? That way you can keep everything (`times`, `gids` etc) as simple objects, e.g., numpy array and only in this method you instantiate `h.Vector()` ?

hnn_core/network.py
```diff
@@ -509,8 +557,99 @@ def plot_input(self, ax=None, show=True):
...
+
+class Spikes(object):
+    '''The Spikes class.
+
+    Parameters
+    ----------
+    times : tuple (n_trials, ) of list of float | None
+        Each element of the tuple is a trial.
+        The list contains the time stamps of spikes.
+    gids : tuple (n_trials, ) of list of float | None
```
jasmainak, line 569: oh wait, now I understand why you don't use numpy array. It's because each trial has different number of spikes? I think it would still be good to use list of list instead of tuple of list. Tuples are immutable. You'd use them when you don't expect to update the array

hnn_core/network.py
```diff
@@ -509,8 +557,99 @@ def plot_input(self, ax=None, show=True):
...
+        self.types = types
+
+    def __repr__(self):
+        class_name = self.__class__.__name__
+        num_trials = len(self.times)
+        return '<%s | %d trials>' % (class_name, num_trials)
+
+    def __eq__(self, other):
+        if not isinstance(other, Spikes):
+            return NotImplemented
```
jasmainak, line 620: just raise a `ValueError` saying you can't compare `spikes` object to an object that's of a different type

hnn_core/network.py
```diff
@@ -509,8 +557,99 @@ def plot_input(self, ax=None, show=True):
...
+        # Round each element of each list within the tuple
+        times_self = [[round(time, 3) for time in trial]
+                      for trial in self.times]
+        times_other = [[round(time, 3) for time in trial]
+                       for trial in other.times]
```
jasmainak, line 625: Again you need some input validation. Before you check equality, you need to ensure that the shapes are consistent

hnn_core/network.py
```diff
@@ -14,6 +14,54 @@
...
+
+        # Note that legacy HNN 'spk.txt' files don't contain a 3rd column for
+        # spike type. If reading a legacy version, validate that a gid_dict is
+        # provided.
+        if spike_trial.shape[1] == 3:
+            spiketypes += (list(spike_trial[:, 2].astype(str)),)
+        else:
+            assert gid_dict is not None, ("Error: gid_dict must be provided "
+                                          "if spike types are unspecified in "
+                                          "'spk.txt' file")
```
jasmainak, line 55: the right way to raise errors is:

```suggestion
                                          "'spk.txt' file")
```

jasmainak, line 55: The right way to raise errors is:

```suggestion
            if gid_dict is None:
                raise ValueError('gid_dict must be provided blah blah')
```

then you should add a test to check if these errors are actually being tested You can use `pytest.raises`

hnn_core/network.py
```diff
@@ -509,8 +557,99 @@ def plot_input(self, ax=None, show=True):
...
+        if times is None:
+            self.times = h.Vector()
+        self.gids = gids
+        if gids is None:
+            self.gids = h.Vector()
+        self.types = types
+
+    def __repr__(self):
+        class_name = self.__class__.__name__
+        num_trials = len(self.times)
```
jasmainak, line 615:
```suggestion
        n_trials = len(self.times)
```

hnn_core/network.py
jasmainak: or maybe `TypeError`

hnn_core/network.py
```diff
@@ -509,8 +557,99 @@ def plot_input(self, ax=None, show=True):
...
+        Parameters
+        ----------
+        gid_dict : dict of lists or range objects
+            Dictionary with keys 'evprox1', 'evdist1' etc.
+            containing the range of Cell or input IDs of different
+            cell or input types.
+        """
+
+        spiketypes = ()
+        for trl_idx in range(len(self.times)):
```
jasmainak, line 642:
```suggestion
        for trial_idx in range(len(self.times)):
```

hnn_core/network.py
```diff
@@ -525,25 +664,53 @@ def plot_spikes(self, ax=None, show=True):
...
+            where rows correspond to spikes, and columns,
+            delimited by '\\t', correspond to
+            1) spike time (s),
+            2) spike gid, and
+            3) gid type
+        """
+
+        for trial_idx in range(len(self.times)):
+            with open(fname % (trial_idx,), 'w') as f:
+                for spk_idx in range(len(self.times[trial_idx])):
```
jasmainak, line 712:
```suggestion
                for spike_idx in range(len(self.times[trial_idx])):
```

jasmainak:
@rythorpe I did a pass on comments. Most of them should be pretty easy to address. Let me know if you get stuck!

hnn_core/network.py
jasmainak: actually never mind this comment, I'm not so sure of this

hnn_core/network.py
jasmainak: Thinking a bit more about it, do we even need the "list of list". Why not just a flat list? Do we ever need to know which spikes belong to which trials?

rythorpe: > oh wait, now I understand why you don't use numpy array. It's because each trial has different number of spikes? I think it would still be good to use list of list instead of tuple of list. Tuples are immutable. You'd use them when you don't expect to update the array

I like that spike data is stored in tuples because it makes it clear that spike data should only be modified by the `Network` class and should thus remain immutable to the user. The trade-off is that this makes our code a little less sleek. If you prefer otherwise, however, I'm happy to change it to a list of lists.

>Thinking a bit more about it, do we even need the "list of list". Why not just a flat list? Do we ever need to know which spikes belong to which trials?

HNN-gui currently plots spiking activity according to trial so that the user can visualize both aggregate and single-trial spikes. This is reflected in the current HNN I/O convention to write spikes to files on a single-trial basis. Furthermore, if we were ever to attempt running stats on different spiking states that emerge across trials, it would be really important to distinguish between trials.

jasmainak: > I like that spike data is stored in tuples because it makes it clear that spike data should only be modified by the Network class and should thus remain immutable to the user.

I see, the way you would do this then is through a private attribute e.g., `_times` and then use a "property" decorator on a method called "times", so the user can only access it but not modify it (since they are not supposed to touch private attributes that start with an underscore). See [here](https://github.com/mne-tools/mne-python/blob/master/mne/epochs.py#L1452-L1455) for an example. I feel mixing tuples and lists might confuse someone reading the code in the future.

jasmainak: > HNN-gui currently plots spiking activity according to trial so that the user can visualize both aggregate and single-trial spikes. This is reflected in the current HNN I/O convention to write spikes to files on a single-trial basis. Furthermore, if we were ever to attempt running stats on different spiking states that emerge across trials, it would be really important to distinguish between trials.

okay fair enough!

rythorpe: Would we then need to add a `set_times()` function so that `Network` can modify the `Spikes::times` attribute?

jasmainak: umm ... I don't think you need it. You are allowed to modify private attributes in the hnn-core code. But the user is not supposed to touch it. A `set_times` might be useful if it's non-trivial to change the `spikes.times` -- e.g., if it's linked to other attributes so to change it safely to add some logic.

rythorpe:
@jasmainak I've incorporated input validation for spikes-related functions, added a bunch more tests, and made the `Spikes` attributes private along with a bunch of small cosmetic changes per your recommendations. As soon as we pass these most recent changes I'll update `api.rst` and `whats_new.rst`.

hnn_core/network.py
rythorpe: Based on my limited understanding of special methods, returning `NotImplemented` allows the interpreter to run `other.__eq__(self)` if `self.__eq__(other)` fails.

examples/plot_simulate_evoked.py
```diff
@@ -42,12 +42,17 @@
...
 
 ###############################################################################
-# Finally, we can also plot the spikes.
-net.plot_spikes()
+# Also, we can plot the spikes and write them to txt files.
+# Note that we can use formatting syntax to specify the filename pattern
+# with which each trial will be written. To read spikes back in, we can use
+# wildcard expressions.
+net.spikes.write('spk_%d.txt')
+spikes = read_spikes('spk_*.txt')
```
jasmainak, line 54: could we write this to a temporary directory to avoid messing the current directory that the user is in? You can use something here: https://docs.python.org/3/library/tempfile.html. Maybe `mkdtemp`

hnn_core/tests/test_network.py
```diff
@@ -28,3 +31,51 @@ def test_network():
...
+    # (e.g., when times is a tuple of lists)
+    with pytest.raises(TypeError) as excinfo:
+        spiketimes_tuple = ([2.3456, 7.89], [4.2812, 93.2])
+        spikes = Spikes(times=spiketimes_tuple, gids=spikegids,
+                        types=spiketypes)
+    assert "times should be a list of lists" in str(excinfo.value)
+
+    # TypeError should be raised when one of the args is entered as a list of
+    # non-lists (e.g., when times is a list of ints)
+    with pytest.raises(TypeError) as excinfo:
```
jasmainak, line 60: instead of creating `excinfo` you can use the argument `'match'` in `pytest.raises`. I think it's a cleaner way to do it.

hnn_core/network.py
```diff
@@ -523,27 +694,54 @@ def plot_spikes(self, ax=None, show=True):
...
+
+        Parameters
+        ----------
+        fname : str
+            String format (e.g., '<pathname>/spk_%d.txt') of the
+            path to the output spike file(s).
+
+        Outputs
+        -------
+        A numerically labelled txt file for each trial
```
jasmainak, line 733:
```suggestion
        A tab separated txt file for each trial
```

These kind of files are usually called tsv files: https://en.wikipedia.org/wiki/Tab-separated_values

hnn_core/network.py
```diff
@@ -523,27 +694,54 @@ def plot_spikes(self, ax=None, show=True):
...
         ax.get_yaxis().set_visible(False)
         ax.set_ylim((-1, 4.5))
+        ax.set_xlim(left=0)
 
         if show:
             plt.show()
         return ax.get_figure()
+
+    def write(self, fname):
+        """Write spiking activity to a collection of spike trial files.
```
jasmainak, line 723:
```suggestion
        """Write spiking activity per trial to a collection of files.
```
Maybe clearer?

hnn_core/network.py
```diff
@@ -509,8 +554,134 @@ def plot_input(self, ax=None, show=True):
...
+    types : list [n_trials, ] of list of float | None
+        Each element of the 1st-order list is a trial.
+        The 2nd-order list contains the type of spike (e.g., evprox1
+        or L2_pyramidal) that occured at the corresonding time stamp.
+        Each gid corresponds to a type via Network::gid_dict.
+
+    Attributes
+    ----------
+    times : list [n_trials, ] of list of float
+        Each element of the 1st-order list is a trial.
```
jasmainak, line 579:
```suggestion
        Each element of the outer list is a trial.
```

hnn_core/network.py
```diff
@@ -509,8 +554,134 @@ def plot_input(self, ax=None, show=True):
...
+        The 2nd-order list contains the time stamps of spikes.
```
jasmainak, line 580:
```suggestion
        The inner list contains the time stamps of spikes.
```

hnn_core/network.py
```diff
@@ -509,8 +554,134 @@ def plot_input(self, ax=None, show=True):
...
 
-    def plot_spikes(self, ax=None, show=True):
-        """Plot the spiking activity for each cell type.
+
+class Spikes(object):
+    '''The Spikes class.
+
+    Parameters
+    ----------
+    times : list [n_trials, ] of list of float | None
```
jasmainak, line 563:
```suggestion
    times : list (n_trials, ) of list (n_spikes) of float, shape | None
```

size is usually a tuple. Following numpy convention. It's immutable

hnn_core/network.py
```diff
@@ -509,8 +554,134 @@ def plot_input(self, ax=None, show=True):
...
+        The 2nd-order list contains the time stamps of spikes.
+    gids : list [n_trials, ] of list of float | None
+        Each element of the 1st-order list is a trial.
+        The 2nd-order list contains the cell IDs of neurons that
+        spiked.
+    types : list [n_trials, ] of list of float | None
+        Each element of the 1st-order list is a trial.
+        The 2nd-order list contains the type of spike (e.g., evprox1
+        or L2_pyramidal) that occured at the corresonding time stamp.
+        Each gid corresponds to a type via Network::gid_dict.
```
jasmainak, line 574:
```suggestion
        Each gid corresponds to a type via Network().gid_dict.
```

isn't that convention from C or C++?

hnn_core/network.py
```diff
@@ -509,8 +554,134 @@ def plot_input(self, ax=None, show=True):
...
+    plot(ax=None, show=True)
+        Plot and return a matplotlib Figure object showing the
+        aggregate network spiking activity according to cell type.
+    write(fname)
+        Write spiking activity to a collection of spike trial files.
+    '''
+
+    def __init__(self, times=None, gids=None, types=None):
+        if times is None:
+            times = []
```
jasmainak, line 604:
```suggestion
            times = list()
```

We realized some time back that it's very easy to confuse [] and (). It's better to use the full name

hnn_core/network.py
```diff
@@ -509,8 +554,134 @@ def plot_input(self, ax=None, show=True):
...
+            Dictionary with keys 'evprox1', 'evdist1' etc.
+            containing the range of Cell or input IDs of different
+            cell or input types.
+        """
+
+        spiketypes = []
+        for trial_idx in range(len(self._times)):
+            spiketypes_trial = np.empty_like(self._times[trial_idx],
+                                             dtype='<U36')
+            for gidtype, gids in gid_dict.items():
```
jasmainak, line 677:
```suggestion
            for gid_type, gids in gid_dict.items():
```

I think it's better to put underscore so that it's clear where one word of the variable name ends and the second starts.

hnn_core/network.py
```diff
@@ -509,8 +554,134 @@ def plot_input(self, ax=None, show=True):
...
+
+        Parameters
+        ----------
+        gid_dict : dict of lists or range objects
+            Dictionary with keys 'evprox1', 'evdist1' etc.
+            containing the range of Cell or input IDs of different
+            cell or input types.
+        """
+
+        spiketypes = []
```
jasmainak, line 673:
```suggestion
        spike_types = list()
```

hnn_core/network.py
```diff
@@ -509,8 +554,134 @@ def plot_input(self, ax=None, show=True):
...
+        for trial_idx in range(len(self._times)):
+            spiketypes_trial = np.empty_like(self._times[trial_idx],
+                                             dtype='<U36')
+            for gidtype, gids in gid_dict.items():
+                spikegids_mask = np.in1d(self._gids[trial_idx], gids)
+                spiketypes_trial[spikegids_mask] = gidtype
+            spiketypes += [list(spiketypes_trial)]
```
jasmainak, line 680: you can probably check by using `timeit` but I suspect this is slower than doing:
```suggestion
            spiketypes.append(spiketypes_trial)
```
also clearer. You don't need to do it the old way now that you don't have a tuple

hnn_core/network.py
```diff
@@ -509,8 +554,134 @@ def plot_input(self, ax=None, show=True):
...
+            cell or input types.
+        """
+
+        spiketypes = []
+        for trial_idx in range(len(self._times)):
+            spiketypes_trial = np.empty_like(self._times[trial_idx],
+                                             dtype='<U36')
+            for gidtype, gids in gid_dict.items():
+                spikegids_mask = np.in1d(self._gids[trial_idx], gids)
+                spiketypes_trial[spikegids_mask] = gidtype
```
jasmainak, line 679: what happens if some of the `gid_dict` values have overlapping ranges? Should we add a check for that?

hnn_core/network.py
jasmainak: actually never mind this comment, I realized you need to do this because you use the `numpy` array.

hnn_core/network.py
```diff
@@ -14,6 +15,52 @@
...
+    ----------
+    spikes : Spikes
+        An instance of the Spikes object.
+    """
+
+    spiketimes = []
+    spikegids = []
+    spiketypes = []
+    for file in sorted(glob(fname)):
+        spike_trial = np.loadtxt(file, dtype=str)
```
jasmainak, line 42: what do you think of using a convention that singular is a single trial and plural is multiple trails?

So `spike` when it's just one trial and `spikes` when it's multiple trials.

hnn_core/network.py
```diff
@@ -14,6 +15,52 @@
...
+        spiketimes += [list(spike_trial[:, 0].astype(float))]
+        spikegids += [list(spike_trial[:, 1].astype(int))]
```
jasmainak, line 44:
```suggestion
        spike_gids += [list(spike_trial[:, 1].astype(int))]
```

hnn_core/network.py
```diff
@@ -14,6 +15,52 @@
...
+
+        # Note that legacy HNN 'spk.txt' files don't contain a 3rd column for
+        # spike type. If reading a legacy version, validate that a gid_dict is
+        # provided.
+        if spike_trial.shape[1] == 3:
+            spiketypes += [list(spike_trial[:, 2].astype(str))]
+        else:
+            if gid_dict is None:
+                raise ValueError("gid_dict must be provided if spike types "
+                                 "are unspecified in 'spk.txt' file")
```
jasmainak, line 54:
```suggestion
                                 "are unspecified in the file %s" % fname)
```

jasmainak:
Thanks for hanging in there @rythorpe ! I made some more comments. Mostly stylistic, so should be easy to address. We have never been closer!

jasmainak:
Please also update the `api.rst` so the next iteration, hopefully the PR is almost mergeable!

jasmainak:
@cjayb since you have been looking at the code recently, perhaps you might have some comments on this pull request?

hnn_core/network.py
```diff
@@ -523,27 +694,54 @@ def plot_spikes(self, ax=None, show=True):
...
+            where rows correspond to spikes, and columns,
+            delimited by '\\t', correspond to
+            1) spike time (s),
+            2) spike gid, and
+            3) gid type
+        """
+
+        for trial_idx in range(len(self._times)):
+            with open(fname % (trial_idx,), 'w') as f:
+                for spk_idx in range(len(self._times[trial_idx])):
```
jasmainak, line 743:
```suggestion
                for spike_idx in range(len(self._times[trial_idx])):
```

hnn_core/tests/test_network.py
```diff
@@ -28,3 +31,51 @@ def test_network():
...
+        spikes = Spikes(times=[1, 2], gids=spikegids, types=spiketypes)
+    assert "times should be a list of lists" in str(excinfo.value)
+
+    # ValueError should be raised when one of the args is entered as an
+    # incongruent number of trials (e.g., 1 trial when all other args have 2)
```
jasmainak, line 65: avoid too many comments, it's bad practice (cf [Clean code](https://www.investigatii.md/uploads/resurse/Clean_Code.pdf)). Code should be written so that it can be read with comments adding only additional information.

hnn_core/tests/test_network.py
```diff
@@ -28,3 +31,51 @@ def test_network():
     for ev_input in params['t_ev*']:
         type_key = ev_input[2: -2] + ev_input[-1]
         assert len(net.gid_dict[type_key]) == net.n_cells
+
+    # Assert that an empty Spikes object is created as an attribute
+    assert net.spikes == Spikes()
+
+
+def test_spikes():
+    '''Test spikes object.'''
```
jasmainak, line 40:
```suggestion
    """Test spikes object."""
```

hnn_core/network.py
rythorpe: I'm afraid this will be ambiguous for others who haven't been a part of this conversation since the name of the class is `Spikes` (i.e, which includes spikes from either one or multiple trials).

hnn_core/network.py
rythorpe: Oh yeah, my bad.

rythorpe:
@jasmainak I've added `gid_dict` validation for `Spikes().update_types()` so that it checks for overlapping gid ranges. Note that I also shifted `read_spikes()` to create a `Spikes` object and then run `Spikes().update_types()` when supplied with the `gid_dict` arg (i.e., as opposed to having `read_spikes()` run its own code to populate the `Spikes()._types` attribute) to consolidate code. `api.rst` and `whats_new.rst` are updated as well.

hnn_core/network.py
```diff
@@ -14,6 +15,52 @@
...
+        spike_times += [list(spike_trial[:, 0].astype(float))]
+        spike_gids += [list(spike_trial[:, 1].astype(int))]
+
+        # Note that legacy HNN 'spk.txt' files don't contain a 3rd column for
+        # spike type. If reading a legacy version, validate that a gid_dict is
+        # provided.
+        if spike_trial.shape[1] == 3:
+            spike_types += [list(spike_trial[:, 2].astype(str))]
+        else:
+            if gid_dict is None:
```
jasmainak, line 52:
```suggestion
        elif gid_dict is None:
```

sorry I didn't notice this before. Avoid nesting when possible. See:

```sh
$ python -c 'import this' | head -7 | tail -1
Flat is better than nested
```

rythorpe: Oh good point, idk how I missed that. Any other stylistic changes I can make in this batch?

hnn_core/network.py
```diff
@@ -509,8 +554,144 @@ def plot_input(self, ax=None, show=True):
...
+        if not isinstance(other, Spikes):
+            return NotImplemented
+        # Round each time element
+        times_self = [[round(time, 3) for time in trial]
+                      for trial in self._times]
+        times_other = [[round(time, 3) for time in trial]
+                       for trial in other._times]
+        return (times_self == times_other and
+                self._gids == other._gids and
+                self._types == other._types)
```
jasmainak, line 648: I think if you did `np.allclose(times_self, times_other, atol=1e-3) and self._types == other._types` it would do the same thing?

hnn_core/network.py
```diff
@@ -509,8 +554,144 @@ def plot_input(self, ax=None, show=True):
...
+
+        Parameters
+        ----------
+        gid_dict : dict of lists or range objects
+            Dictionary with keys 'evprox1', 'evdist1' etc.
+            containing the range of Cell or input IDs of different
+            cell or input types.
+        """
+
+        # Validate gid_dict
```
jasmainak, line 673:
```suggestion
```

jasmainak:
@rythorpe just two comments. Many tests added, we're going in the right direction!

LGTM once the two comments are addressed

hnn_core/network.py
```diff
@@ -509,8 +554,144 @@ def plot_input(self, ax=None, show=True):
             plt.show()
         return ax.get_figure()
 
-    def plot_spikes(self, ax=None, show=True):
-        """Plot the spiking activity for each cell type.
+
+class Spikes(object):
+    '''The Spikes class.
```
jasmainak, line 559:
```suggestion
    """The Spikes class.
```

hnn_core/network.py
```diff
@@ -509,8 +554,144 @@ def plot_input(self, ax=None, show=True):
...
+    Methods
+    -------
+    update_types(gid_dict)
+        Update spike types in the current instance of Spikes.
+    plot(ax=None, show=True)
+        Plot and return a matplotlib Figure object showing the
+        aggregate network spiking activity according to cell type.
+    write(fname)
+        Write spiking activity to a collection of spike trial files.
+    '''
```
jasmainak, line 600:
```suggestion
    """
```

hnn_core/network.py
rythorpe: I tried it in the beginning but didn't care for the fact that `np.allclose()` uses broadcasting (i.e., it won't control for the case when there has been a formatting discrepancy between `self._times` and `other._times`.

hnn_core/network.py
jasmainak: So, if I do `Spikes(...) == 5` will it return `False` ? Is that how this will work?

This part is not tested is it?

hnn_core/network.py
jasmainak: okay fair enough, we can think  of this later

hnn_core/network.py
jasmainak: humm ... but you won't have any formatting discrepancy now? Because you do input validation.

jasmainak:
That's it from me, let's try to merge this PR as soon as possible so we can move on :)

hnn_core/network.py
rythorpe: There shouldn't be any formatting discrepancies due to inputs, but if a bug were to emerge in `Network` that results in a formatting discrepancy (e.g., if someone were to mess with some square brackets so that the lists append along the first dimension), it would benefit us to have an `__eq__()` function that is independent. What if we use `np.array_equal()`?

jasmainak: ufff right, the Network class. Okay let's leave it like this for now. We can revisit it later.

examples/plot_simulate_evoked.py
```diff
@@ -42,12 +43,19 @@
...
-net.plot_spikes()
+# Also, we can plot the spikes and write them to txt files.
+# Note that we can use formatting syntax to specify the filename pattern
+# with which each trial will be written. To read spikes back in, we can use
+# wildcard expressions.
+with tempfile.TemporaryDirectory() as tmp_dir_name:
+    print(tmp_dir_name)
+    net.spikes.write(op.join(tmp_dir_name, 'spk_%d.txt'))
+    spikes = read_spikes(op.join(tmp_dir_name, 'spk_*.txt'))
+spikes.plot()
```
cjayb, line 58: I get that it's good to demo the use of the `write` method and `read_spikes` function. But I wonder if it confuses the new user (likely to look at this example first). The writing to and reading from disk are not _needed_ for plotting. Maybe both? First `net.spikes.plot()`, then the I/O-bit?

hnn_core/network.py
rythorpe: Correct. A return of `False` is not shown in the tests, but I have tried it on my own and it works. Should I add this as a test?

hnn_core/tests/test_compare_hnn.py
```diff
@@ -36,7 +36,7 @@ def test_hnn_core():
 
     # Test spike type counts
     spiketype_counts = {}
-    for spikegid in net.spikegids[0]:
+    for spikegid in net.spikes.gids[0]:
```
cjayb, line 39: I'm sorry if this is obvious, but what's the difference between `spikes.gids` and `spikes._gids`?

jasmainak: No difference. It's a safety mechanism so that the user doesn't modify `spikes.gids`.

cjayb:
L really GTM, my coding skills don't warrant deeper nit-picks.

jasmainak:
Thanks for taking a look!

@rythorpe the CIs are failing

rythorpe:
@jasmainak it turns out that there was a reason for the nested statement

rythorpe:
@jasmainak last thing, I just added the inputs and outputs section to `api.rst` as you had recommended earlier

rythorpe:
@jasmainak I think we're good to go. Do you want to move forward with the merge?

jasmainak:
Couple of things for future:

Inputs and Outputs should probably be formatted as a subheadheading: https://54-168215891-gh.circle-artifacts.com/0/html/api.html. Also `read_params` should be under it. We could have another subheading for cells. Anyway, future work :) I'm really done with this PR. Let's merge

jasmainak:
Merged, thanks @rythorpe for the contribution!

PR #103, opened 2020-05-28
[MRG] cleanup variable names and logic related to "common" inputs

cjayb:
This PR is an attempt to clarify the code in terms of "external inputs" a.k.a. "external feeds". A lot of the diff is cosmetic, _e.g._, variable name changes. New tests and validations are also included.

One key point is renaming `extinput` to `common`. Our instinct was to go for `rhythmic`, as that's the only thing these are [currently used for](https://jonescompneurolab.github.io/hnn-tutorials/alpha_and_beta/alpha_and_beta). However, the input mechanism is completely generic and could in principle support non-rhythmic inputs as well (_e.g._, Poisson bursts?).

The parameter handling of "common" and "unique" external feeds is still a little opaque in the code, but this should clarify. A future refactor/PR could remove the `ExtFeed` class (and parameter dicts) in favour of functions with signatures that clearly specify what parameters it expects: e.g., t_start, t_std, freq etc

### DIFF
--- a/hnn_core/feed.py
+++ b/hnn_core/feed.py
@@ -22,7 +22,13 @@ class ExtFeed(object):
         'extgauss' : Gaussian-distributed input to proximal dendrites
         'evprox' : Proximal input at specified time (or Gaussian spread)
         'evdist' : Distal input at specified time (or Gaussian spread)
-        'common' : Common inputs to all cells (locations parameter-dependent)
+
+        'common' : As opposed to other feed types, these have timing that is
+        identical (synchronous) for all real cells in the network. Proximal
+        and distal dendrites have separate parameter sets, and need not be
+        synchronous. Note that not all cells classes (types) are required to
+        receive 'common' input---separate conductivity values can be assigned
+        to basket vs. pyramidal cells and AMPA vs. NMDA synapses
     cell_type : str | None
         The cell type, e.g., 'L2_basket', 'L5_pyramidal', etc., or None for
         common inputs (cell population-level weights for common inputs
--- a/hnn_core/feed.py
+++ b/hnn_core/feed.py
@@ -2,65 +2,82 @@
 
 # Authors: Mainak Jas <mainak.jas@telecom-paristech.fr>
 #          Sam Neymotin <samnemo@gmail.com>
+#          Christopher Bailey <bailey.cj@gmail.com>
 
 import numpy as np
 from neuron import h
 
 
 class ExtFeed(object):
-    """"The ExtFeed class.
+    """"The ExtFeed class of external spike input times.
+
+    An external input "feed" to the network, i.e., one that is independent of
+    the spiking output of cells in the network.
 
     Parameters
     ----------
-    ty : str
-        The feed type. Can be:
-        'extpois', 'evprox', 'evdist', 'extgauss', 'extinput'
-    celltype : str | None
-        The cell type.
-    p_ext : dict | list (XXX: check)
-        Parameters of external input.
-        p_ext has a different structure for the extinput
-        usually, p_ext is a dict of cell types
+    feed_type : str
+        The feed type, which is one of
+        'extpois' : Poisson-distributed input to proximal dendrites
+        'extgauss' : Gaussian-distributed input to proximal dendrites
+        'evprox' : Proximal input at specified time (or Gaussian spread)
+        'evdist' : Distal input at specified time (or Gaussian spread)
+        'common' : Common pulses to proximal and distal dendrites
+    cell_type : str | None
+        The cell type, e.g., 'L2_basket', 'L5_pyramidal', etc., or None for
+        common inputs (cell population-level weights for common inputs
+        are defined separately in the parameters)
+    p_ext : dict
+        Parameters of the external input feed, arranged into a dictionary.
     gid : int
         The cell ID.
 
     Attributes
     ----------
+    nrn_EventVec : NEURON hoc.HocObject
+        A vector (list) of event times
+    nrn_VecStim : NEURON hoc.HocObject
+        A VecStim is an artificial spiking cell that generates events at
+        times that are specified in a (NEURON) Vector (see vecevent.mod)
     seed : int
         The seed
     gid : int
         The cell ID
     """
 
-    def __init__(self, ty, celltype, p_ext, gid):
+    def __init__(self, feed_type, cell_type, p_ext, gid):
         # VecStim setup
-        self.eventvec = h.Vector()
-        self.vs = h.VecStim()
-        # self.p_unique = p_unique[type]
+        self.nrn_EventVec = h.Vector()
+        self.nrn_VecStim = h.VecStim()
         self.p_ext = p_ext
-        self.celltype = celltype
-        self.ty = ty  # feed type
+        # used to determine cell type-specific parameters for
+        # (not used for 'common', such as for rhythmic alpha/beta input)
+        self.cell_type = cell_type  # XXX rename for all cell types?
+        self.feed_type = feed_type
         self.gid = gid
         self.set_prng()  # sets seeds for random num generator
-        # sets event times into self.eventvec and plays into self.vs (VecStim)
+        # sets event times into self.nrn_EventVec (Vector)
+        # and plays into self.nrn_VecStim (VecStim)
         self.set_event_times()
 
     def __repr__(self):
         class_name = self.__class__.__name__
-        s = 'seed %d, gid %d' % (self.seed, self.gid)
-        return '<%s | %s>' % (class_name, s)
+        repr_str = "<%s of type '%s' " % (class_name, self.feed_type)
+        repr_str += 'with %d events ' % len(self.nrn_EventVec)
+        repr_str += '| seed %d, gid %d>' % (self.seed, self.gid)
+        return repr_str
 
     def set_prng(self, seed=None):
         if seed is None:  # no seed specified then use p_ext to determine seed
             # random generator for this instance
             # qnd hack to make the seeds the same across all gids
             # for just evoked
-            if self.ty.startswith(('evprox', 'evdist')):
+            if self.feed_type.startswith(('evprox', 'evdist')):
                 if self.p_ext['sync_evinput']:
                     self.seed = self.p_ext['prng_seedcore']
                 else:
                     self.seed = self.p_ext['prng_seedcore'] + self.gid - 2
-            elif self.ty.startswith('extinput'):
+            elif self.feed_type.startswith('common'):
                 # seed for events assuming a given start time
                 self.seed = self.p_ext['prng_seedcore'] + self.gid
                 # separate seed for start times
@@ -74,21 +91,21 @@ class ExtFeed(object):
         self.prng = np.random.RandomState(self.seed)
         if hasattr(self, 'seed2'):
             self.prng2 = np.random.RandomState(self.seed2)
-        # print('ty,seed:',self.ty,self.seed)
+        # print('feed_type,seed:',self.feed_type,self.seed)
 
     def set_event_times(self, inc_evinput=0.0):
         # print('self.p_ext:',self.p_ext)
-        # each of these methods creates self.eventvec for playback
-        if self.ty == 'extpois':
+        # each of these methods creates self.nrn_EventVec for playback
+        if self.feed_type == 'extpois':
             self.__create_extpois()
-        elif self.ty.startswith(('evprox', 'evdist')):
+        elif self.feed_type.startswith(('evprox', 'evdist')):
             self.__create_evoked(inc_evinput)
-        elif self.ty == 'extgauss':
+        elif self.feed_type == 'extgauss':
             self.__create_extgauss()
-        elif self.ty == 'extinput':
-            self.__create_extinput()
+        elif self.feed_type == 'common':
+            self.__create_common_input()
         # load eventvec into VecStim object
-        self.vs.play(self.eventvec)
+        self.nrn_VecStim.play(self.nrn_EventVec)
 
     # based on cdf for exp wait time distribution from unif [0, 1)
     # returns in ms based on lamtha in Hz
@@ -98,13 +115,13 @@ class ExtFeed(object):
     # new external pois designation
     def __create_extpois(self):
         # print("__create_extpois")
-        if self.p_ext[self.celltype][0] <= 0.0 and \
-                self.p_ext[self.celltype][1] <= 0.0:
+        if self.p_ext[self.cell_type][0] <= 0.0 and \
+                self.p_ext[self.cell_type][1] <= 0.0:
             return False  # 0 ampa and 0 nmda weight
         # check the t interval
         t0 = self.p_ext['t_interval'][0]
         T = self.p_ext['t_interval'][1]
-        lamtha = self.p_ext[self.celltype][3]  # index 3 is frequency (lamtha)
+        lamtha = self.p_ext[self.cell_type][3]  # index 3 is frequency (lamtha)
         # values MUST be sorted for VecStim()!
         # start the initial value
         if lamtha > 0.:
@@ -127,15 +144,15 @@ class ExtFeed(object):
         #     print(lamtha, np.mean(xdiff), np.var(xdiff), 1/lamtha**2)
         # Convert array into nrn vector
         # if len(val_pois)>0: print('val_pois:',val_pois)
-        self.eventvec.from_python(val_pois)
-        return self.eventvec.size() > 0
+        self.nrn_EventVec.from_python(val_pois)
+        return self.nrn_EventVec.size() > 0
 
     # mu and sigma vals come from p
     def __create_evoked(self, inc=0.0):
-        if self.celltype in self.p_ext.keys():
+        if self.cell_type in self.p_ext.keys():
             # assign the params
             mu = self.p_ext['t0'] + inc
-            sigma = self.p_ext[self.celltype][3]  # index 3 is sigma_t_ (stdev)
+            sigma = self.p_ext[self.cell_type][3]  # index 3 is sigma_t (stdev)
             numspikes = int(self.p_ext['numspikes'])
             # print('mu:',mu,'sigma:',sigma,'inc:',inc)
             # if a non-zero sigma is specified
@@ -148,20 +165,20 @@ class ExtFeed(object):
             # vals must be sorted
             val_evoked.sort()
             # print('__create_evoked val_evoked:',val_evoked)
-            self.eventvec.from_python(val_evoked)
+            self.nrn_EventVec.from_python(val_evoked)
         else:
             # return an empty eventvec list
-            self.eventvec.from_python([])
-        return self.eventvec.size() > 0
+            self.nrn_EventVec.from_python([])
+        return self.nrn_EventVec.size() > 0
 
     def __create_extgauss(self):
         # assign the params
-        if self.p_ext[self.celltype][0] <= 0.0 and \
-                self.p_ext[self.celltype][1] <= 0.0:
+        if self.p_ext[self.cell_type][0] <= 0.0 and \
+                self.p_ext[self.cell_type][1] <= 0.0:
             return False  # 0 ampa and 0 nmda weight
-        # print('gauss params:',self.p_ext[self.celltype])
-        mu = self.p_ext[self.celltype][3]
-        sigma = self.p_ext[self.celltype][4]
+        # print('gauss params:',self.p_ext[self.cell_type])
+        mu = self.p_ext[self.cell_type][3]
+        sigma = self.p_ext[self.cell_type][4]
         # mu and sigma values come from p
         # one single value from Gaussian dist.
         # values MUST be sorted for VecStim()!
@@ -173,12 +190,15 @@ class ExtFeed(object):
         val_gauss.sort()
         # if len(val_gauss)>0: print('val_gauss:',val_gauss)
         # Convert array into nrn vector
-        self.eventvec.from_python(val_gauss)
-        return self.eventvec.size() > 0
+        self.nrn_EventVec.from_python(val_gauss)
+        return self.nrn_EventVec.size() > 0
 
-    def __create_extinput(self):
-        """Creates the ongoing external inputs (rhythmic)."""
-        # print("__create_extinput")
+    def __create_common_input(self):
+        """Creates the common ongoing external inputs.
+        
+        Used for, e.g., for rhythmic inputs in alpha/beta generation
+        """
+        # print("__create_common_input")
         # Return if all synaptic weights are 0
         all_syn_weights_zero = True
         for key in self.p_ext.keys():
@@ -254,15 +274,15 @@ class ExtFeed(object):
             t_input.sort()
         else:
             print("Indicated distribution not recognized. "
-                  "Not making any alpha feeds.")
+                  "Not making any common feeds.")
             t_input = []
         # Convert array into nrn vector
-        self.eventvec.from_python(t_input)
-        return self.eventvec.size() > 0
+        self.nrn_EventVec.from_python(t_input)
+        return self.nrn_EventVec.size() > 0
 
     # for parallel, maybe be that postsyn for this is just nil (None)
     def connect_to_target(self, threshold):
         # print("connect_to_target")
-        nc = h.NetCon(self.vs, None)  # why is target always nil??
+        nc = h.NetCon(self.nrn_VecStim, None)  # why is target always nil??
         nc.threshold = threshold
         return nc
--- a/hnn_core/feed.py
+++ b/hnn_core/feed.py
@@ -9,7 +9,7 @@ from neuron import h
 
 
 class ExtFeed(object):
-    """"The ExtFeed class of external spike input times.
+    """The ExtFeed class of external spike input times.
 
     An external input "feed" to the network, i.e., one that is independent of
     the spiking output of cells in the network.
@@ -22,7 +22,7 @@ class ExtFeed(object):
         'extgauss' : Gaussian-distributed input to proximal dendrites
         'evprox' : Proximal input at specified time (or Gaussian spread)
         'evdist' : Distal input at specified time (or Gaussian spread)
-        'common' : Common pulses to proximal and distal dendrites
+        'common' : Common inputs to all cells (locations parameter-dependent)
     cell_type : str | None
         The cell type, e.g., 'L2_basket', 'L5_pyramidal', etc., or None for
         common inputs (cell population-level weights for common inputs
@@ -34,9 +34,9 @@ class ExtFeed(object):
 
     Attributes
     ----------
-    nrn_EventVec : NEURON hoc.HocObject
-        A vector (list) of event times
-    nrn_VecStim : NEURON hoc.HocObject
+    nrn_eventvec : instance of NEURON Vector
+        A vector of event times
+    nrn_vecstim : instance of NEURON VecStim
         A VecStim is an artificial spiking cell that generates events at
         times that are specified in a (NEURON) Vector (see vecevent.mod)
     seed : int
@@ -47,8 +47,8 @@ class ExtFeed(object):
 
     def __init__(self, feed_type, cell_type, p_ext, gid):
         # VecStim setup
-        self.nrn_EventVec = h.Vector()
-        self.nrn_VecStim = h.VecStim()
+        self.nrn_eventvec = h.Vector()
+        self.nrn_vecstim = h.VecStim()
         self.p_ext = p_ext
         # used to determine cell type-specific parameters for
         # (not used for 'common', such as for rhythmic alpha/beta input)
@@ -56,14 +56,14 @@ class ExtFeed(object):
         self.feed_type = feed_type
         self.gid = gid
         self.set_prng()  # sets seeds for random num generator
-        # sets event times into self.nrn_EventVec (Vector)
-        # and plays into self.nrn_VecStim (VecStim)
+        # sets event times into self.nrn_eventvec (Vector)
+        # and plays into self.nrn_vecstim (VecStim)
         self.set_event_times()
 
     def __repr__(self):
         class_name = self.__class__.__name__
         repr_str = "<%s of type '%s' " % (class_name, self.feed_type)
-        repr_str += 'with %d events ' % len(self.nrn_EventVec)
+        repr_str += 'with %d events ' % len(self.nrn_eventvec)
         repr_str += '| seed %d, gid %d>' % (self.seed, self.gid)
         return repr_str
 
@@ -94,27 +94,39 @@ class ExtFeed(object):
         # print('feed_type,seed:',self.feed_type,self.seed)
 
     def set_event_times(self, inc_evinput=0.0):
-        # print('self.p_ext:',self.p_ext)
-        # each of these methods creates self.nrn_EventVec for playback
+
+        # check feed name validity, allowing substring matches ('evprox1' etc)
+        valid_feeds = ['extpois', 'extgauss', 'common', 'evprox', 'evdist']
+        # NB check if self.feed_type has a valid substring, not vice versa
+        matches = [f for f in valid_feeds if f in self.feed_type]
+        if len(matches) == 0:
+            raise ValueError('Invalid external feed: %s' % self.feed_type)
+        elif len(matches) > 1:
+            raise ValueError('Ambiguous external feed: %s' % self.feed_type)
+
+        # Each of these methods creates self.nrn_eventvec for playback
+        # Return values not checked: False if all weights for given feed type
+        # are zero. Designed to be silent so that zeroing input weights
+        # effectively disables each.
         if self.feed_type == 'extpois':
-            self.__create_extpois()
+            self._create_extpois()
         elif self.feed_type.startswith(('evprox', 'evdist')):
-            self.__create_evoked(inc_evinput)
+            self._create_evoked(inc_evinput)
         elif self.feed_type == 'extgauss':
-            self.__create_extgauss()
+            self._create_extgauss()
         elif self.feed_type == 'common':
-            self.__create_common_input()
+            self._create_common_input()
         # load eventvec into VecStim object
-        self.nrn_VecStim.play(self.nrn_EventVec)
+        self.nrn_vecstim.play(self.nrn_eventvec)
 
     # based on cdf for exp wait time distribution from unif [0, 1)
     # returns in ms based on lamtha in Hz
-    def __t_wait(self, lamtha):
+    def _t_wait(self, lamtha):
         return -1000. * np.log(1. - self.prng.rand()) / lamtha
 
     # new external pois designation
-    def __create_extpois(self):
-        # print("__create_extpois")
+    def _create_extpois(self):
+        # print("_create_extpois")
         if self.p_ext[self.cell_type][0] <= 0.0 and \
                 self.p_ext[self.cell_type][1] <= 0.0:
             return False  # 0 ampa and 0 nmda weight
@@ -125,7 +137,7 @@ class ExtFeed(object):
         # values MUST be sorted for VecStim()!
         # start the initial value
         if lamtha > 0.:
-            t_gen = t0 + self.__t_wait(lamtha)
+            t_gen = t0 + self._t_wait(lamtha)
             val_pois = np.array([])
             if t_gen < T:
                 np.append(val_pois, t_gen)
@@ -133,7 +145,7 @@ class ExtFeed(object):
             # sort
             while t_gen < T:
                 # so as to not clobber confusingly base off of t_gen ...
-                t_gen += self.__t_wait(lamtha)
+                t_gen += self._t_wait(lamtha)
                 if t_gen < T:
                     val_pois = np.append(val_pois, t_gen)
         else:
@@ -144,11 +156,11 @@ class ExtFeed(object):
         #     print(lamtha, np.mean(xdiff), np.var(xdiff), 1/lamtha**2)
         # Convert array into nrn vector
         # if len(val_pois)>0: print('val_pois:',val_pois)
-        self.nrn_EventVec.from_python(val_pois)
-        return self.nrn_EventVec.size() > 0
+        self.nrn_eventvec.from_python(val_pois)
+        return self.nrn_eventvec.size() > 0
 
     # mu and sigma vals come from p
-    def __create_evoked(self, inc=0.0):
+    def _create_evoked(self, inc=0.0):
         if self.cell_type in self.p_ext.keys():
             # assign the params
             mu = self.p_ext['t0'] + inc
@@ -164,14 +176,14 @@ class ExtFeed(object):
             val_evoked = val_evoked[val_evoked > 0]
             # vals must be sorted
             val_evoked.sort()
-            # print('__create_evoked val_evoked:',val_evoked)
-            self.nrn_EventVec.from_python(val_evoked)
+            # print('_create_evoked val_evoked:',val_evoked)
+            self.nrn_eventvec.from_python(val_evoked)
         else:
             # return an empty eventvec list
-            self.nrn_EventVec.from_python([])
-        return self.nrn_EventVec.size() > 0
+            self.nrn_eventvec.from_python([])
+        return self.nrn_eventvec.size() > 0
 
-    def __create_extgauss(self):
+    def _create_extgauss(self):
         # assign the params
         if self.p_ext[self.cell_type][0] <= 0.0 and \
                 self.p_ext[self.cell_type][1] <= 0.0:
@@ -190,15 +202,15 @@ class ExtFeed(object):
         val_gauss.sort()
         # if len(val_gauss)>0: print('val_gauss:',val_gauss)
         # Convert array into nrn vector
-        self.nrn_EventVec.from_python(val_gauss)
-        return self.nrn_EventVec.size() > 0
+        self.nrn_eventvec.from_python(val_gauss)
+        return self.nrn_eventvec.size() > 0
 
-    def __create_common_input(self):
+    def _create_common_input(self):
         """Creates the common ongoing external inputs.
-        
+
         Used for, e.g., for rhythmic inputs in alpha/beta generation
         """
-        # print("__create_common_input")
+        # print("_create_common_input")
         # Return if all synaptic weights are 0
         all_syn_weights_zero = True
         for key in self.p_ext.keys():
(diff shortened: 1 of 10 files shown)


hnn_core/feed.py
```diff
@@ -173,12 +190,15 @@ def __create_extgauss(self):
...
         # Convert array into nrn vector
-        self.eventvec.from_python(val_gauss)
-        return self.eventvec.size() > 0
+        self.nrn_EventVec.from_python(val_gauss)
+        return self.nrn_EventVec.size() > 0
 
-    def __create_extinput(self):
-        """Creates the ongoing external inputs (rhythmic)."""
-        # print("__create_extinput")
+    def __create_common_input(self):
```
jasmainak, line 196: I did this already but not completely. We should avoid methods to have double underscores. It mangles the function names. Just single underscore is good for a private method

hnn_core/network.py
```diff
@@ -76,27 +76,27 @@ def __init__(self, params, n_jobs=1):
...
+        # params of common external feeds inputs in p_common
         # Global number of external inputs ... automatic counting
         # makes more sense
         # p_unique represent ext inputs that are going to go to each cell
-        self.p_ext, self.p_unique = create_pext(self.params,
-                                                self.params['tstop'])
-        self.N_extinput = len(self.p_ext)
+        self.p_common, self.p_unique = create_pext(self.params,
+                                                   self.params['tstop'])
+        self.N_common_feeds = len(self.p_common)
```
jasmainak, line 85:
```suggestion
        self.n_common_feeds = len(self.p_common)
```
like in MNE and other packages. All small letters.

hnn_core/network.py
```diff
@@ -285,10 +284,10 @@ def _gid_assign(self):
                 pc.set_gid2node(gid_input, rank)
                 self._gid_list.append(gid_input)
         # legacy handling of the external inputs
-        # NOT perfectly balanced for now
-        for gid_base in range(rank, self.N_extinput, nhosts):
-            # shift the gid_base to the extinput gid
-            gid = gid_base + self.gid_dict['extinput'][0]
+        # NOT perfectly balanced for now  XXX wat?
```
jasmainak, line 287: just delete this comment if it makes no sense. We can always look at HNN or check an older commit. I guess it's helpful to nobody to have comments that make no sense.

hnn_core/network.py
```diff
@@ -371,25 +383,26 @@ def _parnet_connect(self):
         from .parallel import pc
 
         # loop over target zipped gids and cells
-        # cells has NO extinputs anyway. also no extgausses
+        # cells has NO common feeds anyway. also no extgausses
+        # XXX no comprendo commento
```
jasmainak, line 387: lol, this is why I just remove comments unless they really help comprehension :)

hnn_core/feed.py
```diff
@@ -127,15 +144,15 @@ def __create_extpois(self):
         #     print(lamtha, np.mean(xdiff), np.var(xdiff), 1/lamtha**2)
         # Convert array into nrn vector
         # if len(val_pois)>0: print('val_pois:',val_pois)
-        self.eventvec.from_python(val_pois)
-        return self.eventvec.size() > 0
+        self.nrn_EventVec.from_python(val_pois)
```
jasmainak, line 147: variable names should be in snake case unless they are classes!

hnn_core/feed.py
```diff
@@ -2,65 +2,82 @@
...
+        are defined separately in the parameters)
+    p_ext : dict
+        Parameters of the external input feed, arranged into a dictionary.
     gid : int
         The cell ID.
 
     Attributes
     ----------
+    nrn_EventVec : NEURON hoc.HocObject
+        A vector (list) of event times
```
jasmainak, line 38: this is an object from Neuron right? It's a peculiar object that's unlike a list. I would just name the actual object or remove the (list)

hnn_core/network.py
```diff
@@ -315,44 +314,57 @@ def _create_all_src(self):
...
+                elif cell_type == 'common':
+                    # print('cell_type',cell_type)
                     # to find param index, take difference between REAL gid
                     # here and gid start point of the items
-                    p_ind = gid - self.gid_dict['extinput'][0]
+                    p_ind = gid - self.gid_dict['common'][0]
+
+                    # new ExtFeed: cell type irrelevant (None) since input
+                    # timing is unaffected by it; cell_type == feed_type
+                    self.common_feed_list.append(
```
jasmainak, line 345:
```suggestion
                    self.common_feeds.append(
```
It's a bit superfluous to say common_feed_list when you see `append`

hnn_core/tests/test_feed.py
```diff
@@ -5,7 +5,23 @@
 import os.path as op
 
 import hnn_core
-from hnn_core import read_params, Network
+from hnn_core import read_params, Network, Params
+from hnn_core.feed import ExtFeed
+from hnn_core.params import create_pext
+
+
+def test_ExtFeed_object():
```
jasmainak, line 13:
```suggestion
def test_extfeed():
```

hnn_core/feed.py
```diff
@@ -2,65 +2,82 @@
...
-        Parameters of external input.
-        p_ext has a different structure for the extinput
-        usually, p_ext is a dict of cell types
+    feed_type : str
+        The feed type, which is one of
+        'extpois' : Poisson-distributed input to proximal dendrites
+        'extgauss' : Gaussian-distributed input to proximal dendrites
+        'evprox' : Proximal input at specified time (or Gaussian spread)
+        'evdist' : Distal input at specified time (or Gaussian spread)
+        'common' : Common pulses to proximal and distal dendrites
```
jasmainak, line 25: Nice thanks! How did you figure out that it's "common"?

I think we should probably get rid of this class and separate these methods into functions: `create_evoked`, `create_uniform` with function signatures that clearly specify what parameters it expects: e.g., t_start, t_std, freq etc.

jasmainak: But this could be another PR

rythorpe: I was under the impression that what we previously called 'extinput' was a common input to all neurons in at the soma (i.e., a modification of the proximal input) rather than to both the proximal and distal input levels. Is this wrong?

hnn_core/tests/test_feed.py
```diff
@@ -5,7 +5,23 @@
...
+    """Test the different external feeds."""
+
+    params = Params()
+    p_common, p_unique = create_pext(params,
+                                     params['tstop'])
+    for feed_type in ['extpois', 'extgauss']:
+        ef = ExtFeed(feed_type, 'L2_basket', p_unique[feed_type], 0)
```
jasmainak, line 20:
```suggestion
        feed = ExtFeed(feed_type, 'L2_basket', p_unique[feed_type], 0)
```

hnn_core/feed.py
```diff
@@ -2,65 +2,82 @@
...
 #          Sam Neymotin <samnemo@gmail.com>
+#          Christopher Bailey <bailey.cj@gmail.com>
 
 import numpy as np
 from neuron import h
 
 
 class ExtFeed(object):
-    """"The ExtFeed class.
+    """"The ExtFeed class of external spike input times.
```
jasmainak, line 12:
```suggestion
    """The ExtFeed class of external spike input times.
```

hnn_core/feed.py
```diff
@@ -74,21 +91,21 @@ def set_prng(self, seed=None):
...
             self.prng2 = np.random.RandomState(self.seed2)
-        # print('ty,seed:',self.ty,self.seed)
+        # print('feed_type,seed:',self.feed_type,self.seed)
 
     def set_event_times(self, inc_evinput=0.0):
         # print('self.p_ext:',self.p_ext)
-        # each of these methods creates self.eventvec for playback
-        if self.ty == 'extpois':
+        # each of these methods creates self.nrn_EventVec for playback
+        if self.feed_type == 'extpois':
```
jasmainak, line 99: let's do some input validation! If the feed_type is not one of the recognized choices, raise a  ValueError

hnn_core/feed.py
```diff
@@ -254,15 +274,15 @@ def __create_extinput(self):
...
-        self.eventvec.from_python(t_input)
-        return self.eventvec.size() > 0
+        self.nrn_EventVec.from_python(t_input)
+        return self.nrn_EventVec.size() > 0
 
     # for parallel, maybe be that postsyn for this is just nil (None)
     def connect_to_target(self, threshold):
         # print("connect_to_target")
-        nc = h.NetCon(self.vs, None)  # why is target always nil??
+        nc = h.NetCon(self.nrn_VecStim, None)  # why is target always nil??
```
jasmainak, line 286:
```suggestion
        nc = h.NetCon(self.nrn_VecStim, None)  # why is target always None??
```

hnn_core/params.py
```diff
@@ -323,9 +322,9 @@ def create_pext(p, tstop):
     }
 
     # ensures time interval makes sense
-    p_ext = _feed_validate(p_ext, feed_prox, tstop)
+    p_common = _feed_validate(p_common, feed_prox, tstop)
```
jasmainak, line 325:
```suggestion
    p_prox = _feed_validate(p_prox, feed_prox, tstop)
```

? I'm a bit confused btw this common and prox thing

hnn_core/network.py
cjayb: Sure, but reason was to distinguish it from `unique_feed_dict`, which is a `dict`! It's quite confusing that there are two containers for what seems to be similar things. I thought it helped around L345 and L363 to keep tabs on what's being sent to `ExtFeed`. I'll take another look though

hnn_core/feed.py
cjayb: It took me some time to wrap my head around this! Create the `net` for the `alpha` example in `master` & look at `net.p_ext` (renamed in this PR): it's a `len==2` list, where each dict has the key `'loc'` being `'distal'` or `'proximal'`.

Figure 3 of the [alpha/beta tutorial](https://jonescompneurolab.github.io/hnn-tutorials/alpha_and_beta/alpha_and_beta) has the schematic. I didn't actually follow the code (yet) to confirm this, but there's a lot of 'evidence' to these being dendritic inputs for pyramidals & somatic for baskets, though for some reason [L5 baskets are not included in the parameter list](https://github.com/cjayb/hnn-core/blob/822d23945ff2f0907b28896d1d329be6bd271742/hnn_core/params.py#L328). Separate investigation/PR...

hnn_core/params.py
cjayb: -1 the function `_feed_validate` appends to the list `p_common` so that both distal and proximal "common" inputs end up in the same parameter _list_. Note that `p_unique` in the same function (`create_pext`) has always been a `dict`... See also comment above regarding the 'commonness' of what we all thought were somehow specific to 'rhythmic'.

hnn_core/feed.py
cjayb: I interpret this as variables names should NOT be in snake unless ...

hnn_core/network.py
cjayb: It should be OK

hnn_core/feed.py
rythorpe: Let me correct my previous statement: an 'extinput' (or what you're labelling 'common') can either be at the proximal location, distal location, or both. The distinguishing factor when comparing it to a proximal evoked input, for example, is that the 'extinput' at the proximal location treats the feed as a single input directly to all L5 pyramidal cells. Evoked inputs, on the other hand, create a separate feed for each 'proximal' cell (i.e., one for each of the 10x10 L5 pyramidal cells, the 10x10 L2/3 pyramidal cells, the 35 L5 basket cells and 35 L2/3 basket cells) dependent on the local network connectivity. @samnemo is this correct?

@cjayb Maybe this is what you were already thinking, I just want to make sure we find a clear way to represent this in the code so that it is less confusing in the future.

cjayb: @rythorpe that’s exactly my understanding too, yes. Whether or not ‘common’ is a good term is debatable. I think the way params are handled (list of dict in one place, dict of dict in another) is a little muddy still. Perhaps @jasmainak ‘s suggestion of refactoring into functions rather than an object (which hardly uses statefulness as it is anyway) will help.

jasmainak: I see, so "common" is somehow opposed to "unique"? Could you clarify above in the docstring that it can be either proximal or distal?

also are you able to tell [based on the code](https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/network.py#L336-L356) how one is unique to each cell and the other is not?

The list of dict and dict of dict is indeed an issue. Let's open an issue for that so we can understand it better? I would begin by documenting what exactly these data structures do.

hnn_core/network.py
```diff
@@ -315,45 +313,58 @@ def _create_all_src(self):
...
+                    # print('cell_type',cell_type)
                     # to find param index, take difference between REAL gid
                     # here and gid start point of the items
-                    p_ind = gid - self.gid_dict['extinput'][0]
+                    p_ind = gid - self.gid_dict['common'][0]
+
+                    # new ExtFeed: cell type irrelevant (None) since input
+                    # timing is unaffected by it; cell_type == feed_type
+                    self.common_feeds.append(
+                        ExtFeed(cell_type, None, self.p_common[p_ind], gid))
```
jasmainak, line 345: using keyword arguments might be clearer here so we know what is `None` without having to look at class signature

jasmainak:
@cjayb did you verify that the docs build alright and the examples don't change?

jasmainak:
@cjayb if you address the last couple of comments, I'll merge the PR. Thanks a lot!

hnn_core/feed.py
rythorpe: > also are you able to tell [based on the code](https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/network.py#L336-L356) how one is unique to each cell and the other is not?
> 
I find the code to be fairly opaque in this regard.

cjayb:
See if these help, there's surely plenty more for a future PR too ;)

I can confirm that the `auto_examples` in docs build (with some warnings)

jasmainak:
@cjayb something is broken. I'm not sure if the alpha example was completely fixed in the previous PR or if it's a different issue.

I see this:
sphx_glr_plot_simulate_alpha_001

which is different from what's on the website:

website

cjayb:
Bugger... I'll have to investigate. It was still working before (a82f20f3a922a) I came in to fix the bug.

jasmainak:
thank you! may the force be with you

jasmainak:
@cjayb you inspired me to finish setting up the Circle CI. If you rebase against master, now you should see it building. This is what I got from the last build:

https://41-168215891-gh.circle-artifacts.com/0/html/auto_examples/index.html

hnn_core/feed.py
```diff
@@ -173,12 +207,15 @@ def __create_extgauss(self):
...
+
+    def _create_common_input(self):
+        """Creates the common ongoing external inputs.
 
-    def __create_extinput(self):
-        """Creates the ongoing external inputs (rhythmic)."""
-        # print("__create_extinput")
+        Used for, e.g., for rhythmic inputs in alpha/beta generation
+        """
+        # print("_create_common_input")
```
jasmainak, line 218: you can remove this bogus print statement.

hnn_core/feed.py
```diff
@@ -2,65 +2,87 @@
...
         # VecStim setup
-        self.eventvec = h.Vector()
-        self.vs = h.VecStim()
-        # self.p_unique = p_unique[type]
-        self.p_ext = p_ext
-        self.celltype = celltype
-        self.ty = ty  # feed type
+        self.nrn_eventvec = h.Vector()
+        self.nrn_vecstim = h.VecStim()
+        self.p_ext = params
```
jasmainak, line 57: also, any reason to continue calling this `p_ext` in the attribute?

cjayb:
First of all, sorry about the git-mess: I managed to make a mess of my tree in the rebase...

I found the problem (fixed in 
e26d98b), which was introduced into `params.py` with the bug I originally set out to fix (94fc21a).

But I also had to undo b2036e7 , which made `test_compare_hnn` fail due to mismatch with `test_data/dpl.txt`.

@rythorpe WDYT? I've basically rolled back what you set out to do in #98, and fixed the test you fixed by rolling back yet another commit! I'm sure I don't know whether anything makes sense anymore... Are we sure our gold standard is golden?

I think `params.py:create_pext()` and `_feed_validate()` are killing us. I'll open an issue to start discussing how to refactor...

hnn_core/tests/test_network.py
```diff
@@ -22,7 +22,7 @@ def test_network():
     print(net.cells[:2])
 
     # Assert that proper number of gids are created for Network inputs
-    assert len(net.gid_dict['common']) == 2
+    assert len(net.gid_dict['common']) == 0
```
rythorpe, line 25: My concern here is that it is nice for user sanity to see that a network object has created gids for `common` inputs without any `common` input spikes occurring when running a simulation. This is consistent with the current convention where e.g. gids are created for `extgauss` and `extpois` regardless of their actual activation in a dipole simulation.

cjayb:
yeah I figured the -2 was there for such reasons. TBH I was a little surprised that removing it allowed `test_compare_hnn` to pass... Can we conclude that all is OK then? Alpha seems to be working again, and the dipoles match.

jasmainak:
I can confirm that the alpha simulation works now:

https://13-265240816-gh.circle-artifacts.com/0/html/auto_examples/plot_simulate_alpha.html#sphx-glr-auto-examples-plot-simulate-alpha-py

Let me check the rest of the code @cjayb !

hnn_core/feed.py
```diff
@@ -2,121 +2,153 @@
...
-        self.vs = h.VecStim()
-        # self.p_unique = p_unique[type]
-        self.p_ext = p_ext
-        self.celltype = celltype
-        self.ty = ty  # feed type
+        self.nrn_eventvec = h.Vector()
+        self.nrn_vecstim = h.VecStim()
+        self.params = params
+        # used to determine cell type-specific parameters for
+        # (not used for 'common', such as for rhythmic alpha/beta input)
```
jasmainak, line 59: phew, nasty!

hnn_core/params.py
```diff
@@ -190,48 +190,49 @@ def _feed_validate(p_ext, p_ext_d, tstop):
...
-        for key in p_ext_d.keys():
-            if key.startswith('L5'):
-                if p_ext_d['L2Pyr'][1] != 0.1:
-                    p_ext_d[key] = (p_ext_d[key][0], p_ext_d['L2Pyr'][1])
-                else:
-                    p_ext_d[key] = (p_ext_d[key][0], 1.)
-
-    p_ext.append(p_ext_d)
+    # only append if t0 is less than simulation tstop
+    if tstop > p_ext_d['t0']:
```
jasmainak, line 206: is this a rebase issue or was this undone in your previous PR somehow? I seem to recall seeing this diff in @rythorpe 's PR

hnn_core/network.py
```diff
@@ -311,55 +322,75 @@ def _create_all_src(self):
...
+                            params=self.p_unique[src_type],
+                            gid=gid))
+                pc.cell(gid,
+                        self.unique_feeds[src_type][-1].connect_to_target(
+                            self.params['threshold']))
             else:
-                print("None of these types in Net()")
-                exit()
+                raise ValueError('No parameters specified for external feed '
+                                 'type: %s' % src_type)
```
jasmainak, line 393: Excellent!

We should add a test too, but I'm fine leaving this for another PR. The diff is pretty big already.

hnn_core/network.py
```diff
@@ -311,55 +322,75 @@ def _create_all_src(self):
...
-            if pc.gid_exists(gid):
-                # get type of cell and pos via gid
-                # now should be valid for ext inputs
-                type = self.gid_to_type(gid)
-                type_pos_ind = gid - self.gid_dict[type][0]
-                pos = self.pos_dict[type][type_pos_ind]
+            if not pc.gid_exists(gid):
+                msg = ('Source of type %s with ID %d does not exists in '
+                       'Network' % (src_type, gid))
+                raise RuntimeError(msg)
```
jasmainak, line 332: Nice again! But we'll need a test at some point :-)

hnn_core/network.py
```diff
@@ -301,7 +299,20 @@ def gid_to_type(self, gid):
...
+    def _get_src_type_and_pos(self, gid):
+        """Source type, position and whether it's a cell or artificial feed"""
+
+        # get type of cell and pos via gid
+        src_type = self.gid_to_type(gid)
+        type_pos_ind = gid - self.gid_dict[src_type][0]
+        src_pos = self.pos_dict[src_type][type_pos_ind]
+
+        real_cell_types = ['L2_pyramidal', 'L5_pyramidal',
+                           'L2_basket', 'L5_basket']
```
jasmainak, line 311: this should probably be a global if it's used in other functions but I think we can leave it here for now.

hnn_core/network.py
```diff
@@ -311,55 +322,75 @@ def _create_all_src(self):
...
+                    self.cells.append(Cell(gid, src_pos, self.params))
                 else:
-                    print("None of these types in Net()")
-                    exit()
+                    self.cells.append(Cell(gid, src_pos))
+
+                pc.cell(gid, self.cells[-1].connect_to_target(
+                        None, self.params['threshold']))
+
+            # external inputs are special types of pseudo-cells
```
jasmainak, line 349:
```suggestion
            # external inputs are special types of artificial cells
```
better to be consistent. It's very easy to confuse newcomers!

hnn_core/network.py
```diff
@@ -311,55 +322,75 @@ def _create_all_src(self):
...
+            # 'common': all cells impacted with identical TIMING of spike
+            # events. NB: cell types can still have different weights for how
+            # such 'common' spikes influence them
+            elif src_type == 'common':
+                # print('cell_type',cell_type)
```
jasmainak, line 354:
```suggestion
```

hnn_core/network.py
```diff
@@ -311,55 +322,75 @@ def _create_all_src(self):
...
+                # to find param index, take difference between REAL gid
+                # here and gid start point of the items
+                p_ind = gid - self.gid_dict['common'][0]
+
+                # new ExtFeed: target cell type irrelevant (None) since input
+                # timing will be identical for all cells
+                # XXX common_feeds is a list of dict (comp. unique_feeds)
```
jasmainak, line 361: what's comp. ? compare against? I would just delete that part in the brackets for now. We'll need to clarify it somewhere more prominent.

hnn_core/network.py
```diff
@@ -311,55 +322,75 @@ def _create_all_src(self):
...
+                self.common_feeds.append(
+                    ExtFeed(feed_type=src_type,
+                            target_cell_type=None,
+                            params=self.p_common[p_ind],
+                            gid=gid))
+
+                # now use the param index in the params and create
```
jasmainak, line 368: param index?

jasmainak:
regarding the random state, I think we should set the random state to be `None` by default (so they start at a truly random value) and then in the tests and examples, set them to match what `HNN-GUI` does. It's what we do in MNE and other packages. The `random_state` should not matter _that_ much although it will certainly change the output. @rythorpe what do you think?

jasmainak:
but let's reserve random state changes for another PR

jasmainak:
@rythorpe is the PR good to go for you?

samnemo:
I wouldn't do that. People who run neuronal network simulations (which are pretty complex) want the same result each time and to have control over it.

hnn_core/params.py
rythorpe: This was undone from my previous PR and is evident in the change I reviewed above in `test_network.py`. Even though this appears to fix the discrepancy with the alpha simulation, I don't think this is a good solution because it results in no `common` input gids being created in the network object.

rythorpe:
> @rythorpe is the PR good to go for you?

No, see my comment above. I think we need to resolve the problem without undoing #98 since it affects the Network.gid_dict attribute.

jasmainak:
> I wouldn't do that. People who run neuronal network simulations (which are pretty complex) want the same result each time and to have control over it.

I see, I am not suggesting removing the option of setting random state. However, it does seem a bit odd to me that you would pick one random state over another for _all_ users. This also makes me question how we can know for sure that the optimization routines were initialized correctly. cc @blakecaldwell . Let's reserve this for another issue or internal discussion!

cjayb:
> @cjayb it sounds like the problem you encountered with the alpha example was caused by a seed discrepancy introduced by #98. Is that correct?

Nope, the problem was 1st that _no_ common ('external') feeds were being included, ever. I thought it that simple, but the change to `param.py` was also causing a tripup: the alpha example is distal-only, where the proximal `t0` is set to `>'tstop'`. The random seed is just a minor detail (doesn't explain the core problem this PR was trying to address).

I don't understand how your fix in #98 caused the alpha example to fail. Based on this discussion, I'd like to take a little longer to dig into what is going on. We might as well get this right... There's something really strange about proximal inputs affecting the alpha-example, even when all weights are zero! Hopefully I can sort the evoked example out as well.

rythorpe:
> I don't understand how your fix in #98 caused the alpha example to fail. Based on this discussion, I'd like to take a little longer to dig into what is going on. We might as well get this right... There's something really strange about proximal inputs affecting the alpha-example, even when all weights are zero! Hopefully I can sort the evoked example out as well.

This is certainly perplexing. I'll try to take a closer look at it as well. Thanks for working on this @cjayb. Your changes to the treatment of "common" inputs were much needed!

hnn_core/tests/test_network.py
jasmainak: It's a bit weird though that you have inputs artificial cells for every imaginable input already in the network. I would think that you want to add input cells on a need basis.

jasmainak:
Hey guys, unless there is a trivial fix for the issue @rythorpe raised, I propose adding an XXX comment to the test saying what should pass and then addressing it in the next PR. I think that problem neatly fits into the issue of refactoring the `ExtInput` class and making everything consistent in that class.

This PR is pretty big already. It does improve things and fixes the alpha example, so I would call it progress.

hnn_core/tests/test_network.py
cjayb: +1 for per-need inclusion of inputs into `net`, which the `gid_dict` would then reflect.

Resolving here, but making note in #104 where it seems to fit nicely.

hnn_core/network.py
cjayb: I don't understand that comment...

jasmainak:
Let me know when it's good to go @cjayb . Please set the PR status to [MRG] if you're ready.

cjayb:
The remaining great mystery is this: why does the alpha-example [subtly fail](https://github.com/jonescompneurolab/hnn-core/pull/103#issuecomment-636353522) when proximal common feeds are included in the `gid_list`, even though all (AMPA) weights are zero?

My money is on the way `parreceive` (and `parconnect_from_src`) is called for these inputs [in the receiving pyramidal cells](https://github.com/cjayb/hnn-core/blob/f1f257f495917b67ef04c694f42b43f7222442ed/hnn_core/pyramidal.py#L876). Even though the weight is zero, the delay-value is not, which might lead to [some wonky in the `NetCon` object](https://github.com/cjayb/hnn-core/blob/f1f257f495917b67ef04c694f42b43f7222442ed/hnn_core/cell.py#L373). But that's beyond my current limits of understanding...

> I think we need to resolve the problem without completely undoing parts of #98 since it affects the Network.gid_dict attribute.

@rythorpe I see your concern, but don't you think `network.py:_get_src_type_and_pos()` should gracefully handle some source types missing?

```python
        # get type of cell and pos via gid
        src_type = self.gid_to_type(gid)
        type_pos_ind = gid - self.gid_dict[src_type][0]
        src_pos = self.pos_dict[src_type][type_pos_ind]
```

In any case, if `dpl.txt` is the gold standard, then our passing tests should mean we're OK for now, right?

@jasmainak assuming the last cosmetic commits don't break anything (!), I'm go for merge.

jasmainak:
I'm now wondering how the parameter file handles the external inputs. Do you need to have all the parameters present for all the inputs? Are certain inputs turned off by making the weights = 0? Do you know @rythorpe ?

Yeah, let's merge this PR to keep the momentum and discuss more in a following PR.

jasmainak:
It's in, thanks for the great contribution @cjayb !

rythorpe:
> @rythorpe I see your concern, but don't you think `network.py:_get_src_type_and_pos()` should gracefully handle some source types missing?
> 
> ```python
>         # get type of cell and pos via gid
>         src_type = self.gid_to_type(gid)
>         type_pos_ind = gid - self.gid_dict[src_type][0]
>         src_pos = self.pos_dict[src_type][type_pos_ind]
> ```
@cjayb yes I agree, but we just have to be consistent. As a user I'd like to see either all possible gids initialized in a network object or only the applicable ones.

cjayb:
> Is the plan to just readdress the issue I encountered in #98 in a separate PR?
I thought the test you included (dpl.txt) was checking for the issue you were dealing with in #98. So because this PR also passes the (same) test, the issue should now be solved, no?

As for GID handling, I totally agree care should be taken in the next round of parameter handling cleanups!

@jasmainak currently the default way to ‘turn off’ an input is to set its start time to >stop time of the simulation... Only for some reason it didn’t seem to work for the proximal ‘common’ feeds!?

rythorpe:
>I thought the test you included (dpl.txt) was checking for the issue you were dealing with in #98. So because this PR also passes the (same) test, the issue should now be solved, no?

No. In #98 I implemented two tests: the first was in 'test_network.py' and validated whether or not all gids had been created. The values initialized in `Network:p_common` are used to assign `Network:gid_dict` and `Network:_gid_list` which are subsequently passed into `Network:_create_all_spike_sources()` where the `ExtFeed` objects are initialized. Thus, the stability of the gids passed into the `ExtFeed` constructor allows for stability of the seed assignment. (It's a mess, I know, but this is what caused the bug in #98.) The second test I implemented occurred, as you have noted, in 'test_compare_hnn.py` where we ensure that a simulation produces only the correct number and types of spike gids.

As the code is now, this PR can produce the buggy scenario addressed in #98 where shifting the start time of null rhythmic inputs (i.e., rhythmic inputs with zero synaptic weight) from `>tstop` to `<tstop` produces a different dipole:

Screenshot from 2020-06-03 13-20-04

jasmainak:
> As the code is now, this PR can produce the buggy scenario addressed in #98 where shifting the start time of null rhythmic inputs (i.e., rhythmic inputs with zero synaptic weight) from >tstop to <tstop produces a different dipole:

But this is how the rhythmic inputs work I think? There seems to be some inconsistency how the different kinds of inputs are handled. Some inputs are turned off by changing t0 > tstop and some are turned off using 0 weights?

jasmainak:
Also I think we really need to separate out the methods into functions and document all the input parameters. Otherwise, it's really hard to get an overview of the different input types and what they do.

rythorpe:
> > As the code is now, this PR can produce the buggy scenario addressed in #98 where shifting the start time of null rhythmic inputs (i.e., rhythmic inputs with zero synaptic weight) from >tstop to <tstop produces a different dipole:
> 
> But this is how the rhythmic inputs work I think? There seems to be some inconsistency how the different kinds of inputs are handled. Some inputs are turned off by changing t0 > tstop and some are turned off using 0 weights?

This is a very inconsistent and confusing way to handle turning on or off inputs. The whole point of #98 was to address this issue.

jasmainak:
I agree, but let's clarify first what existed in `master` before #98, what was changed in the PR and what actually we want to achieve.

With respect to turning off the inputs, what happens with 0 weights, what happens when t0 > tstop, what happens to the gid_dict and what happens to the input spike array -- for different types of inputs? Can we summarize this somehow in a table or something?

rythorpe:
>With respect to turning off the inputs, what happens with 0 weights, what happens when t0 > tstop, what happens to the gid_dict and what happens to the input spike array -- for different types of inputs? Can we summarize this somehow in a table or something?

Here's my attempt to summarize it in a table. Hopefully this helps. I've attached a .pdf version for easy viewing and a zipped .odt file in case anyone wants to download and edit the 'desired' column of the table. Note that I used the default.json param file.

[PR_98_summary_table.pdf](https://github.com/jonescompneurolab/hnn-core/files/4726630/PR_98_summary_table.pdf)
[PR_98_summary_table.zip](https://github.com/jonescompneurolab/hnn-core/files/4726631/PR_98_summary_table.zip)

jasmainak:
@rythorpe some questions:

- does #98 fix only seeds and gids or something else too?
- what if the user provided seed is 0 and your gid is 0, how will you provide a negative seed?
- it seems like a bug to me that the seed depends on gid. Do you know why this is the case?

again, if changing the seeds is going to have such a [drastic effect on the waveforms](https://github.com/jonescompneurolab/hnn-core/pull/103#issuecomment-636353522), I think we should not set them by default. Otherwise, people will have false impressions about the reliability of their results and claims. Or am I missing something?

jasmainak:
regarding 2nd point, see below:

```py
In [6]: np.random.RandomState(-2)                                               
ValueError                                Traceback (most recent call last)
<ipython-input-6-9809f15e3c8a> in <module>
----> 1 np.random.RandomState(-2)

mtrand.pyx in numpy.random.mtrand.RandomState.__init__()

mt19937.pyx in numpy.random.mt19937.MT19937._legacy_seeding()

mt19937.pyx in numpy.random.mt19937.MT19937._legacy_seeding()

ValueError: Seed must be between 0 and 2**32 - 1
```

jasmainak:
regarding last point, see the [numpy documentation](https://numpy.org/doc/1.18/reference/random/generated/numpy.random.Generator.uniform.html#numpy.random.Generator.uniform). You only need to set one seed even to generate multiple random sequences. You just provide the right size and numpy will do the job for you.

rythorpe:
>     * does #98 fix the issue of seeds and the range of values gids can take or something else too?
Yes, #98 stabilizes the gid assignment amidst changes to `t0`. This allows the user to 'turn off' a rhythmic input by setting its synaptic weights to zero without changing the seed (i.e., consistent with all the other input types).
>     * what if the user provided seed is 0 and your gid is 0, how will you provide a negative seed?
Given the order of gid assignment in `Network` and seed assignment in `ExtFeed`, the only way for this scenario to occur is if the network was initialized with 0 cells. Since such a scenario would render the whole simulation pointless (and most likely cause other errors to be thrown), I judged my somewhat janky solution of subtracting 2 from the seed value to be safe _temporary_ solution. Also keep in mind that the only reason the '-2' was necessary was to make the seed match that of HNN. HNN contains the same buggy behavior where a user can load the 'default.param' parameter set, make a few modifications that might include setting `tstop` to longer (i.e., inadvertently causing `t0<tstop` for a null rhythmic input), and get an entirely different single-trial waveform from that shown in the tutorials.
>     * it seems like a bug to me that the seed depends on gid. Do you know why this is the case?
I'm not sure why it was set up this way, however, I think it's only a bug if the `gid_dict` fails to be the same every time. Using gids to set unique randomization seeds seems to be a reasonable approach, especially if a seed has to be assigned to each cell or unique input for parallelization where a random number generator may be required to run separately in each case. One alternative would be for the user to set one seed, and then use that seed to generate a random sequence of seeds to pass to each cell or input. This, however, would render a discrepancy with the current 'ground truth' dipole simulations we use in our tests.
> again, if changing the seeds is going to have such a [drastic effect on the waveforms](https://github.com/jonescompneurolab/hnn-core/pull/103#issuecomment-636353522), I think we should not set them by default. Otherwise, people will have false impressions about the reliability of their results and claims. Or am I missing something?

I disagree. First of all, the seed values we assign by default actually incorporate values the user has set or imported in their param file. Secondly, the only time someone really compares a single trial simulation to another is to do a sanity check where they are trying to replicate the exact dynamics of a previous simulation. This is what we do in our HNN comparison tests and a handful of tutorials. The tutorials, however, also emphasize that we are simulating a process that is only meaningful over many trials where the influence of any specific seed becomes negligible.

cjayb:
> As the code is now, this PR can produce the buggy scenario addressed in #98 where shifting the start time of null rhythmic inputs (i.e., rhythmic inputs with zero synaptic weight) from `>tstop` to `<tstop` produces a different dipole:

I see. So the problem is in fact identical to [that shown here](https://github.com/jonescompneurolab/hnn-core/pull/103#issuecomment-636353522) as a subtle change in the alpha-example: somehow the all-zero-weight proximal 'common' input affects the simulation. I haven't been able to figure out why. But it's not a seed-issue, that's certain. Still think it's got to be in `pyramidal.py:L5pyr().parreceive()`, but no luck yet...

> Using gids to set unique randomization seeds seems to be a reasonable approach, especially if a seed has to be assigned to each cell or unique input for parallelization where a random number generator may be required to run separately in each case.

Good point: the fact that several RNDs might be run on different CPUs means we have to assign gid-based seeds on initialisation.

I think the only reasonable way to approach to this is to figure out the underlying problem of common proximal inputs affecting the sim's when they shouldn't. Then write a fix that keeps the seed assignment consistent/predictable. 

> This, however, would render a discrepancy with the current 'ground truth' dipole simulations we use in our tests.

I think it might be reasonable to ask whether keeping the ground truth when several bugs have been found & fixed since it was created? It should be possible to re-create any tutorial waveform with a 'suitable' parameter set, of course.

jasmainak:
I feel we might need a 3-way zoom/skype call to discuss some of the issues in person. Maybe I can discuss with @rythorpe first to understand what is the issue and then we can set a 3-way call if necessary!

Let me respond to a couple of things above:

> But it's not a seed-issue, that's certain

humm, I'm not so certain. Changing the gid_dict changed the seeds so @rythorpe added -2 to account for it. Or do I misunderstand something?

> I judged my somewhat janky solution of subtracting 2 from the seed value to be safe temporary solution.

okay this is fair enough

> Secondly, the only time someone really compares a single trial simulation to another is to do a sanity check where they are trying to replicate the exact dynamics of a previous simulation

Isn't the optimization that is implemented in the GUI also doing seed-based single-trial simulations?

> The tutorials, however, also emphasize that we are simulating a process that is only meaningful over many trials where the influence of any specific seed becomes negligible.

I know but I feel this is not enough. Most scientific software (cf MNE, scikit-learn) do not set the seed. The reason is that people will realize it's stochastic, then set the seed and do the experiment.

 >Good point: the fact that several RNDs might be run on different CPUs means we have to assign gid-based seeds on initialisation.

Not really, we can create the input vectors before sending things to parallel jobs.

> One alternative would be for the user to set one seed, and then use that seed to generate a random sequence of seeds to pass to each cell or input.

This is exactly what numpy is already doing for you. We shouldn't reinvent the wheel.

> I think it might be reasonable to ask whether keeping the ground truth when several bugs have been found & fixed since it was created?

I'm with you on that but I think this needs to be a broader discussion with @stephanie-r-jones . What would be a reasonable new ground truth? After fixing how many bugs do we consider that a new ground truth should be created?

rythorpe:
I was just able to recreate the [alpha discrepancy](https://github.com/jonescompneurolab/hnn-core/pull/103#issuecomment-636353522) by modifying the seed.

In the attached .zip file I have include a modified version of the alpha example where I run 2 simulations for 710ms. Note that I change the seed core for the distal rhythmic input to 5 when previously it was 4. This was caused by the fact that I normalized the `gid_dict` attribute in #98 which cause an offset in the seed for a simulation when only a distal (and not proximal) input is set with `t0<tstop`. Here is an excerpt of my modified alpha example code as well as the output:
```
net = Network(params)
print('gid_dict =',net.gid_dict)
dpl = simulate_dipole(net)
dpl[0].plot()

# Increment seed for distal rhythmic input by 1
# Rerun simulatio
params['prng_seedcore_input_dist'] = 5
net = Network(params)
print('gid_dict =',net.gid_dict)
dpl = simulate_dipole(net)
dpl[0].plot()
```

Original:
Screenshot from 2020-06-04 14-33-29

With modified seed:
Screenshot from 2020-06-04 14-35-35



[alpha_example.zip](https://github.com/jonescompneurolab/hnn-core/files/4732132/alpha_example.zip)

blakecaldwell:
Hi everyone, lots of progress! Great work! I've been following along and thought now would be a good time to jump in.

> 
> > Secondly, the only time someone really compares a single trial simulation to another is to do a sanity check where they are trying to replicate the exact dynamics of a previous simulation
> 
> Isn't the optimization that is implemented in the GUI also doing seed-based single-trial simulations?

No, optimization is done with as many trials as you want. The tutorial uses 3:
https://jonescompneurolab.github.io/hnn-tutorials/optimization/optimization

> 
> > I think it might be reasonable to ask whether keeping the ground truth when several bugs have been found & fixed since it was created?
> 
> I'm with you on that but I think this needs to be a broader discussion with @stephanie-r-jones . What would be a reasonable new ground truth? After fixing how many bugs do we consider that a new ground truth should be created?

We could say "consistent" instead of "ground truth". The ability to produce consistent results also means that we can explain any differences, whether a bug or a design choice. Right now this is difficult because hnn-core and HNN GUI have their own code for running simulations. However, the plan is to have HNN GUI run simulations with hnn-core in the near future.

jasmainak:
@cjayb I had a video call with @rythorpe just a moment ago. To summarize the discussion:

- Before #98, "common" inputs could be turned off by having t0 > tstop and in that case these inputs were not present in gid_dict
- in #98 the "common" input was made to be _always_ present in the gid_dict. This effectively means that the "common" inputs can no longer be turned off by having t0 > tstop. Since #98, the only way to turn them off is through 0 synaptic weights.
- Unfortunately there were two issues with this pull request: 1) the "common" inputs were turned off (having an empty vector of input spikes) if any of the synaptic weights were 0 rather than "all" of them, and 2) the change in gid caused the seeds to change. The hack of gid - 2 did not actually fix the problem because you could have either 1 or 2 "common" inputs and this made the input spikes dependent on the number of "common" inputs
- In #102 you fixed problem 1) but 2) was not fixed
- The problem of 2) was discovered by me in #103 
- This was "fixed" in #103 by undoing "common" inputs to be always present (done in #98)

I proposed to @rythorpe to fix this by doing the following:
- remove the hack of gid - 2
- adjust the seeds in the example to account for the fact that the gid_dict always has "common" inputs now
- fix the gid_test test to where it was before #103 so you check that "common" inputs are always present

Moving forward we would need to have a better way to set seeds but I think this can happen only after you make the pull request to refactor ExtFeed into separate functions. In the meanwhile @rythorpe can work on a pull request to fix these discrepancies. How does this sound like you? We can also do a video call if it helps you understanding the situation better.

cjayb:
Thanks for the update, LGTM! Glad it was the seed after all, really. 

I’ll work on the extfeed next, leave sorting out the seeds vs gids to others. I’m still a little weirded out about having feeds in the net that aren’t actually doing anything... I don’t have an alternate suggestion though for fixing the train of seeds, so will leave it to others to decide. 

> the plan is to have HNN GUI run simulations with hnn-core in the near future.

Naming this project ‘core’ was a dead giveaway ;) ‘Consistent’ works for me. I think the design goal should be to make the code (and certainly the API) as clear as possible. Getting consistent results with an ‘old’ code base might get incrementally harder, but as long as we can do it that’s fine, of course.

PR #171, opened 2020-09-11
[MRG] fix inconsistent random state assignment between parallel trials

### LINKED ISSUE #136: discrepancy between simulation trial variability in HNN vs hnn-core
I'm experiencing an issue where I get significantly different results, particularly noticeable in the inter-trial variability, in HNN versus hnn-core. Running the attached param file in HNN (top) and hnn-core (bottom) using the attached python script, I get the figures shown below. Furthermore, I haven't been able to reproduce this issue using the default parameter set.

Screenshot from 2020-08-11 15-05-02
Screenshot from 2020-08-11 15-20-03

[issue_demo.zip](https://github.com/jonescompneurolab/hnn-core/files/5059115/issue_demo.zip)

rythorpe:
fixes #136 

Running the code [here](https://gist.github.com/rythorpe/e7bac070bff41f108996b98c8b58ac14) creates this plot with `MPI_Backend`:
mn_mpi_10trials

and this plot for `JoblibBackend`:

mn_joblib_10trials

Note that while individual trials are not exactly the same between the two versions (most likely due to an inconsistency in the trajectories of random states), both show 10 distinct trials that emerged from distinct random states. If anyone can readily spot where the discrepancy is occurring it be much appreciated. Otherwise we might want to address that in another PR since the tests are still passing.

This PR also addresses some of the inconsistent messaging that gets output during a simulation in the terminal.

### DIFF
--- a/hnn_core/network_builder.py
+++ b/hnn_core/network_builder.py
@@ -11,6 +11,7 @@ from .feed import ExtFeed
 from .cell import _ArtificialCell
 from .pyramidal import L2Pyr, L5Pyr
 from .basket import L2Basket, L5Basket
+from .params import create_pext
 
 # a few globals
 _PC = None
@@ -120,8 +121,6 @@ def _simulate_single_trial(neuron_net):
         dpl.convert_fAm_to_nAm()
         dpl.scale(neuron_net.net.params['dipole_scalefctr'])
         dpl.smooth(neuron_net.net.params['dipole_smooth_win'] / h.dt)
-
-    neuron_net.net.trial_idx += 1
 
     return dpl
 
@@ -353,6 +352,11 @@ class NetworkBuilder(object):
         External inputs are not targets.
         """
         params = self.net.params
+        # Re-create external feed param dictionaries
+        # Note that the only thing being updated here are the param['prng_*']
+        # values
+        self.net.p_common, self.net.p_unique = create_pext(params,
+                                                           params['tstop'])
 
         # loop through gids on this node
         for gid in self.net._gid_list:
--- a/hnn_core/network_builder.py
+++ b/hnn_core/network_builder.py
@@ -24,7 +24,7 @@ _CVODE = None
 _LAST_NETWORK = None
 
 
-def _simulate_single_trial(neuron_net):
+def _simulate_single_trial(neuron_net, trial_idx):
     """Simulate one trial."""
 
     from .dipole import Dipole
@@ -41,7 +41,7 @@ def _simulate_single_trial(neuron_net):
     _PC.barrier()  # sync for output to screen
     if rank == 0:
         print("running trial %d on %d cores" %
-              (neuron_net.net.trial_idx + 1, nhosts))
+              (trial_idx + 1, nhosts))
 
     # create or reinitialize scalars in NEURON (hoc) context
     h("dp_total_L2 = 0.")
--- a/hnn_core/parallel_backends.py
+++ b/hnn_core/parallel_backends.py
@@ -86,12 +86,13 @@ class JoblibBackend(object):
         from hnn_core.network_builder import NetworkBuilder
         from hnn_core.network_builder import _simulate_single_trial
 
+        # XXX this should be built into NetworkBuilder
+        # update prng_seedcore params to provide jitter between trials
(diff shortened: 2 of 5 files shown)


hnn_core/parallel_backends.py
```diff
@@ -86,10 +86,11 @@ def _clone_and_simulate(self, net, trial_idx):
         from hnn_core.network_builder import NetworkBuilder
         from hnn_core.network_builder import _simulate_single_trial
 
-        if trial_idx != 0:
-            net.params['prng_*'] = trial_idx
+        for param_key in net.params['prng_*'].keys():
+            net.params[param_key] += trial_idx
 
         neuron_net = NetworkBuilder(net)
+        neuron_net.net.trial_idx = trial_idx
```
jasmainak, line 93: is this equivalent to https://github.com/jonescompneurolab/hnn-core/pull/171/files#diff-ca5f146e027d9dc6ef29ff4db9c21ca2R68 ? I have a suspicion it is not. Can you check?

hnn_core/network_builder.py
```diff
@@ -353,6 +352,11 @@ def _create_cells_and_feeds(self):
         External inputs are not targets.
         """
         params = self.net.params
+        # Re-create external feed param dictionaries
```
jasmainak, line 355: should we add an "XXX" in the comment somewhere? To indicate that this is a hack (right?) and we should get rid of at some point

jasmainak:
I have a hunch that digging [into this](https://github.com/jonescompneurolab/hnn-core/pull/171#discussion_r487367214) might solve your issue. Then ideally, we can add a tiny test to check that MPI vs Joblibs gives same result for 2 trials (with a 3 x 3 network to make it run fast). If it doesn't work, I am fine merging the PR as is if it's good by @blakecaldwell 

do make sure to update `whats_new.rst`. I see that you found a fun way to end your week ;-)

hnn_core/network_builder.py
rythorpe: Hmm, I'm not sure in this scenario what delineates a hack from a long-term bug fix. We plan to refactor the params and feed modules wrt how feed objects and their respective parameters are passed to `Network` and `NetworkBuilder`. Along the way we will end up simplifying the code by e.g. removing the need for calling `create_pext()` here. That said, I'm happy to update the comment if you think it'd be clearer.

jasmainak: yeah, I agree it might go away but just leaving an XXX tells future contributors that it's a hack that needs to fixed.

hnn_core/parallel_backends.py
rythorpe: You're right, I messed this up. Fixed in https://github.com/jonescompneurolab/hnn-core/pull/171/commits/b340fe0150c7b333d40a56f3e87ea0c0ea85004f.

rythorpe:
Update with https://github.com/jonescompneurolab/hnn-core/pull/171/commits/b340fe0150c7b333d40a56f3e87ea0c0ea85004f: the script that generated the plots shown at the [beginning of this PR](https://github.com/jonescompneurolab/hnn-core/pull/171#issue-485560556) now produces the same results for the joblib and mpi backends when running the default param set: 

mn_mpi_10trials_02
mn_joblib_10trials_02

jasmainak:
Amazing! do you think we could add a tiny test?

- perhaps one test that checks that the two trials are different
- and another that checks that the results are consistent between joblib and MPI?

you can use a small network, e.g., `params.update({"N_pyr_x": 3, "N_pyr_y": 3})` and maybe a shorter `tstop` to make it run fast?

also don't forget to update `whats_new.rst`!

rythorpe:
I've updated `whats_new.rst` and added your suggested tests @jasmainak. I also ended up rearranging some of the test functions in `test_compare_hnn.py` in https://github.com/jonescompneurolab/hnn-core/pull/171/commits/b8f5d1d5935cc756c478b5ef2bae92e1cec17508 in order to limit redundant code. Take a look and let me know if the changes are clear and acceptable.

jasmainak (approved):

jasmainak:
Fantastic, thanks for making the changes. PR looks good to me, merged!

PR #191, opened 2020-09-26
[MRG] Refactor creation of network 'feeds' (to allow new mechanism for adding network drives)

### LINKED ISSUE #104: clean up handling of external 'input' feeds

cjayb:
Based on the work done in #177, continue creating new API for adding (and activating/simulating) external drives of the network (currently 'feeds'). Some ideas are discussed in a separate Doc (contact @cjayb for access). This is going to be a test-breaking PR due to very idiosyncratic handling of random seeds used in HNN-GUI feed assignment.

Edit 6 Nov:
Discussed GID-handling with @jasmainak and figured much could be won by disentangling it from NetworkBuilder. As it stands, `Network` creates the `gid_dict` on initialisation, so all that's really left for `Builder` to do is assign them to the different nodes (ranks).

Aims
- [x] clarify GID creation & handling
- [x] move all `ExtFeed` calls into `Network`, allowing instantiation of all external drives before `NetworkBuilder` is called (check old "API" still works)
- [x] move `parallel_backends.py:_clone_and_simulate:L34-35` to `NetworkBuilder` in order to "provide jitter between trials" (`prng_seedcore` needs to be different)
  * turned out not to be needed
- [x] generally refactor `Network` and `NetworkBuilder` so that building a new mechanism easier (separate PR)

For future PR
- demonstrate API ideas to allow incrementally adding drives
- create new tests of `Network` event times
- create new visualisations of `Network` event times 
- discuss and settle on seed handling 
- probably more...

Closes #104

### DIFF
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -22,6 +22,10 @@ class _ArtificialCell:
         associated with a unique gid).
     threshold : float
         Membrane potential threshold that demarks a spike.
+    gid : int or None (optional)
+        Each cell in a network is uniquely identified by it's "global ID": GID.
+        The GID is an integer from 0 to n_cells, or None if the cell is not
+        yet attached to a network. Once the GID is set, it cannot be changed.
 
     Attributes
     ----------
@@ -33,8 +37,10 @@ class _ArtificialCell:
     nrn_netcon : instance of h.NetCon()
         NEURON h.NetCon() object that creates the spike
         source-to-target references for nrn_vecstim.
+    gid : int
+        GID of the cell in a network (or None if not yet assigned)
     """
-    def __init__(self, event_times, threshold):
+    def __init__(self, event_times, threshold, gid=None):
         # Convert event times into nrn vector
         self.nrn_eventvec = h.Vector()
         self.nrn_eventvec.from_python(event_times)
@@ -47,17 +53,35 @@ class _ArtificialCell:
         self.nrn_netcon = h.NetCon(self.nrn_vecstim, None)
         self.nrn_netcon.threshold = threshold
 
+        self._assigned_gid = None
+        self.gid = gid
+
+    @property
+    def gid(self):
+        return self._assigned_gid
+
+    @gid.setter
+    def gid(self, gid):
+        if gid is None:
+            pass
+        elif self._assigned_gid is None:
+            self._assigned_gid = gid
+        else:
+            raise RuntimeError('Global ID for this cell already assigned!')
+
 
 class _Cell(ABC):
     """Create a cell object.
 
     Parameters
     ----------
-    gid : int
-        The cell ID
     soma_props : dict
         The properties of the soma. Must contain
         keys 'L', 'diam', and 'pos'
+    gid : int or None (optional)
+        Each cell in a network is uniquely identified by it's "global ID": GID.
+        The GID is an integer from 0 to n_cells, or None if the cell is not
+        yet attached to a network. Once the GID is set, it cannot be changed.
 
     Attributes
     ----------
@@ -70,15 +94,18 @@ class _Cell(ABC):
     rec_v : h.Vector()
         Recording of somatic voltage. Must be enabled
         by running simulate_dipole(net, record_vsoma=True)
+    gid : int
+        GID of the cell in a network (or None if not yet assigned)
     """
 
-    def __init__(self, gid, soma_props):
-        self.gid = gid
+    def __init__(self, soma_props, gid=None):
         # variable for the list_IClamp
         self.list_IClamp = None
         self.soma_props = soma_props
         self.create_soma()
         self.rec_v = h.Vector()
+        self._assigned_gid = None
+        self.gid = gid
 
     def __repr__(self):
         class_name = self.__class__.__name__
@@ -87,6 +114,19 @@ class _Cell(ABC):
              (soma_props['L'], soma_props['diam'],
               soma_props['Ra'], soma_props['cm']))
         return '<%s | %s>' % (class_name, s)
+
+    @property
+    def gid(self):
+        return self._assigned_gid
+
+    @gid.setter
+    def gid(self, gid):
+        if gid is None:
+            pass
+        elif self._assigned_gid is None:
+            self._assigned_gid = gid
+        else:
+            raise RuntimeError('Global ID for this cell already assigned!')
 
     @abstractmethod
     def get_sections(self):
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -53,18 +53,20 @@ class _ArtificialCell:
         self.nrn_netcon = h.NetCon(self.nrn_vecstim, None)
         self.nrn_netcon.threshold = threshold
 
-        self._assigned_gid = gid
+        self._gid = None
+        if gid is not None:
+            self.gid = gid  # use setter method to check input argument gid
 
     @property
     def gid(self):
-        return self._assigned_gid
+        return self._gid
 
     @gid.setter
     def gid(self, gid):
         if not isinstance(gid, int):
             raise ValueError('gid must be an integer')
-        if self._assigned_gid is None:
-            self._assigned_gid = gid
+        if self._gid is None:
+            self._gid = gid
         else:
             raise RuntimeError('Global ID for this cell already assigned!')
 
@@ -103,7 +105,9 @@ class _Cell(ABC):
         self.soma_props = soma_props
         self.create_soma()
         self.rec_v = h.Vector()
-        self._assigned_gid = gid
+        self._gid = None
+        if gid is not None:
+            self.gid = gid  # use setter method to check input argument gid
 
     def __repr__(self):
         class_name = self.__class__.__name__
@@ -115,14 +119,14 @@ class _Cell(ABC):
 
     @property
     def gid(self):
-        return self._assigned_gid
+        return self._gid
 
     @gid.setter
     def gid(self, gid):
         if not isinstance(gid, int):
             raise ValueError('gid must be an integer')
-        if self._assigned_gid is None:
-            self._assigned_gid = gid
+        if self._gid is None:
+            self._gid = gid
         else:
             raise RuntimeError('Global ID for this cell already assigned!')
 
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -53,8 +53,7 @@ class _ArtificialCell:
         self.nrn_netcon = h.NetCon(self.nrn_vecstim, None)
         self.nrn_netcon.threshold = threshold
 
-        self._assigned_gid = None
-        self.gid = gid
+        self._assigned_gid = gid
 
     @property
     def gid(self):
@@ -62,9 +61,9 @@ class _ArtificialCell:
 
     @gid.setter
     def gid(self, gid):
-        if gid is None:
-            pass
-        elif self._assigned_gid is None:
+        if not isinstance(gid, int):
+            raise ValueError('gid must be an integer')
+        if self._assigned_gid is None:
             self._assigned_gid = gid
         else:
             raise RuntimeError('Global ID for this cell already assigned!')
@@ -104,8 +103,7 @@ class _Cell(ABC):
         self.soma_props = soma_props
         self.create_soma()
         self.rec_v = h.Vector()
-        self._assigned_gid = None
-        self.gid = gid
+        self._assigned_gid = gid
 
     def __repr__(self):
         class_name = self.__class__.__name__
@@ -121,9 +119,9 @@ class _Cell(ABC):
 
     @gid.setter
     def gid(self, gid):
-        if gid is None:
-            pass
-        elif self._assigned_gid is None:
+        if not isinstance(gid, int):
+            raise ValueError('gid must be an integer')
+        if self._assigned_gid is None:
             self._assigned_gid = gid
         else:
             raise RuntimeError('Global ID for this cell already assigned!')
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -44,6 +44,7 @@ def simulate_dipole(net, n_trials=None, record_vsoma=False):
 
     if n_trials is not None:
         net.params['N_trials'] = n_trials
+        net._instantiate_feeds()  # need to redo these if n_trials changed!
     else:
         n_trials = net.params['N_trials']
 
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -44,7 +44,8 @@ def simulate_dipole(net, n_trials=None, record_vsoma=False):
 
     if n_trials is not None:
         net.params['N_trials'] = n_trials
-        net._instantiate_feeds()  # need to redo these if n_trials changed!
+        # need to redo these if n_trials changed after net.__init__()!
+        net._instantiate_feeds(n_trials=n_trials)
     else:
         n_trials = net.params['N_trials']
 
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -173,10 +173,8 @@ class Network(object):
         # place to keep this information
         self.gid_ranges = dict()
 
-        # Number of time points
-        # Originally used to create the empty vec for synaptic currents,
-        # ensuring that they exist on this node irrespective of whether
-        # or not cells of relevant type actually
+        # Create array of equally sampled time points for simulating currents
+        # NB (only) used to initialise self.spikes._times
         times = np.arange(0., params['tstop'] + params['dt'], params['dt'])
         # Create spikes object, initialised with simulation time points
         self.spikes = Spikes(times=times)
@@ -447,8 +445,8 @@ class Spikes(object):
         self._spike_types = spike_types
         self._vsoma = list()
         if times is not None:
-            assert isinstance(times, np.ndarray),\
-                "'times' is an array of simulation times"
+            if not isinstance(times, np.ndarray):
+                raise TypeError("'times' is an np.ndarray of simulation times")
         self._times = times
 
     def __repr__(self):
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -177,10 +177,9 @@ class Network(object):
         # Originally used to create the empty vec for synaptic currents,
         # ensuring that they exist on this node irrespective of whether
         # or not cells of relevant type actually
-        self.times = np.arange(0., params['tstop'] + params['dt'],
-                               params['dt'])
-        # Create empty spikes object
-        self.spikes = Spikes()
+        times = np.arange(0., params['tstop'] + params['dt'], params['dt'])
+        # Create spikes object, initialised with simulation time points
+        self.spikes = Spikes(times=times)
 
         # Source list of names, first real ones only!
         self.cellname_list = [
@@ -397,8 +396,8 @@ class Spikes(object):
         Each gid corresponds to a type via Network::gid_ranges.
     vsoma : dict
         Dictionary indexed by gids containing somatic voltages
-    times : list
-        List of time points for samples in continuous data.
+    times : numpy array
+        Array of time points for samples in continuous data.
         This includes vsoma.
 
     Methods
@@ -415,7 +414,8 @@ class Spikes(object):
         Write spiking activity to a collection of spike trial files.
     """
 
-    def __init__(self, spike_times=None, spike_gids=None, spike_types=None):
+    def __init__(self, spike_times=None, spike_gids=None, spike_types=None,
+                 times=None):
         if spike_times is None:
             spike_times = list()
         if spike_gids is None:
@@ -446,7 +446,10 @@ class Spikes(object):
         self._spike_gids = spike_gids
         self._spike_types = spike_types
         self._vsoma = list()
-        self._times = list()
+        if times is not None:
+            assert isinstance(times, np.ndarray),\
+                "'times' is an array of simulation times"
+        self._times = times
 
     def __repr__(self):
         class_name = self.__class__.__name__
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -255,20 +255,24 @@ class Network(object):
         self._update_gid_ranges()
 
         # Create the feed dynamics (event_times)
-        self._instantiate_feeds()
+        self._instantiate_feeds(n_trials=self.params['N_trials'])
 
-    def _instantiate_feeds(self):
-        '''Creates event_time vectors for all feeds and all trials
+    def _instantiate_feeds(self, n_trials=1):
+        """Creates event_time vectors for all feeds and all trials
+
+        Parameters
+        ----------
+        n_trials : int
+            Number of trials to create events for (default: 1)
 
         NB this must be a separate method because dipole.py:simulate_dipole
         accepts an n_trials-argument, which overrides the N_trials-parameter
         used at intialisation time. The good news is that only the event_times
         need to be recalculated, all the GIDs etc remain the same.
-        '''
+        """
         # each trial needs unique event time vectors
 
         self.trial_event_times = []  # reset if called again from dipole.py
-        n_trials = self.params['N_trials']
 
         cur_params = self.params.copy()  # these get mangled below!
         for trial_idx in range(n_trials):
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -7,12 +7,13 @@
 import itertools as it
 import numpy as np
 from glob import glob
+from .feed import ExtFeed
 
 from .params import create_pext
 from .viz import plot_spikes_hist, plot_spikes_raster, plot_cells
 
 
-def read_spikes(fname, gid_dict=None):
+def read_spikes(fname, gid_ranges=None):
     """Read spiking activity from a collection of spike trial files.
 
     Parameters
@@ -20,7 +21,7 @@ def read_spikes(fname, gid_dict=None):
     fname : str
         Wildcard expression (e.g., '<pathname>/spk_*.txt') of the
         path to the spike file(s).
-    gid_dict : dict of lists or range objects | None
+    gid_ranges : dict of lists or range objects | None
         Dictionary with keys 'evprox1', 'evdist1' etc.
         containing the range of Cell or input IDs of different
         cell or input types. If None, each spike file must contain
@@ -41,20 +42,20 @@ def read_spikes(fname, gid_dict=None):
         spike_gids += [list(spike_trial[:, 1].astype(int))]
 
         # Note that legacy HNN 'spk.txt' files don't contain a 3rd column for
-        # spike type. If reading a legacy version, validate that a gid_dict is
-        # provided.
+        # spike type. If reading a legacy version, validate that a gid_ranges
+        # is provided.
         if spike_trial.shape[1] == 3:
             spike_types += [list(spike_trial[:, 2].astype(str))]
         else:
-            if gid_dict is None:
-                raise ValueError("gid_dict must be provided if spike types "
+            if gid_ranges is None:
+                raise ValueError("gid_ranges must be provided if spike types "
                                  "are unspecified in the file %s" % (file,))
             spike_types += [[]]
 
     spikes = Spikes(spike_times=spike_times, spike_gids=spike_gids,
                     spike_types=spike_types)
-    if gid_dict is not None:
-        spikes.update_types(gid_dict)
+    if gid_ranges is not None:
+        spikes.update_types(gid_ranges)
 
     return Spikes(spike_times=spike_times, spike_gids=spike_gids,
                   spike_types=spike_types)
@@ -147,12 +148,12 @@ class Network(object):
         The names of real cell types in the network (e.g. 'L2_basket')
     feedname_list : list
         The names of external drivers ('feeds') to the network (e.g. 'evdist1')
-    gid_dict : dict
+    gid_ranges : dict
         A dictionary of unique identifiers of each real and artificial cell
         in the network. Every cell type is represented by a key read from
         cellname_list, followed by entries read from feedname_list. The value
         of each key is a range of ints, one for each cell in given category.
-        Examples: 'L2_basal': range(0, 270), 'evdist1': range(272, 542), etc
+        Examples: 'L2_basket': range(0, 270), 'evdist1': range(272, 542), etc
     pos_dict : dict
         Dictionary containing the coordinate positions of all cells.
         Keys are 'L2_pyramidal', 'L5_pyramidal', 'L2_basket', 'L5_basket',
@@ -164,22 +165,13 @@ class Network(object):
     def __init__(self, params):
         # Save the parameters used to create the Network
         self.params = params
-        # Initialise a dictionary of cell ID's, which get assigned when the
+        # Initialise a dictionary of cell ID's, which get used when the
         # network is constructed ('built') in NetworkBuilder
         # We want it to remain in each Network object, so that the user can
         # interrogate a built and simulated net. In addition, Spikes are
         # attached to a Network during simulation---Network is the natural
         # place to keep this information
-        # XXX rename gid_dict in future
-        self.gid_dict = dict()
-
-        # When computing the network dynamics in parallel, the nodes of the
-        # network (real and artificial cells) potentially get distributed
-        # on different host machines/threads. NetworkBuilder._gid_assign
-        # assigns each node, identified by its unique GID, to one of the
-        # possible hosts/threads for computations. _gid_list here contains
-        # the GIDs of all the nodes assigned to the current host/thread.
-        self._gid_list = []
+        self.gid_ranges = dict()
 
         # Number of time points
         # Originally used to create the empty vec for synaptic currents,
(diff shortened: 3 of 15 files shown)


cjayb:
Just getting off to a start to make @jasmainak happy ;) Your last MAINT to #177 really helped clear it up for me, thanks! Hope to push some API ideas in soon.

hnn_core/network.py
```diff
@@ -251,8 +253,15 @@ def _add_external_feeds(self):
...
-        """Creates gid dicts."""
+        self._update_gid_dict()  # add feeds -> update the gid_dict
+
+    def _update_gid_dict(self):
+        """Creates gid dict from scratch every time called.
+
+        Any method that adds real or artificial cells to the network must
+        call this to update the list of GIDs. Note that it's based on the
+        content of pos_dict and the lists of real and artificial cell names.
+        """
```
rythorpe, line 264: This is a great idea @cjayb and will allow the user to more flexibly create or destroy cells or feeds! Since `Network.gid_dict` and `Network.pos_dict` are so intimately tied, would it make sense to combine their assignment under a unified method?

hnn_core/network_builder.py
```diff
@@ -300,10 +294,14 @@ def _build(self):
 
         self._gid_assign()
 
+        # initialise vectors for synaptic currents based on simulation params
+        n_times = np.arange(0., self.net.params['tstop'],
+                            self.net.params['dt']).size + 1
```
rythorpe, line 299: In support of this change, I can't think of any reason why `n_times` needs to be an attribute of `Network`. @blakecaldwell are there any GUI-related considerations we're missing here?

blakecaldwell: I agree, it doesn't seem necessary. Thanks for notifying me @rythorpe.

jasmainak:
@cjayb do you want to pair program sometime for this?

and by pair program, I mean really code and not discuss ;-)

cjayb:
Hey man, yes I think that would be a good idea! I'm 6 hours ahead, but I'm
free most evenings next week. How about Tuesday 3rd after 11 am/5 pm? I'm
assuming you can't vote ;) Weds and Thur also work, both from mid-afternoon
onwards. /C

cjayb:
@jasmainak as a discussion-starter for tomorrow. Sorry about the git-hell, will prob. have to ditch this and start from scratch anyway! Bottom line: a lot harder to reach the type of API we were discussing. Some ideas here (like a curry'ed "generating function"), but looking forward to your insight :)

cjayb:
As the commit says: WIP So no need to review just yet ;) There's a lot of cruft I need to weed out, but basically very confident that `ExtFeed` is now out of `NetworkBuilder` for good. I'll give a shout when I think it might make sense to take a closer peek.

cjayb:
Awh crap, MPI is not happy about my changes. Will try to backtrack order changes in `NetworkBuilder`.

hnn_core/tests/test_parallel_backends.py
```diff
@@ -104,16 +104,18 @@ def test_compare_across_backends():
...
         # account for rounding error incured during MPI parallelization
         if dpls_reduced_mpi:
             assert_allclose(dpls_reduced_default[trial_idx].data['agg'],
                             dpls_reduced_mpi[trial_idx].data['agg'], rtol=0,
-                            atol=1e-14)
+                            atol=1e-14, err_msg='MPI values differ too much!')
         assert_array_equal(dpls_reduced_default[trial_idx].data['agg'],
-                           dpls_reduced_joblib[trial_idx].data['agg'])
+                           dpls_reduced_joblib[trial_idx].data['agg'],
+                           err_msg='Joblib values not equal to serial impl.')
```
cjayb, line 118: I got this comment wrong (will fix in next commit): the test is to make sure that two trials have _different_ model prediction values! Have updated PR description to indicate where the problem lies. No easy solution in sight. Good news is that MPI tests now pass (single-trial runs produce the expected spike times).

cjayb: Well I found a way around it, with smaller edits than I had feared

hnn_core/network.py
```diff
@@ -135,34 +136,62 @@ class Network(object):
...
+        # Initialise a dictionary of cell ID's, which get assigned when the
+        # network is constructed ('built') in NetworkBuilder
+        # We want it to remain in each Network object, so that the user can
+        # interrogate a built and simulated net. In addition, Spikes are
+        # attached to a Network during simulation---Network is the natural
+        # place to keep this information
+        # XXX rename gid_dict in future
+        self.gid_dict = dict()
+
+        # When computing the network dynamics in parallel, the nodes of the
```
jasmainak, line 175: I thought "node" meant the different machines that you distribute your jobs to.

hnn_core/network.py
```diff
@@ -135,34 +136,62 @@ class Network(object):
...
+
+        # XXX Note that GIDs are already defined here at the Network-level
+        # All that's left for NetworkBuilder is then to:
+        # - _PC.set_gid2node(gid, rank)
+        # - _PC.cell(gid, nrn_netcon) (or _PC.cell(feed_cell.gid, nrn_netcon))
+
+        # dict with keys 'common', 'evdist1', ...
+        # XXX needs better semantics, here to get first pass working
+        # list (len == n_trials) of dicts (keys: src_type) of event times
+        self._tmp_feed_event_times = []
```
jasmainak, line 191: document it in docstring not in comments :) it's more easy to find there.

hnn_core/network_builder.py
```diff
@@ -377,78 +385,59 @@ def _create_cells_and_feeds(self):
...
-                               target_cell_type=target_cell_type,
-                               params=self.net.p_unique[src_type],
-                               gid=gid)
-                feed_cell = _ArtificialCell(feed.event_times,
-                                            params['threshold'])
-                self._feed_cells.append(feed_cell)
-            else:
-                raise ValueError('No parameters specified for external feed '
-                                 'type: %s' % src_type)
+    def _assign_cells_and_feeds_to_neuron_network(self):
```
jasmainak, line 425: how about:

```suggestion
    def _gid_to_pc(self):
```

descriptive enough? I don't want the name to be too vague ...

also maybe add an argument called `threshold` so we don't accidentally add more dependencies on `params` inside the function.

hnn_core/network_builder.py
```diff
@@ -265,8 +267,9 @@ class NetworkBuilder(object):
     `self.net.params` and the network is ready for another simulation.
     """
 
-    def __init__(self, net):
+    def __init__(self, net, trial_idx=0):
         self.net = net
+        self.cur_trial = trial_idx
```
jasmainak, line 272: self.trial_idx ?

to avoid learning a new variable name

hnn_core/network.py
```diff
@@ -251,8 +264,58 @@ def _add_external_feeds(self):
...
+            # needs to be re-run to create the dicts going into ExtFeed
+            p_common, p_unique = create_pext(cur_params,
+                                             cur_params['tstop'])
+
+            src_types = self.feedname_list
+            event_times_per_source = {}
+            for idx, src_type in enumerate(src_types):
+                event_times = []
+                if src_type == 'common':
+                    for ic, gid in enumerate(self.gid_dict['common']):
```
jasmainak, line 292: ic -> idx

it's what we use as short for "index" elsewhere

hnn_core/network.py
```diff
@@ -251,8 +264,58 @@ def _add_external_feeds(self):
         unique_keys = sorted(self.p_unique.keys())
         self.feedname_list += unique_keys
 
-    def _create_gid_dict(self):
-        """Creates gid dicts."""
+        self._update_gid_dict()  # add feeds -> update the gid_dict
+        self.instantiate_feed_events()
+
+    def instantiate_feed_events(self):
```
jasmainak, line 270: `instantiate_feeds` ?

I think names of python objects/variables are more precise than descriptive names. Because now I was wondering what is "feed events"

hnn_core/cell.py
```diff
@@ -51,30 +51,46 @@ def __init__(self, event_times, threshold):
...
+        self._assigned_gid = None
+        if gid is not None:
+            self.gid = gid
+
+    @property
+    def gid(self):
+        return self._assigned_gid
+
+    @gid.setter
+    def gid(self, gid):
```
jasmainak, line 63: is there a realistic usecase for being able to update cell gids? Otherwise, I would just hide the functionality. The less you expose, the better

jasmainak:
@cjayb most of my comments are nitpicky about variable names.

I think the next step is to write tests for the new methods. This way you'll know --> a) they work as you expect b) they are modular enough to be part of tests

hnn_core/network_builder.py
```diff
@@ -298,15 +295,20 @@ def _build(self):
...
 
         # Create a h.Vector() with size 1xself.N_t, zero'd
         self.current = {
-            'L5Pyr_soma': h.Vector(self.net.n_times, 0),
-            'L2Pyr_soma': h.Vector(self.net.n_times, 0),
+            'L5Pyr_soma': h.Vector(n_times, 0),
+            'L2Pyr_soma': h.Vector(n_times, 0),
         }
 
+        self._gid_assign()
```
jasmainak, line 308: I think `_gid_assign` should happen after `_create_cells_and_feeds`, no?

hnn_core/network_builder.py
```diff
@@ -298,15 +291,21 @@ def _build(self):
...
         self.current = {
-            'L5Pyr_soma': h.Vector(self.net.n_times, 0),
-            'L2Pyr_soma': h.Vector(self.net.n_times, 0),
+            'L5Pyr_soma': h.Vector(n_times, 0),
+            'L2Pyr_soma': h.Vector(n_times, 0),
         }
 
         self._create_cells_and_feeds()
+
+        self._gid_assign()
```
jasmainak, line 308: I have been thinking a bit about what is inside `_gid_assign`. I think ideally `NetworkBuilder` should take `rank` in the `__init__` that is passed from `mpi_child.py` and then `_clone_and_simulate`. Then inside the `__init__` you do:

```py
self.rank = rank
```

and use `self.rank` inside `_gid_assign` instead of `_get_rank`. This will make it clear that `NetworkBuilder` is actually on a particular node. That way you will also limit nested calls to Neuron which we know can behave somewhat arbitrarily. Probably for another PR though

hnn_core/network.py
cjayb: Well I guess there are plenty of inconsistencies in the variable naming, as there are in the comments. Not sure how to address that, other than keep on cleaning up? What would you call the cells of the network? "Cells", I suppose, though in terms of a network, they are just the "nodes" of a graph.

hnn_core/network.py
cjayb: Totally. Here I'm using a really _bad_ variable name (`_tmp...`) to remind me to get rid of it! The whole idea of this series of PRs is to get rid of 'feeds' and move to 'drives'. Semantics aside, the variable here turned out to be a list-of-dict-of-list-of-list---not sure that's the way to go :) Much of the work here is in order to get the tests to pass, _i.e._, match 'HNN Classic' models.

hnn_core/cell.py
cjayb: Well this is a little stronger than hiding, no? I'm not sure the `Cell` gid ever needs to be updated, certainly not by users! Once it gets set, it should raise all hell if anything tries to change it. However, it would be nice to be able to _get_ the value using `cell.gid`, hence the `@property`. I know this complicates the `_Cell` class a little, but may make reading downstream code clearer (see `network_builder:L431-440`)?

hnn_core/network_builder.py
cjayb: To my surprise too: No! Turns out L398 is the key. The for-loop there must be on `self.net._gid_list`, which contains precisely those cells (gids) that exist on _this rank/host_. So the `_gid_assign` (note the round-robin for-loop) essentially 'picks out' the gids that belong to the present rank, after which `create_...` instantiates them _here_.

hnn_core/network_builder.py
cjayb: I like it, will keep in mind, as I'm pretty sure it'll belong to this PR.

hnn_core/network_builder.py
cjayb: +1 on name for now, might change once it becomes clearer how to implement drives later.
added the param, as it's trivially easy here (much harder in `_create_cells...`)

hnn_core/network.py
cjayb: OK. I'm also still wrapping my head around this. Was stoked that I managed to instantiate all `event_times`  prior to `NetworkBuilder`! Wasn't easy, but now `network_builder` no longer does _any_ random number generation at all.

cjayb:
> I think the next step is to write tests for the new methods. This way you'll know --> a) they work as you expect b) they are modular enough to be part of tests

Thanks for the comments so far. I'm very much in favour of writing tests, but wonder if it might make sense to discuss where to go from here first?

This PR is already substantial, and it's a major step forward: no more `event_times` created in `NetworkBuilder`. NB now takes care of GID assignment, cell creation & connection etc. All 'feeds' are created in `Network`, so they can be e.g. visualised prior to building & solving the network itself.

How about I take a break here & add some tests to `test_network` on expected `event_times` when using the default/testing parameters? But kinda leave it at that? The reason is that I have no intention of leaving `network.py` in the sorry state it is in! I want to implement a totally new API for getting the `event_times` instantiated, so making tests for the current bloody mess seems like an effort misplaced?

My proposal is: change the scope of this PR to the present (plus a few more tests). Then merge & move the remaining aims to a new branch. WDYT?

cjayb:
Um so the rebase was a mess:
https://github.com/cjayb/hnn-core/pull/new/network_drive_pr_rebase
OK to force-push here? I guess the review history for this PR will be lost?

Obviously, I think my changes to `cell` and `pyramidal` are improvements, such as being explicit about `override_params` in `__init__` etc. Note that with the last commit (https://github.com/jonescompneurolab/hnn-core/commit/34dbddbba7909c790fe3f0134f838e991e7f4dea), `_create_cells_and_feeds` and `_gids_to_parallel_context` no longer rely on `params`!

cjayb:
Um so the rebase was a mess:
https://github.com/cjayb/hnn-core/pull/new/network_drive_pr_rebase
OK to force-push here? I guess the review history for this PR will be lost?

Obviously, I think my changes to `cell` and `pyramidal` are improvements, such as being explicit about `override_params` in `__init__` etc. Note that with the last commit (), `_create_cells_and_feeds` and `_gids_to_parallel_context` no longer rely on `params`!

jasmainak:
Feel free to do a force push! I will take a look at the PR later in the
afternoon

hnn_core/network.py
```diff
@@ -135,34 +136,49 @@ class Network(object):
...
         The names of external drivers ('feeds') to the network (e.g. 'evdist1')
     gid_dict : dict
-        Dictionary with keys 'evprox1', 'evdist1' etc.
-        containing the range of Cell IDs of different cell
-        (or input) types.
+        A dictionary of unique identifiers of each real and artificial cell
+        in the network. Every cell type is represented by a key read from
+        cellname_list, followed by entries read from feedname_list. The value
+        of each key is a range of ints, one for each cell in given category.
+        Examples: 'L2_basal': range(0, 270), 'evdist1': range(272, 542), etc
```
jasmainak, line 154: `L2_basket` you mean?

hnn_core/network.py
```diff
@@ -135,34 +136,49 @@ class Network(object):
...
     def __init__(self, params):
         # Save the parameters used to create the Network
         self.params = params
+        # Initialise a dictionary of cell ID's, which get used when the
+        # network is constructed ('built') in NetworkBuilder
+        # We want it to remain in each Network object, so that the user can
+        # interrogate a built and simulated net. In addition, Spikes are
+        # attached to a Network during simulation---Network is the natural
+        # place to keep this information
+        # XXX rename gid_dict in future
```
jasmainak, line 172: let's just do it in this PR. Seems straightforward. `gid_ranges` ?

hnn_core/network_builder.py
```diff
@@ -346,109 +365,89 @@ def _gid_assign(self):
...
+            src_type, src_pos, is_cell = self.net._get_src_type_and_pos(gid)
 
             if is_cell:  # not a feed
                 # figure out which cell type is assoc with the gid
                 # create cells based on loc property
-                # creates a NetCon object internally to Neuron
-                type2class = {'L2_pyramidal': L2Pyr, 'L5_pyramidal': L5Pyr,
-                              'L2_basket': L2Basket, 'L5_basket': L5Basket}
-                Cell = type2class[src_type]
+                # XXX the cell subclasses could use cleaning up!
```
jasmainak, line 415: XXX is reserved for more specific comments. Usually, we go over them before a release and see if some of them can be knocked off easily.

jasmainak:
@cjayb okay can we bring the PR to a mergeable state? I'd be in favor of moving forward in small steps. Some tests can wait for later

hnn_core/network.py
jasmainak: I understand, but I'd still avoid bad variable names and call it `_feed_event_times` :) Who knows maybe you get busy after this PR. Also I will insist on the docstring ;-)

hnn_core/network.py
```diff
@@ -251,8 +251,58 @@ def _add_external_feeds(self):
         unique_keys = sorted(self.p_unique.keys())
         self.feedname_list += unique_keys
 
-    def _create_gid_dict(self):
-        """Creates gid dicts."""
+        self._update_gid_dict()  # add feeds -> update the gid_dict
+        self.instantiate_feeds()
+
+    def instantiate_feeds(self):
```
jasmainak, line 257: docstring missing for public method

hnn_core/network.py
```diff
@@ -251,8 +251,58 @@ def _add_external_feeds(self):
...
+        # each trial needs unique event time vectors
+
+        self._tmp_feed_event_times = []  # reset if called again
```
jasmainak, line 260: this is code smell. Was it already there?

hnn_core/network.py
jasmainak: Just to clarify, this will only add or remove keys, not modify the values, right? E.g., if I have `L2_Pyramidal: range(0, 100)` and `L5_Pyramidal: range(100, 200)`, do you imagine each of the key-value pairs to be updated when new cells are added?

hnn_core/network.py
```diff
@@ -251,8 +251,58 @@ def _add_external_feeds(self):
...
+                        target_cell_type = self.gid_to_type(gid_target)
+                        feed = ExtFeed(feed_type=src_type,
+                                       target_cell_type=target_cell_type,
+                                       params=p_unique[src_type],
+                                       gid=gid)
+                        event_times.append(feed.event_times)
+                event_times_per_source.update({src_type: event_times})
+            self._tmp_feed_event_times.append(event_times_per_source.copy())
+
+    def _update_gid_dict(self):
```
jasmainak, line 299: pass arguments explicitly?

jasmainak:
@cjayb can you rebase and bring the PR to MRG state? I think we're going in the right direction !

hnn_core/network.py
cjayb: I do :)

hnn_core/network_builder.py
cjayb: OK, I'll go through the ones I've left and decide either to do something about or remove.

hnn_core/network.py
cjayb: done

hnn_core/network.py
cjayb: Nothing was already here: all this was hidden deep in `network_builder` nested loops! I've added a docstring to mention the parameter-override `n_jobs` in `dipole.py:simulate` that makes this necessary.

hnn_core/network.py
cjayb: made it private, but added docstring anyway ;)

hnn_core/network.py
cjayb: -1 The reliance on `pos_dict` for all inference is weird and should be fixed. I think it's more than just changing the calling pattern of this method.

jasmainak:
@cjayb looks like you have rebase conflicts. Could you fix them? I'm fetching your PR and looking at it in the meanwhile.

hnn_core/dipole.py
```diff
@@ -42,6 +42,7 @@ def simulate_dipole(net, n_trials=None):
 
     if n_trials is not None:
         net.params['N_trials'] = n_trials
+        net._instantiate_feeds()  # need to redo these if n_trials changed!
```
jasmainak, line 45: ughgh ... how will this work when you instantiate the feeds outside of `simulate_dipole` ? do you have any ideas?

hnn_core/network.py
```diff
@@ -228,42 +219,106 @@ def _add_external_feeds(self):
...
-    def _create_gid_dict(self):
-        """Creates gid dicts."""
+        # Every time pos_dict is updated, gid_ranges must be updated too
+        self._update_gid_ranges()
+
+        # Create the feed dynamics (event_times)
+        self._instantiate_feeds()
+
+    def _instantiate_feeds(self):
+        '''Creates event_time vectors for all feeds and all trials
```
jasmainak, line 253: nitpick, but you should use """ instead of ''' for multiline docstrings

hnn_core/network.py
```diff
@@ -228,42 +219,106 @@ def _add_external_feeds(self):
...
+        self.trial_event_times = []  # reset if called again from dipole.py
+        n_trials = self.params['N_trials']
+
+        cur_params = self.params.copy()  # these get mangled below!
+        for trial_idx in range(n_trials):
+
+            prng_seedcore_initial = cur_params['prng_*'].copy()
+            for param_key in prng_seedcore_initial.keys():
+                cur_params[param_key] =\
+                    prng_seedcore_initial[param_key] + trial_idx
```
jasmainak, line 271: we should rethink this approach now that a lot of `feed.py` code is simplified in #204 and we begin to understand what that code is doing. I tagged you in #113

hnn_core/network_builder.py
jasmainak: But if cells are created _after_ the GIDs are assigned to different nodes, how will we allow users to add feeds? User code should not depend on how the GID assignment is done ...

hnn_core/network_builder.py
```diff
@@ -350,109 +359,77 @@ def _gid_assign(self):
...
                 # figure out which cell type is assoc with the gid
                 # create cells based on loc property
-                # creates a NetCon object internally to Neuron
-                type2class = {'L2_pyramidal': L2Pyr, 'L5_pyramidal': L5Pyr,
-                              'L2_basket': L2Basket, 'L5_basket': L5Basket}
-                Cell = type2class[src_type]
                 if src_type in ('L2_pyramidal', 'L5_pyramidal'):
-                    cell = Cell(gid, src_pos, params)
+                    PyramidalCell = type2class[src_type]
+                    # XXX Why doesn't a _Cell have a .threshold? Would make a
```
jasmainak, line 410: Not sure ... separate PR :)

jasmainak:
PR looks fine to me, happy to merge once CIs come green. @rythorpe do you want to take a look?

hnn_core/dipole.py
cjayb: Just checked: `n_trials` is only an argument to `simulate_dipole`. All other parts of code check `net.params['N_trials']` to figure out the same. So there are no other places to worry about wrt. instantiation of `event_times`.
I know this isn't pretty. I've made `n_trials` an explicit argument of `_instantiate_feeds` to make it clearer there's a dependency there.

hnn_core/network.py
cjayb: Why not merge #204 first, then rebase and fix this (more invasive) PR on that? I also predict  #113 will then just melt away :)

hnn_core/network_builder.py
cjayb: They can't. But that should be OK, no? We can create pretty visualisations of the input to a `net` before it ever gets built. So seeing the effects of adding and removing feeds (histograms etc.) can be done interactively, `_gid_assign` is not run until `Builder`. Note that the reliance on `self.net._p_unique` is something that the future implementation could make prettier, but right now it works as it is.

hnn_core/network_builder.py
cjayb: Wasn't hard, just added rank and n_hosts as attributes to `NetworkBuilder` objects.

cjayb: I take it back: something exploded so I rolled it back. Separate PR!

cjayb:
Had to squash to make rebasing feasible: #190 really made it hard!
The last commit (@da134c7) is quite invasive, so maybe @ntolley would like to take a look? Basically I don't see why `times` should be an attribute of a `Network`? I've moved it into `Spikes`, which seems to me to make more sense (though could be wrong?).
I feel quite strongly though about `times` being the same for all `n_trials`! I've made an in-line comment about it in `parallel_backends.py`, please take a look.

@jasmainak I still think #204 could be merged before this, but I don't feel like another rebase on this one, so hope we could move fast?

hnn_core/parallel_backends.py
```diff
@@ -56,10 +51,9 @@ def _gather_trial_data(sim_data, net, n_trials):
         spikedata = sim_data[idx][1]
         net.spikes._spike_times.append(spikedata[0])
         net.spikes._spike_gids.append(spikedata[1])
-        net.gid_dict = spikedata[2]  # only have one gid_dict
-        net.spikes.update_types(net.gid_dict)
+        net.gid_ranges = spikedata[2]  # only have one gid_ranges
+        net.spikes.update_types(net.gid_ranges)
         net.spikes._vsoma.append(spikedata[3])
-        net.spikes._times.append(net.times.tolist())
```
cjayb, line 62: Is there any reason different trials should be allowed to have different `times`?!

jasmainak:
yes, let's move fast. But another set of eyes other than me would be good. @ntolley or @rythorpe perhaps?

hnn_core/network.py
```diff
@@ -381,7 +446,10 @@ def __init__(self, spike_times=None, spike_gids=None, spike_types=None):
         self._spike_gids = spike_gids
         self._spike_types = spike_types
         self._vsoma = list()
-        self._times = list()
+        if times is not None:
+            assert isinstance(times, np.ndarray),\
+                "'times' is an array of simulation times"
```
jasmainak, line 451: raise a proper `TypeError` ...

jasmainak (approved):

hnn_core/cell.py
rythorpe: Hmm, I'm not sure why we need 2 gid attributes here (i.e., `_assigned_gid` and `gid`). Why can't we just have one attribute named `_ArtificialCell._gid` that is set to the arg `gid` using your decorated setter function, and is retrieved by your decorated property function? Unless I'm missing something nuanced in the logic, such a change would simplify the gid code in this class as well as subsequent cell classes.

cjayb: ah, maybe a little convoluted there yes. I just pushing something now, which I instantly regretted :( Ignore first push, my proposal will be the second one! To answer your question: I'm proposing to protect `.gid` because we don't want to risk anything messing with a `Cell`'a gid once building has begun. Perhaps there's a better way of achieving this?

rythorpe: What I'm imagining is something like

```python
def __init__(self, event_times, threshold, gid=None):
   ...
   self.gid = gid

@gid.setter
def gid(self, gid)
   if self.gid is None:
      self.gid = gid
   else:
      raise RuntimeError('Global ID for this cell already assigned!')
```

cjayb: well, I'm not sure it's getting any better (7d16670)? If not, could you explain your thoughts further? I like the idea of a `Cell` sporting its `gid` for downstream use. There's some duplication in the code now, though, so if you agree on something like this,  might have to create a superclass for `_Cell` and `_ArtificialCell` to inherit from?

rythorpe: Oh sorry, I didn't refresh my browser and made my previous comment before seeing your most recent commit. I'm just trying to eliminate the need for having two attributes that store the gid in order to make the logic more readable. Correct me if I'm wrong, but the only consideration here is that the gid of a given cell should only be assigned to a value other than `None` once. Since this logic can be contained purely within a gid setter function, I don't understand why we need a `self.gid` and `self._gid` attribute.

cjayb: Same here with the browser refresh ;) I'm trying your proposal, but `test_cell` hangs? I would be surprised if you could do `self.gid = gid`, isn't that trying to call the 'getter' and setter at the same time?! I've seen the `_`-prefix in cases like this elsewhere, thought it as a `pattern`...

cjayb: > Correct me if I'm wrong, but the only consideration here is that the gid of a given cell should only be assigned to a value other than `None` once.

That's my goal, yes.

hnn_core/network_builder.py
rythorpe: Just to clarify, @cjayb it sounds like what you're saying is that all aspects of the feed/network that we intend the user to interact with should be defined prior to `NetworkBuilder`? If so, this makes sense since the nrn interface should be invisible to the user.

cjayb: Yes. That wasn't really my goal but it mostly just became so when extracting the feed instantiation into `Network`. Now I think it's a good goal though, as you say ;)

cjayb: Note that I'm not implying that GID handling is "solved" quite yet. Baby steps... One of them being the protected cell attribute.

hnn_core/cell.py
jasmainak: let me shed some light here. You can use private attributes for stuff that you don't want the user to mess with. For example if the object is complex and updating one attribute actually affects the rest of the object. For instance if the object had an attribute called `sfreq` (sampling frequency) and another attribute called `times`, you would have to update both simultaneously. This is why it sometimes makes sense to protect certain attributes and update them only through special functions.

rythorpe: > Same here with the browser refresh ;) I'm trying your proposal, but `test_cell` hangs? I would be surprised if you could do `self.gid = gid`, isn't that trying to call the 'getter' and setter at the same time?!

Shoot, you're right. I think your implementation in 7d16670 is our best bet.

hnn_core/parallel_backends.py
rythorpe: I'm guessing that the goal here was to maintain consistent formatting between the primary attributes of the `Spikes` class, but you're right that this is probably unnecessary. Any considerations we're missing @ntolley?

rythorpe (approved):
Okay, I think I now have a decent grasp of this PR. (Sorry for being so absent; as soon as I get this manuscript I've been working on submitted I'll be much more present.) Your method for creating jittered feed times all within `Network.instantiate_feeds()` such that they can be referenced on a trial-by-trial basis in `NetworkBuilder` and across parallel backends will be super useful as we move forward. Thanks @cjayb!

jasmainak:
It's in, thanks @cjayb ! :-) Let's merge #204 next so you can make your next PR on top of it

hnn_core/cell.py
cjayb: Can always be changed to something smoother later.

cjayb:
Cool! Thanks both, good to get this out of the way :) @jasmainak what's the best way I can help rebase #204?

jasmainak:
@cjayb I am working on my rebase and realized that `trial_event_times` should have `dict()` as the outermost level. Then inside you can probably have numpy arrays instead of "list of list" ... might be the cleanest.

cjayb:
A dict with which keys though? I don't have a strong opinion there so feel free to suggest a change in #204!

jasmainak:
It should be dict (keys are the src_type) of list (gids) of list (n_trials) of list (n_events) ? I think it will help unify things related to trial jitter etc.

hnn_core/parallel_backends.py
ntolley: Consistency was the main motivation, the way it is written should not allow for different `times` across trials. `times` is actually created from directly from `params` when the network is instantiated, this function is really just copying the data.

Not really a big deal to change and probably cleaner this way. Most important thing is for the examples to demonstrate how to use the variable in plotting.

PR #199, opened 2020-10-22
[MRG] Add functionality to record current from all cells

### LINKED ISSUE #187: Make cell currents a model output
HNN GUI saves cell currents for 'L2Pyr_soma' and  'L5Pyr_soma' to the file `i.txt`, but doesn't currently plot them. Adding the functionality to hnn-core to calculate these currents would allow us to explore ways to represent/plot this data in the future.

Function to record currents (called for L2Pyr and L5Pyr):
https://github.com/jonescompneurolab/hnn/blob/bd2abe4007160a5186856bc17dd87ae62055b383/cell.py#L180-L194

Needs to be aggregated for each type of cell: https://github.com/jonescompneurolab/hnn/blob/bd2abe4007160a5186856bc17dd87ae62055b383/network.py#L370-L387

The above function is called after a simulation completes:
https://github.com/jonescompneurolab/hnn/blob/d4c4cf94aa1db8517b99b0e344aa852c16125f50/run.py#L435

After this functionality exists to retrieve currents as a model output (same mechanism as somatic voltages?), the code to save the output to `i.txt` can be added back to HNN GUI. It would be a cleaner version of this code:
https://github.com/jonescompneurolab/hnn/blob/d4c4cf94aa1db8517b99b0e344aa852c16125f50/run.py#L140-L145

ntolley:
closes #187. Starting this PR early to keep track of modifications discussed in #193 

**To-do** 
- [x] Merge `record_current()` and `record_voltage()` into single function
- [x] Merge `aggregate_currents()` and `aggregate_voltages()` into single function
- [x] ~~Modify dipole calculation to utilize individual cell current recordings?~~
- [x] Add tests to check current/voltage waveform recordings agree with `spiketimes`
- [x] Rename `Spikes` class to `CellResponse`

### DIFF
--- a/examples/plot_firing_pattern.py
+++ b/examples/plot_firing_pattern.py
@@ -64,6 +64,7 @@ for idx in range(10):  # only 10 cells per cell-type
     axes[0].plot(net.cell_response.times, vsoma[gid], color='g')
     gid = gid_ranges['L5_pyramidal'][idx]
     axes[0].plot(net.cell_response.times, vsoma[gid], color='r')
-net.cell_response.plot(ax=axes[1])
-net.cell_response.plot_hist(ax=axes[2],
-                            spike_types=['L5_pyramidal', 'L2_pyramidal'])
+net.cell_response.plot_spikes_raster(ax=axes[1])
+net.cell_response.plot_spikes_hist(ax=axes[2],
+                                   spike_types=['L5_pyramidal',
+                                                'L2_pyramidal'])
--- a/examples/plot_firing_pattern.py
+++ b/examples/plot_firing_pattern.py
@@ -39,6 +39,8 @@ print(net.gid_ranges)
 
 ###############################################################################
 # Simulated voltage in the soma is stored in CellResponse as a dictionary.
+# The CellResponse object stores data produced by individual cells including
+# spikes, somatic voltages and currents.
 trial_idx = 0
 vsoma = net.cell_response.vsoma[trial_idx]
 print(vsoma.keys())
--- a/examples/plot_firing_pattern.py
+++ b/examples/plot_firing_pattern.py
@@ -40,14 +40,14 @@ print(net.gid_ranges)
 ###############################################################################
 # Simulated voltage in the soma is stored in CellResponse as a dictionary.
 trial_idx = 0
-vsoma = net.spikes.vsoma[trial_idx]
+vsoma = net.cell_response.vsoma[trial_idx]
 print(vsoma.keys())
 
 ###############################################################################
 # We can plot the firing pattern of individual cells by indexing with the gid
 gid = 170
 plt.figure(figsize=(4, 4))
-plt.plot(net.spikes.times, vsoma[gid])
+plt.plot(net.cell_response.times, vsoma[gid])
 plt.title('%s (gid=%d)' % (net.gid_to_type(gid), gid))
 plt.xlabel('Time (ms)')
 plt.ylabel('Voltage (mV)')
@@ -61,8 +61,9 @@ fig, axes = plt.subplots(3, 1, figsize=(5, 7), sharex=True)
 
 for idx in range(10):  # only 10 cells per cell-type
     gid = gid_ranges['L2_pyramidal'][idx]
-    axes[0].plot(net.spikes.times, vsoma[gid], color='g')
+    axes[0].plot(net.cell_response.times, vsoma[gid], color='g')
     gid = gid_ranges['L5_pyramidal'][idx]
-    axes[0].plot(net.spikes.times, vsoma[gid], color='r')
-net.spikes.plot(ax=axes[1])
-net.spikes.plot_hist(ax=axes[2], spike_types=['L5_pyramidal', 'L2_pyramidal'])
+    axes[0].plot(net.cell_response.times, vsoma[gid], color='r')
+net.cell_response.plot(ax=axes[1])
+net.cell_response.plot_hist(ax=axes[2],
+                            spike_types=['L5_pyramidal', 'L2_pyramidal'])
--- a/examples/plot_firing_pattern.py
+++ b/examples/plot_firing_pattern.py
@@ -38,7 +38,7 @@ gid_ranges = net.gid_ranges
 print(net.gid_ranges)
 
 ###############################################################################
-# Simulated voltage in the soma is stored in the Spikes object as a dictionary.
+# Simulated voltage in the soma is stored in CellResponse as a dictionary.
 trial_idx = 0
 vsoma = net.spikes.vsoma[trial_idx]
 print(vsoma.keys())
--- a/examples/plot_simulate_evoked.py
+++ b/examples/plot_simulate_evoked.py
@@ -53,26 +53,28 @@ with JoblibBackend(n_jobs=1):
 import matplotlib.pyplot as plt
 fig, axes = plt.subplots(2, 1, sharex=True, figsize=(6, 6))
 plot_dipole(dpls, ax=axes[0], layer='agg', show=False)
-net.spikes.plot_hist(ax=axes[1], spike_types=['evprox', 'evdist'])
+net.cell_response.plot_hist(ax=axes[1], spike_types=['evprox', 'evdist'])
 
 ###############################################################################
 # Also, we can plot the spikes and write them to txt files.
 # Note that we can use formatting syntax to specify the filename pattern
 # with which each trial will be written. To read spikes back in, we can use
 # wildcard expressions.
-net.spikes.plot()
+net.cell_response.plot()
 with tempfile.TemporaryDirectory() as tmp_dir_name:
-    net.spikes.write(op.join(tmp_dir_name, 'spk_%d.txt'))
-    spikes = read_spikes(op.join(tmp_dir_name, 'spk_*.txt'))
-spikes.plot()
+    net.cell_response.write(op.join(tmp_dir_name, 'spk_%d.txt'))
+    cell_response = read_spikes(op.join(tmp_dir_name, 'spk_*.txt'))
+cell_response.plot()
 
 ###############################################################################
 # We can additionally calculate the mean spike rates for each cell class by
 # specifying a time window with tstart and tstop.
-all_rates = spikes.mean_rates(tstart=0, tstop=170, gid_ranges=net.gid_ranges,
-                              mean_type='all')
-trial_rates = spikes.mean_rates(tstart=0, tstop=170, gid_ranges=net.gid_ranges,
-                                mean_type='trial')
+all_rates = cell_response.mean_rates(tstart=0, tstop=170,
+                                     gid_ranges=net.gid_ranges,
+                                     mean_type='all')
+trial_rates = cell_response.mean_rates(tstart=0, tstop=170,
+                                       gid_ranges=net.gid_ranges,
+                                       mean_type='trial')
 print('Mean spike rates across trials:')
 print(all_rates)
 print('Mean spike rates for individual trials:')
@@ -97,4 +99,4 @@ with MPIBackend(n_procs=2, mpi_cmd='mpiexec'):
     dpls_sync = simulate_dipole(net_sync, n_trials=1)
 
 dpls_sync[0].plot()
-net_sync.spikes.plot_hist()
+net_sync.cell_response.plot_hist()
--- a/examples/plot_simulate_evoked.py
+++ b/examples/plot_simulate_evoked.py
@@ -53,18 +53,19 @@ with JoblibBackend(n_jobs=1):
 import matplotlib.pyplot as plt
 fig, axes = plt.subplots(2, 1, sharex=True, figsize=(6, 6))
 plot_dipole(dpls, ax=axes[0], layer='agg', show=False)
-net.cell_response.plot_hist(ax=axes[1], spike_types=['evprox', 'evdist'])
+net.cell_response.plot_spikes_hist(ax=axes[1],
+                                   spike_types=['evprox', 'evdist'])
 
 ###############################################################################
 # Also, we can plot the spikes and write them to txt files.
 # Note that we can use formatting syntax to specify the filename pattern
 # with which each trial will be written. To read spikes back in, we can use
 # wildcard expressions.
-net.cell_response.plot()
+net.cell_response.plot_spikes_raster()
 with tempfile.TemporaryDirectory() as tmp_dir_name:
     net.cell_response.write(op.join(tmp_dir_name, 'spk_%d.txt'))
     cell_response = read_spikes(op.join(tmp_dir_name, 'spk_*.txt'))
-cell_response.plot()
+cell_response.plot_spikes_raster()
 
 ###############################################################################
 # We can additionally calculate the mean spike rates for each cell class by
@@ -99,4 +100,4 @@ with MPIBackend(n_procs=2, mpi_cmd='mpiexec'):
     dpls_sync = simulate_dipole(net_sync, n_trials=1)
 
 dpls_sync[0].plot()
-net_sync.cell_response.plot_hist()
+net_sync.cell_response.plot_spikes_hist()
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -107,7 +107,7 @@ class _Cell(ABC):
         self.soma_props = soma_props
         self.create_soma()
         self.rec_v = h.Vector()
-        self.rec_i= dict()
+        self.rec_i = dict()
         self._gid = None
         if gid is not None:
             self.gid = gid  # use setter method to check input argument gid
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -234,7 +234,17 @@ class _Cell(ABC):
         self.dipole = h.Vector().record(self.dpl_ref)
 
     def record_soma(self, record_vsoma=False, record_isoma=False):
-        """Record current and voltage at soma."""
+        """Record current and voltage at soma.
+
+        Parameters
+        ----------
+        record_vsoma : bool
+            Option to record somatic voltages from cells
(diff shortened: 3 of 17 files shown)


hnn_core/dipole.py
```diff
@@ -58,6 +62,12 @@ def simulate_dipole(net, n_trials=None, record_vsoma=False):
         raise TypeError("record_vsoma must be bool, got %s"
                         % type(record_vsoma).__name__)
 
+    if record_isoma is not None and isinstance(record_isoma, bool):
+        net.params['record_isoma'] = record_isoma
```
jasmainak, line 66: is there such a parameter in the `params` file? If not, let's not invent it ...

ntolley: So this was posed as the workaround for enabling/disabling `vsoma` recordings. Do you think this should be an attribute of the `Network` instead?

rythorpe: My vote is to leave `record_isoma` as an argument of `simulate_dipole()` without creating a new param. Same with `record_vsoma`, however, since it's already an argument in the default param file as well as the GUI (right?), I guess it's fine to leave `record_vsoma` in there until after GUI integration. I'm on an anti-params-file kick at the moment so that we can hopefully refocus and simplify the use-case of parameters stored in param files :)

ntolley: Ah I missed this. `save_vsoma` is contained inside `default.json` but I think that's just leftover from copying the GUI params directly. 

Currently this is just an awkward way to pass `record_vsoma` to the `network_builder.py`. The params variable is just a convenient way to pass these variables, but neither is specified in the actual `.params` file. 

I'm not a fan of adding extra arguments to `_BACKEND.simulate(net)` so the alternative would be adding an attribute to `Network` which can be accessed by `network_builder.py` later on.

jasmainak: we have a bit of a mess with parallel backends. Maybe you get away with `net.params` now. But we'll have to bite the bullet at some point and clean it up. If we don't touch the code, we won't realize how to improve it :)

rythorpe: Oh I see. Yeah that will be a bit of a mess, but I feel like `_BACKEND.simulate(net)` is the appropriate place for arguments that determine the types of data that are output/saved in a simulation. Any specific arguments against this?

ntolley: Fair point, my only hesitancy is tacking on a bunch of arguments to the critical functions. More of a clean/readability concern, but these seem sufficiently important.

ntolley: Decided to go with making them attributes of `Network` since they can be updated once and readout by `network_builder.py`. Realized that passing them as arguments would require a lot of copy-pasted code as I tried it.

ntolley:
Inspired by a talk from Eve Marder, she showed off a really nice visualization of ionic current flow for single cells:
image

Thinking of adding something similar to `viz.py`. I think this is particularly relevant to characterizing degenerate parameter sets that produce the same current dipole.

jasmainak:
go for it!

ntolley:
Quick update: I've finished merging the recording functions. For `cell.py` I kept `insert_dipole()` separate from `record_soma()` as the former is significantly more complex. The somatic voltage and current recordings are handled by `record_soma()`. 

However in `network_builder.py` I've combined the dipole, current, and voltage aggregation into a single `aggregate_data()` function as they all require looping over every gid.

hnn_core/cell.py
```diff
@@ -228,28 +233,38 @@ def insert_dipole(self, yscale):
...
-            print(
-                "Warning in Cell(): record_current_soma() was called,"
-                " but no self.synapses dict was found")
-            pass
-
-    def record_voltage_soma(self):
-        """Record voltage at soma."""
-        self.rec_v.record(self.soma(0.5)._ref_v)
+        if record_isoma:
+            try:
```
jasmainak, line 250: would something bad happen if you remove the `try`-`except` ? If not, I'd get rid of it

hnn_core/network.py
```diff
@@ -29,8 +29,8 @@ def read_spikes(fname, gid_ranges=None):
 
     Returns
     ----------
-    spikes : Spikes
-        An instance of the Spikes object.
+    spikes : CellResponse
```
jasmainak, line 32:
```suggestion
   cell_response : CellResponse
```

?

hnn_core/network_builder.py
```diff
@@ -631,27 +625,14 @@ def _record_spikes(self):
...
+        """Aggregate somatic currents, voltages, and dipoles."""
         for cell in self.cells:
             if cell.celltype in ('L5_pyramidal', 'L2_pyramidal'):
                 self.dipoles[cell.celltype].add(cell.dipole)
 
-    def aggregate_voltages(self):
-        """Organize voltage recordings into dictionary indexed by gid"""
-        for cell in self.cells:
             self._vsoma[cell.gid] = cell.rec_v
+            self._isoma[cell.gid] = cell.dict_currents
```
jasmainak, line 635: dict_currents -> rec_i ?

hnn_core/cell.py
ntolley: I don't think so... this is code left over from the original current recording setup. If nothing changes I'll convert it to mimic the voltage recording naming schemes (including the variable name changes below).

hnn_core/network_builder.py
ntolley: May as well. It is a dictionary of `h.Vector()` recordings unlike `vsoma` but anyone who is digging around this code can quickly figure that out.

hnn_core/network.py
ntolley: I'll try it out. This rename will change quite a lot of the code, but we may as well keep the naming consistent until we think of a better alternative.

ntolley:
@jasmainak @rythorpe I've pretty satisfied with the changes so far, definitely feel free to suggest modifications. Once this is merged I'll start a PR converting the `CellResponse` attribute into a list stored by `Network`.

Also some background for voltage/spike timing test in `test_dipole.py`. It's essentially seeing if the spike times fall in the range of increased somatic voltage for a particular gid.

examples/plot_simulate_evoked.py
```diff
@@ -53,26 +53,28 @@
...
-net.spikes.plot_hist(ax=axes[1], spike_types=['evprox', 'evdist'])
+net.cell_response.plot_hist(ax=axes[1], spike_types=['evprox', 'evdist'])
 
 ###############################################################################
 # Also, we can plot the spikes and write them to txt files.
 # Note that we can use formatting syntax to specify the filename pattern
 # with which each trial will be written. To read spikes back in, we can use
 # wildcard expressions.
-net.spikes.plot()
+net.cell_response.plot()
```
jasmainak, line 63: humm ... now it seems this should be `plot_spikes`. We need to make all this consistent, but I'm okay if you want to push it to another PR

hnn_core/tests/test_dipole.py
```diff
@@ -61,7 +61,21 @@ def test_dipole_simulation():
         simulate_dipole(net, n_trials=0)
     with pytest.raises(TypeError, match="record_vsoma must be bool, got int"):
         simulate_dipole(net, n_trials=1, record_vsoma=0)
+    with pytest.raises(TypeError, match="record_isoma must be bool, got int"):
+        simulate_dipole(net, n_trials=1, record_vsoma=False, record_isoma=0)
 
-    simulate_dipole(net, n_trials=2, record_vsoma=True)
-    assert len(net.spikes.vsoma) == 2
-    assert len(net.spikes.vsoma[0]) == 24
+    simulate_dipole(net, n_trials=2, record_vsoma=True, record_isoma=True)
```
jasmainak, line 67: does this work for both the backends? Can you add a test to check that?

hnn_core/dipole.py
```diff
@@ -52,12 +56,18 @@ def simulate_dipole(net, n_trials=None, record_vsoma=False):
     if n_trials < 1:
         raise ValueError("Invalid number of simulations: %d" % n_trials)
 
-    if record_vsoma is not None and isinstance(record_vsoma, bool):
-        net.params['record_vsoma'] = record_vsoma
+    if isinstance(record_vsoma, bool):
+        net._record_vsoma = record_vsoma
```
jasmainak, line 60: I'm a bit suspicious if this will work for MPI backend.

ntolley: Really good catch, my local machine was defaulting to the JobLib backend since I didn't specify `n_procs`. MPI produces returns data identical to `record_vsoma=False`. 

Not sure I totally understand the behavior but will dig in tomorrow. Do you have a grasp of the issue? I'm guessing `net._record_vsoma` is not persisting after being passed to the backend.

jasmainak: my hunch is that the `MPIBackend` only passes the `params` dictionary not the whole object. I think some effort spent in cleaning up the backends and codepaths would hold us in good stead in the long run.

ntolley: Think I found what you're talking. The params are pickled in `parallel_backends.py` [here](https://github.com/jonescompneurolab/hnn-core/blob/b50f91dcf7c0cdeae87189c9a1dc97f56bddc0b4/hnn_core/parallel_backends.py#L356) and `mpi_child.py` instantiates a new network based solely on the `params` [here](https://github.com/jonescompneurolab/hnn-core/blob/b50f91dcf7c0cdeae87189c9a1dc97f56bddc0b4/hnn_core/mpi_child.py#L103). Short term solution would be reverting back to tacking `record_vsoma` and `record_isoma` onto the `params` dictionary. 

Can you open an issue/PR for a discussion on what needs to be changed for the backends? I don't feel I have a good sense of what's good or bad about the current implementation.

jasmainak: Take a look at #142 but you need to muck with the code to realize the limitations/issues. We can't depend on one person for that code. For example, joblib backend allows any picklable object, not just `params` (will open an issue and tag you). Try passing the `record_vsoma` to `_Backend.simulate()` and you'll see how nested and deep the calls quickly get. Shunting through `net.params` is [code smell](https://en.wikipedia.org/wiki/Code_smell)

ntolley: Do you want to pause merging this PR until the parallel issues are addressed? I agree it's not the best solution but I don't see clear workaround besides hardcoding the parameter passing.

jasmainak: no, let's move on. Are you able to make it work for both backends though?

ntolley: Yes they both work locally after reverting back to passing through `params`. I've added explicit tests for both backends in `test_dipole.py` (L68). They're a bit over complete as both involve checking the size of the recorded output, let me know if you think it should be cut down.

examples/plot_simulate_evoked.py
ntolley: Renamed it to `plot_spikes_raster()`. Additionally changed `plot_hist()` to `plot_spikes_hist()` for consistency.

examples/plot_firing_pattern.py
```diff
@@ -38,16 +38,16 @@
 print(net.gid_ranges)
 
 ###############################################################################
-# Simulated voltage in the soma is stored in the Spikes object as a dictionary.
+# Simulated voltage in the soma is stored in CellResponse as a dictionary.
```
jasmainak, line 41: can you say a bit about the `CellResponse` object here? What does it contain?

hnn_core/dipole.py
jasmainak: do you mind leaving a comment where these parameters are used again? I guess in `cell.py` or `_simulate_single_trial`?

hnn_core/tests/test_dipole.py
```diff
@@ -61,7 +62,33 @@ def test_dipole_simulation():
...
         simulate_dipole(net, n_trials=1, record_vsoma=0)
+    with pytest.raises(TypeError, match="record_isoma must be bool, got int"):
+        simulate_dipole(net, n_trials=1, record_vsoma=False, record_isoma=0)
 
-    simulate_dipole(net, n_trials=2, record_vsoma=True)
-    assert len(net.spikes.vsoma) == 2
-    assert len(net.spikes.vsoma[0]) == 24
+    with MPIBackend(n_procs=None, mpi_cmd='mpiexec'):
+        simulate_dipole(net, n_trials=2, record_vsoma=True, record_isoma=True)
+        assert len(net.cell_response.vsoma) == 2
```
jasmainak, line 70: define `n_trials = 2` above and compare to `n_trials`

hnn_core/tests/test_dipole.py
```diff
@@ -61,7 +62,33 @@ def test_dipole_simulation():
...
+        assert len(net.cell_response.isoma) == 2
+        n_times = np.arange(0., params['tstop'] + params['dt'],
+                            params['dt']).size
+        assert len(net.cell_response.vsoma[0][8]) == n_times
```
jasmainak, line 74: what's 8?

jasmainak (approved):
PR looks okay for now, let's iterate more in future PRs. @rythorpe do you want to take a look and merge if you're happy?

don't forget to update `whats_new.rst` ?

jasmainak:
you might also need to update `api.rst`

hnn_core/dipole.py
```diff
@@ -30,6 +31,9 @@ def simulate_dipole(net, n_trials=None, record_vsoma=False):
         net.params['N_trials'] will be used
     record_vsoma : bool
         Option to record somatic voltages from cells
+            net.params['N_trials'] will be used
```
jasmainak, line 34: ?? something is missing ...

hnn_core/tests/test_dipole.py
```diff
@@ -61,7 +62,33 @@ def test_dipole_simulation():
...
+    with JoblibBackend(n_jobs=1):
+        simulate_dipole(net, n_trials=2, record_vsoma=True, record_isoma=True)
+        assert len(net.cell_response.vsoma) == 2
+        assert len(net.cell_response.isoma) == 2
+        n_times = np.arange(0., params['tstop'] + params['dt'],
+                            params['dt']).size
+        assert len(net.cell_response.vsoma[0][8]) == n_times
+        assert len(net.cell_response.isoma[0][8]['soma_gabaa']) == n_times
+
+    gid, v_thresh = 7, 0.0
```
jasmainak, line 87: add comment above L87 what this test is doing in plain English

hnn_core/tests/test_dipole.py
ntolley: It's just the gid `vsoma` is recorded from. I'll assign explicit variables along with `n_trials` above.

ntolley:
Glad I caught the MPI aggregation bug before merging... Previously the MPI backend would only store cell recordings from node 0. To avoid similar problems in the future I've added a test to explicitly compare the outputs of the MPI and JobLib backends.

@jasmainak what do you have in mind for `api.rst`?

jasmainak:
> @jasmainak what do you have in mind for api.rst?

don't you need to rename `Spikes` to `CellResponse` there?

jasmainak:
Other than `api.rst`, PR LGTM, I let @rythorpe merge if he's happy

hnn_core/network.py
```diff
@@ -398,14 +399,16 @@ class Spikes(object):
         Each gid corresponds to a type via Network::gid_ranges.
     vsoma : dict
         Dictionary indexed by gids containing somatic voltages
+    isoma : dict
+        Dictionary indexed by gids containing somatic currents
```
rythorpe, line 403: Aren't these actually lists (n_trials,) of dicts?

hnn_core/network.py
```diff
@@ -398,14 +399,16 @@ class Spikes(object):
...
     times : numpy array
```
rythorpe, line 404: In keeping with our previous convention to use built-in python objects when possible, could we change `times` to type list?

rythorpe: This is super nitpicky so we can also save this for the next PR if we want.

rythorpe (approved):
Aside from my 2 very minor comments re: the `CellResponse` docstring, this looks really good @ntolley! As soon as you respond, let's merge!

rythorpe:
Oh, and also update `api.rst`!

hnn_core/network.py
ntolley: Eh this is tough. There is a lot of code the presumes `times` is an ndarray for logical indexing and other conveniences. There is even an explicit test to ensure that `times` is instantiated as an array and not a list in [`test_network.py`](https://github.com/jonescompneurolab/hnn-core/blob/b50f91dcf7c0cdeae87189c9a1dc97f56bddc0b4/hnn_core/tests/test_network.py#L49). 

The change would just involved recasting with `np.array()` in a variety of placed around the code. I'm fine with changing to a list in this PR but I wonder if `times` is worth making an exception...

hnn_core/network.py
ntolley: Good catch! Updated.

hnn_core/network.py
jasmainak: I believe `times` is an array in MNE. It is indeed quite convenient in many places.

rythorpe: Let's leave it then.

rythorpe:
Thanks @ntolley!

jasmainak:
Great stuff! Next PR is to add the somatic current visualization to an example?

ntolley:
I was planning to leave that until after converting `CellResponse` to a list. Do you think that should wait until the remaining integration PR's are merged?

jasmainak:
I'm guessing the `CellResponse` PR will take a while to merge? So it should be fine to get started on that. Meanwhile we should try to get @kohl-carmen 's PR merged. @kohl-carmen has unwittingly stepped on the `MPIBackend` landmine ... hope to get it resolved soon.

PR #250, opened 2021-01-21
[MRG] Improve gamma example by adding new `viz` methods

cjayb:
This PR

* improves the narrative of the gamma example
* adds `viz.plot_tfr_morlet`: utility for plotting TFR of single-trial dpl time course (uses MNE and SciPy)
* adds `viz.plot_spectrogram`: utility for plotting Welch periodogram of single-trial dpl time course (uses SciPy)

Visualisation functions are poorly-tested, and input validation is largely missing. However, the readability of the example has improved dramatically, IMHO.


cjayb:
@jasmainak I don't get why the new `plot_tfr_morlet` and `plot_spectrogram` don't appear as hyperlinks in the examples? Some init file I need to include them in? They're in `api.rst`

@ntolley take a look at @2218c9c : you could use the new function in your alpha/beta demo?

For a 0.1-release, perhaps the missing parameter checks are an acceptable omission...?

hnn_core/viz.py
```diff
@@ -252,3 +292,141 @@ def plot_cells(net, ax=None, show=True):
...
+def plot_tfr_morlet(dpl, *, freqs, n_cycles=7., tmin=None, tmax=None,
+                    layer='agg', decim=False, ax=None,
+                    colorbar=True, show=True):
+    """Plot Morlet time-frequency representation of dipole time course
+
+    Parameters
+    ----------
+    dpl : instance of Dipole | list of Dipole instances
+        The Dipole object.
+    fregs : array
```
jasmainak, line 306: freqs

examples/plot_simulate_gamma.py
```diff
@@ -22,10 +35,12 @@
 hnn_core_root = op.dirname(hnn_core.__file__)
 
 ###############################################################################
-# Then we read the parameters file
+# Read the parameter file and print the between-cell connectivity parameters.
+# Note that these are different compared with the 'default' parameter set used
```
jasmainak, line 39: Actually they shouldn't be dramatically different. Maybe only the layer2-layer5 connections are turned off and the rest is same? I'd like to move away from importing different param files to initialize the network. Do you think using `default.param` would work?

hnn_core/viz.py
```diff
@@ -252,3 +292,141 @@ def plot_cells(net, ax=None, show=True):
...
+
+    data, times = _get_plot_data(dpl, layer, tmin, tmax)
+
+    sfreq = dpl.sfreq
+    if decim is not None:
+        data, times = _decimate_plot_data(decim, data, times)
+        sfreq = sfreq / decim
+
+    # mirror padding!
+    data = np.r_[data[-1:0:-1], data, data[-2::-1]]
```
jasmainak, line 343: there seems to be a strong edge effect now!

examples/plot_simulate_gamma.py
cjayb: They are: just compare `param['gbar_L*]`. The paper (Law & Jones, 2013) is pretty clear about it too. I haven't tried using the default parameters, though, could be interesting...

hnn_core/viz.py
cjayb: I noticed that, should I zero-pad instead?

jasmainak: maybe give it a quick try? Let's not be too distracted by this

examples/plot_simulate_gamma.py
jasmainak: indeed ... let's leave it like this then

examples/plot_simulate_gamma.py
```diff
@@ -44,44 +59,35 @@
...
+# representation together with the signal. Note that the network requires some
+# time to reach steady state. Hence, we omit the first 50 ms in our analysis.
 
-tstart = 50
+tmin = 50
 trial_idx = 0  # pick first trial
-mask = dpls[trial_idx].times > tstart
-times = dpls[trial_idx].times[mask]
-data = dpls[trial_idx].data['agg'][mask]
+decim = 8  # decimate plots by a factor of 8 (from 40 to 5 kHz)
```
jasmainak, line 76: do we really need to do this? I think the fewer the parameters ... the better

jasmainak (approved):
Other than minor nitpicks, it's good to go @cjayb ! Thanks for improving the example

examples/plot_simulate_gamma.py
```diff
@@ -3,13 +3,26 @@
...
-Replicates: https://jonescompneurolab.github.io/hnn-tutorials/gamma/gamma
+This example demonstrates how to simulate gamma rhythms generated by the "PING"
+mechanism using ``hnn-core`` [1,2]. Note that the parameter file used is
+different from other examples: the intrinsic connectivity of the network was
+tuned in [1] by adjusting the synaptic weights between excitatory (pyramidal)
+and inbitory (basket) cells until gamma rhythms could be observed. The network
+dynamics are driven by Poisson-distributed (noisy) spike trains to layer II/III
+and layer V pyramidal cell populations.
+
+[1] Lee, S. & Jones, S. R. Distinguishing mechanisms of gamma frequency
```
jasmainak, line 14: Maybe a heading "References" above this?

jasmainak: or alternatively add this to the end of the example?

hnn_core/viz.py
cjayb: I've added zero-padding as the default, arriving in the next commit. Looks better than with mirroring for this dataset.

examples/plot_simulate_gamma.py
cjayb: My plotting is insanely slow without some serious decimating, can you test the interactive performance on yours?

cjayb:
Thanks @jasmainak ! Any idea how I can get `plot_spectogram` and `plot_tfr_morlet` to be clickable in the example?

Screenshot 2021-01-21 at 20 02 50

jasmainak:
> Any idea how I can get plot_spectogram and plot_tfr_morlet to be clickable in the example

I think it will be fixed once we merge this PR :-)

jasmainak:
> /Users/mainak/anaconda3/lib/python3.7/site-packages/scipy/signal/spectral.py:1969: UserWarning: nperseg = 8192 is greater than input length  = 8000, using nperseg = 8000
  .format(nperseg, input_length))

I get this warning when running the example. Ideally, examples shouldn't throw warnings so user is not distracted.

jasmainak:
Here's another warning I get:

```py
/Users/mainak/Documents/github_repos/hnn-core/hnn_core/viz.py:378: MatplotlibDeprecationWarning: Adding an axes using the same arguments as a previous axes currently reuses the earlier instance.  In a future version, a new instance will always be created and returned.  Meanwhile, this warning can be suppressed, and the future behavior ensured, by passing a unique label to each axes instance.
  cbar_ax = fig.add_axes([l + w + 0.05, cb_b, 0.03, cb_h])
```

examples/plot_simulate_gamma.py
```diff
@@ -44,44 +54,35 @@
...
-times = dpls[trial_idx].times[mask]
-data = dpls[trial_idx].data['agg'][mask]
+decim = 8  # decimate plots by a factor of 8 (from 40 to 5 kHz)
 
-###############################################################################
-# We can plot the time-frequency response using MNE
-import numpy as np
+# plot dipole time course and time-frequency representation in same figure
 import matplotlib.pyplot as plt
-from mne.time_frequency import tfr_array_morlet
```
jasmainak, line 60: so now the issue is that the user will be surprised with an import warning since the installation instructions didn't talk about installing MNE and neither does the example.

examples/plot_simulate_gamma.py
jasmainak: It's not "insanely slow" but yeah the `decim` does make it faster. I can live with it :)

```py
In [6]: %timeit plot_tfr_morlet(dpls[trial_idx], freqs=freqs, n_cycles=7, tmin=t
   ...: min, decim=8, ax=axes[1])                                               
/Users/mainak/Documents/github_repos/hnn-core/hnn_core/viz.py:378: MatplotlibDeprecationWarning: Adding an axes using the same arguments as a previous axes currently reuses the earlier instance.  In a future version, a new instance will always be created and returned.  Meanwhile, this warning can be suppressed, and the future behavior ensured, by passing a unique label to each axes instance.
  cbar_ax = fig.add_axes([l + w + 0.05, cb_b, 0.03, cb_h])
140 ms ± 5.53 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit plot_tfr_morlet(dpls[trial_idx], freqs=freqs, n_cycles=7, tmin=t
   ...: min, decim=1, ax=axes[1])                                               
/Users/mainak/Documents/github_repos/hnn-core/hnn_core/viz.py:378: MatplotlibDeprecationWarning: Adding an axes using the same arguments as a previous axes currently reuses the earlier instance.  In a future version, a new instance will always be created and returned.  Meanwhile, this warning can be suppressed, and the future behavior ensured, by passing a unique label to each axes instance.
  cbar_ax = fig.add_axes([l + w + 0.05, cb_b, 0.03, cb_h])
940 ms ± 14.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

cjayb:
I'll fix the MPL deprecation warning tomorrow and figure out a solution to the `nperseg` warning too (my version didn't throw that one...). As for `mne`-dependency... I agree we should probably remove deep ones like that, most likely copying relevant parts into `hnn_core`, as you mentioned. But perhaps acceptable for now (0.1)?

jasmainak:
yes, we can do it after 0.1. My comment was mostly to say that let's add a line in the tutorial asking them to install `mne` for the particular example.

jasmainak:
happy to merge once the warnings are fixed!

examples/plot_simulate_gamma.py
rythorpe: Mine ran with a reasonable execution time regardless of decimation as well, ~39 ms with decimation vs. ~295 ms without. For simplicity's sake, I'm not convinced we need an extra line defining the `decim` variable (just like we don't define a separate var for `n_cycles`) but that's stylistic.

rythorpe (approved):

cjayb:
Alright, the last three commits should reduce the warnings and includes mention of `mne`. Please check that this is the case.

Note that the new `plt_show` method copied from `mne` should suppress warnings in all examples!

I'm done here!

cjayb:
> Mine ran with a reasonable execution time regardless of decimation as well, ~39 ms with decimation vs. ~295 ms without. For simplicity's sake, I'm not convinced we need an extra line defining the decim variable (just like we don't define a separate var for n_cycles) but that's stylistic.

Ah, just noticed this. Maybe my macOS renderer is just inefficient. Feel free to remove the `decim`'s from the example before merging!

jasmainak:
Thanks a ton @cjayb for these improvements!

PR #299, opened 2021-03-30
MRG: fix test_compare_hnn_core

blakecaldwell:
A typo here kept MPI backend from running. Default Joblib backend was being compared to itself.


jasmainak (approved):
flake8 issue to fix

otherwise LGTM

Issue #293, opened 2021-03-25
Clean up connectivity API and add tests


ntolley:
This was raised by @rythorpe in #276. The current connectivity API allows for duplicates of the same connection when specified through `net.add_connection()`. Currently connectivity information is stored under `net.connectivity` as a list, so implementing this check would require looping through the entire list every time a new connection as added which seems like an unnecessary overhead. I think this check will require converting `net.connectivity` to easily lookup/index existing connections between cells. 

Another much needed check is ensuring valid cell-cell connections are specified. Currently invalid connection types (inhibitory synapses on L2 proximal/distal dendrites for example) only throw an error when the network is built. Ideally this will be caught in `add_connection()`. This should also come with clear documentation of what constitutes valid cell-cell connections.

ntolley:
Renaming this issue to reflect problems that came up during #318 

Here are the major points that should be addressed in a followup PR

- [ ] More verbose warnings for invalid connections (as described above)
- [ ] Merge the functions used for drives and cell-cell connections. This is partially done as everything is funneled through net.add_connection, but the next step is allowing the drives to target specific gids in a similar API
- [ ] Clean up network class, convert methods related to hard coded network features into functions.

ntolley:
@kenloi this is the issue I was describing earlier about verbose warnings. This would also be a good opportunity to learn about what is considered a "valid" connection in the canonical microcircuit.

ntolley:
I recall from an older discussion that we moved away from making `_Connectivity` it's own class. Also #419 was the major goal for adding more verbose warnings so I think it's save to close this issue!

PR #318, opened 2021-04-19
[MRG] Add functions to modify and visualize connectivity

ntolley:
At long last we're here! This PR is meant to flush out the API for modifying and visualizing the connectivity now that it has been decoupled from `network_builder.py` (#276), and the data structure is lightweight and interpretable (#313). 

Thanks to @jasmainak's suggestion to store separate connection types as objects, there is now a very clean path towards modifying existing connections. To demonstrate, this PR enables modification of existing connections with `net.connectivity[idx].update_probability(0.5)`. I have additionally added a `probability` argument to `net.add_connection`.

Here are my currently planned functions to add:
- `net.connectivity[idx].plot` which will produce a "checkerboard" to visualize connections
- `net.connectivity[idx].update_srcs` and `net.connectivity[idx].update_targets` to filter connections to a subset of srcs and targets. This will be really useful for modifying drives to the network that target different sub populations of cells. 

Let me know if you guys have any other suggestions! 

P.S. I think this mode of modifying connections will be quite amenable to the GUI since it modifies everything in place.

### DIFF
--- a/examples/plot_connectivity.py
+++ b/examples/plot_connectivity.py
@@ -114,6 +114,7 @@ for target in ['L5_pyramidal', 'L2_basket']:
 
 dpl_sparse = simulate_dipole(net_sparse, n_trials=1)
 net_sparse.cell_response.plot_spikes_raster()
+
 net_sparse.connectivity[-2].plot()
 net_sparse.connectivity[-1].plot()
 
--- a/examples/plot_connectivity.py
+++ b/examples/plot_connectivity.py
@@ -45,8 +45,8 @@ print(net_erp.connectivity[0:2])
 
 ###############################################################################
 # Data recorded during simulations are stored under
-# :class:`~hnn_core.Cell_Response`. To test multiple network structures, we can
-# create a copy of the original network. The copied network is then simulated.
+# :class:`~hnn_core.Cell_Response`. Spiking activity can be visualized after
+# a simulation is using :meth:`~hnn_core.Cell_Response.plot_spikes_raster`
 dpl_erp = simulate_dipole(net_erp, n_trials=1)
 net_erp.cell_response.plot_spikes_raster()
 
--- a/examples/plot_connectivity.py
+++ b/examples/plot_connectivity.py
@@ -52,8 +52,9 @@ net_erp.cell_response.plot_spikes_raster()
 # We can modify the connectivity list to test the effect of different
 # connectivity patterns. For example, we can remove 90% all layer 2 inhibitory
 # connections.
-# This can be achieved with :meth:`hnn_core._Connectivity.update_probability`,
-# a function that is available for each element of ``net.connectivity``.
+# This can be achieved with ``net.connectivity[idx].drop``, a function that is
+# available for each element of ``net.connectivity``. We can also visualize
+# any connection using ``net.connectivity.plot``.
 # Note that in the default network, the src_gids of each connection are
 # all the same cell type allowing for easy modifications.
 for conn in net.connectivity:
@@ -63,6 +64,7 @@ for conn in net.connectivity:
 net_remove = net.copy()
 dpl_remove = simulate_dipole(net_remove, n_trials=1)
 net_remove.cell_response.plot_spikes_raster()
+net.connectivity[10].plot()
 
 ###############################################################################
 # That's a lot of spiking! Since basket cells are inhibitory, removing these
@@ -85,6 +87,7 @@ net.add_connection(src_gid, target_gids, location, receptor,
 net_add = net.copy()
 dpl_add = simulate_dipole(net_add, n_trials=1)
 net_add.cell_response.plot_spikes_raster()
+net.connectivity[-1].plot()
 
 ###############################################################################
 # Adding a single inhibitory connection didn't completely restored the normal
--- a/examples/plot_connectivity.py
+++ b/examples/plot_connectivity.py
@@ -8,6 +8,9 @@ This example demonstrates how to modify the network connectivity.
 
 # Author: Nick Tolley <nicholas_tolley@brown.edu>
 
+# sphinx_gallery_thumbnail_number = 4
+
+from hnn_core.network import Network
 import os.path as op
 
 ###############################################################################
@@ -30,64 +33,85 @@ params = read_params(params_fname)
 # explore how it changes with new connections. We first instantiate the
 # network. (Note: Setting ``add_drives_from_params=True`` loads a set of
 # predefined drives without the drives API shown previously).
-net = default_network(params, add_drives_from_params=True)
+net_erp = default_network(params, add_drives_from_params=True)
 
 ###############################################################################
 # Instantiating the network comes with a predefined set of connections that
 # reflect the canonical neocortical microcircuit. ``net.connectivity``
 # is a list of dictionaries which detail every cell-cell, and drive-cell
 # connection.
-print(len(net.connectivity))
-print(net.connectivity[0:2])
+print(len(net_erp.connectivity))
+print(net_erp.connectivity[0:2])
 
 ###############################################################################
 # Data recorded during simulations are stored under
 # :class:`~hnn_core.Cell_Response`. To test multiple network structures, we can
 # create a copy of the original network. The copied network is then simulated.
-net_erp = net.copy()
 dpl_erp = simulate_dipole(net_erp, n_trials=1)
 net_erp.cell_response.plot_spikes_raster()
 
 ###############################################################################
-# We can modify the connectivity list to test the effect of different
-# connectivity patterns. For example, we can remove 90% all layer 2 inhibitory
-# connections.
-# This can be achieved with ``net.connectivity[idx].drop``, a function that is
-# available for each element of ``net.connectivity``. We can also visualize
-# any connection using ``net.connectivity.plot``.
-# Note that in the default network, the src_gids of each connection are
-# all the same cell type allowing for easy modifications.
-for conn in net.connectivity:
-    if conn['src_type'] == 'L2_basket':
-        conn.drop(0.1)
-
-net_remove = net.copy()
-dpl_remove = simulate_dipole(net_remove, n_trials=1)
-net_remove.cell_response.plot_spikes_raster()
-net.connectivity[10].plot()
-
-###############################################################################
-# That's a lot of spiking! Since basket cells are inhibitory, removing these
-# connections increases network wide excitability. We can additionally add
-# new connections using ``net.add_connection()``. Let's try connecting a
-# single layer 2 basket cell, to every layer 2 pyramidal cell. We can utilize
-# ``net.gid_ranges`` to help find the gids of interest.
+# We can also define our own connections to test the effect of different
+# connectivity patterns. To start, an empty network is instantiated.
+# ``add_drives_from_params`` can still be used with an empty Network, but
+# all cells will start disconnected. ``net.add_connection`` is then
+# used to create a custom network. Let us first create an all-to-all
+# connectivity pattern between the L5 pyramidal cells, and L2 basket cells.
 # :meth:`hnn_core.Network.add_connection` allows connections to be specified
 # with either cell names, or the gids directly. If multiple gids are provided
 # for either the sources or the targets, they will be connected in an
 # all-to-all pattern.
-print(net.gid_ranges)
-src_gid = net.gid_ranges['L2_basket'][0]
-target_gids = 'L2_pyramidal'
+
+net_all = Network(params, add_drives_from_params=True)
+
+# Pyramidal cell connections
+location, receptor = 'distal', 'ampa'
+weight, delay, lamtha = 1.0, 1.0, 70
+src = 'L5_pyramidal'
+for target in ['L5_pyramidal', 'L5_pyramidal']:
+    net_all.add_connection(src, target, location, receptor,
+                           delay, weight, lamtha)
+
+# Basket cell connections
 location, receptor = 'soma', 'gabaa'
 weight, delay, lamtha = 1.0, 1.0, 70
-net.add_connection(src_gid, target_gids, location, receptor,
-                   delay, weight, lamtha)
+src = 'L2_basket'
+for target in ['L2_basket', 'L5_pyramidal']:
+    net_all.add_connection(src, target, location, receptor,
+                           delay, weight, lamtha)
 
-net_add = net.copy()
-dpl_add = simulate_dipole(net_add, n_trials=1)
-net_add.cell_response.plot_spikes_raster()
-net.connectivity[-1].plot()
+dpl_all = simulate_dipole(net_all, n_trials=1)
+net_all.cell_response.plot_spikes_raster()
+
+###############################################################################
+# That's a lot of spiking! We can additionally use the ``probability``.
+# argument to create a sparse connectivity pattern instead of all-to-all. Let's
+# try creating the same network with a 10% change of cells connecting
+# to each other. The resulting connectivity pattern can also be visualized
+# with ``net.connectivity[idx].plot()``
+probability = 0.1
+net_sparse = Network(params, add_drives_from_params=True)
+
+# Pyramidal cell connections
+location, receptor = 'distal', 'ampa'
+weight, delay, lamtha = 1.0, 1.0, 70
+src = 'L5_pyramidal'
+for target in ['L5_pyramidal', 'L2_basket']:
+    net_sparse.add_connection(src, target, location, receptor,
+                              delay, weight, lamtha, probability)
+
+# Basket cell connections
+location, receptor = 'soma', 'gabaa'
+weight, delay, lamtha = 1.0, 1.0, 70
+src = 'L2_basket'
+for target in ['L2_basket', 'L5_pyramidal']:
+    net_sparse.add_connection(src, target, location, receptor,
+                              delay, weight, lamtha, probability)
+
+dpl_sparse = simulate_dipole(net_sparse, n_trials=1)
+net_sparse.cell_response.plot_spikes_raster()
+net_sparse.connectivity[-2].plot()
+net_sparse.connectivity[-1].plot()
 
 ###############################################################################
 # Adding a single inhibitory connection didn't completely restored the normal
@@ -98,8 +122,8 @@ import matplotlib.pyplot as plt
 from hnn_core.viz import plot_dipole
 fig, axes = plt.subplots(2, 1, sharex=True, figsize=(6, 6),
                          constrained_layout=True)
-dpls = [dpl_erp[0], dpl_remove[0], dpl_add[0]]
+dpls = [dpl_erp[0], dpl_all[0], dpl_sparse[0]]
 plot_dipole(dpls, ax=axes[0], layer='agg', show=False)
-axes[0].legend(['Normal', 'No L2 Basket', 'Single L2 Basket'])
+axes[0].legend(['Default', 'Custom All', 'Custom Sparse'])
 net_erp.cell_response.plot_spikes_hist(
     ax=axes[1], spike_types=['evprox', 'evdist'])
--- a/examples/plot_connectivity.py
+++ b/examples/plot_connectivity.py
@@ -53,7 +53,7 @@ net_erp.cell_response.plot_spikes_raster()
 # connectivity patterns. For example, we can remove 90% all layer 2 inhibitory
 # connections.
 # This can be achieved with :meth:`hnn_core._Connectivity.update_probability`,
-# a function that is available for each element of `net.connectivity`.
+# a function that is available for each element of ``net.connectivity``.
 # Note that in the default network, the src_gids of each connection are
 # all the same cell type allowing for easy modifications.
 for conn in net.connectivity:
@@ -87,9 +87,10 @@ dpl_add = simulate_dipole(net_add, n_trials=1)
 net_add.cell_response.plot_spikes_raster()
 
 ###############################################################################
-# Adding a single inhibitory connection almost completely restored the normal
-# spiking. As a final step, we can see how this change in spiking
-# activity impacts the aggregate current dipole.
+# Adding a single inhibitory connection didn't completely restored the normal
+# spiking. However, layer 2 firing is interrupted at 70 and 120 ms.
+# As a final step, we can see how this change in spiking activity impacts
+# the aggregate current dipole.
 import matplotlib.pyplot as plt
 from hnn_core.viz import plot_dipole
 fig, axes = plt.subplots(2, 1, sharex=True, figsize=(6, 6),
--- a/examples/plot_connectivity.py
+++ b/examples/plot_connectivity.py
@@ -50,15 +50,15 @@ net_erp.cell_response.plot_spikes_raster()
 
 ###############################################################################
 # We can modify the connectivity list to test the effect of different
-# connectivity patterns. For example, we can remove all layer 2 inhibitory
-# connections. In the default network, the src_gids of each connection are
-# all the same cell type.. Connections are stored under ``conn['gid_pairs']``
-# as a dictionary indexed by src_gid:
-# ``{src_gid1: [target_gid1, target_gid2], ...]``. Each src_gid indexes a
-# list with its target gids
-new_connectivity = [conn for conn in net.connectivity
-                    if conn['src_type'] != 'L2_basket']
-net.connectivity = new_connectivity
+# connectivity patterns. For example, we can remove 90% all layer 2 inhibitory
+# connections.
+# This can be achieved with :meth:`hnn_core._Connectivity.update_probability`,
+# a function that is available for each element of `net.connectivity`.
+# Note that in the default network, the src_gids of each connection are
+# all the same cell type allowing for easy modifications.
+for conn in net.connectivity:
+    if conn['src_type'] == 'L2_basket':
+        conn.update_probability(0.1)
 
 net_remove = net.copy()
 dpl_remove = simulate_dipole(net_remove, n_trials=1)
@@ -87,9 +87,8 @@ dpl_add = simulate_dipole(net_add, n_trials=1)
 net_add.cell_response.plot_spikes_raster()
 
 ###############################################################################
-# Adding more inhibitory connections did not completely restore the normal
-# spiking. L2 basket and pyramidal cells rhythymically fire in the gamma
-# range (30-80 Hz). As a final step, we can see how this change in spiking
+# Adding a single inhibitory connection almost completely restored the normal
+# spiking. As a final step, we can see how this change in spiking
 # activity impacts the aggregate current dipole.
 import matplotlib.pyplot as plt
 from hnn_core.viz import plot_dipole
--- a/examples/plot_connectivity.py
+++ b/examples/plot_connectivity.py
@@ -10,7 +10,6 @@ This example demonstrates how to modify the network connectivity.
 
 # sphinx_gallery_thumbnail_number = 5
 
-from hnn_core.network import Network
 import os.path as op
 
 ###############################################################################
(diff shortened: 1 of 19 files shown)


hnn_core/network.py
```diff
@@ -1170,6 +1185,58 @@ def __repr__(self):
 
         return entr
 
+    def update_probability(self, probability):
```
jasmainak, line 1188:
```suggestion
    def drop(self, probability):
```

hnn_core/network.py
```diff
@@ -1170,6 +1185,58 @@ def __repr__(self):
...
+        These variables are meant to describe the set of original connections
+        before they are randomly removed.
+
+        The probability attribute will store the most recent value passed to
+        this function. As such, this number does not accurately describe the
+        connections probability of the original set after successive calls.
+        """
+        _validate_type(probability, float, 'probability')
+        if probability <= 0.0 or probability >= 1.0:
+            raise ValueError('probability must be in the range [0,1]')
```
jasmainak, line 1209:
```suggestion
            raise ValueError('probability must be in the range (0,1)')
```

since 0 and 1 are not included

examples/plot_connectivity.py
```diff
@@ -50,15 +50,15 @@
...
-net.connectivity = new_connectivity
+# connectivity patterns. For example, we can remove 90% all layer 2 inhibitory
+# connections.
+# This can be achieved with :meth:`hnn_core._Connectivity.update_probability`,
+# a function that is available for each element of ``net.connectivity``.
+# Note that in the default network, the src_gids of each connection are
+# all the same cell type allowing for easy modifications.
+for conn in net.connectivity:
+    if conn['src_type'] == 'L2_basket':
+        conn.update_probability(0.1)
```
jasmainak, line 61: Can we use the `net.add_connection(..., probability)` method so we can still play around with `_Connectivity` class. I suspect we'll need to make changes when we unify it with the drive connectivity

jasmainak:
I like that connectivity is becoming a proper object slowly! But it's not yet clear to me if the object we expose to the user stores the "list of connectivities" or just a single connectivity instance? It seems to me it might be more useful to do:

```py
net.connectivity.plot()
net.connectivity.filter(drop_probability, src_types, target_types)
```

etc. In order to not let this decision slow down the development, I suggest starting from functions `plot_connectivity(...)`, `filter_connectivity` etc. and then building methods (in a later PR) when it becomes clear some functions belong together in a class.

jasmainak:
@ntolley are you familiar with the Iris dataset plot? 

image

I am imagining something like this might be useful to show connections between different cell classes? Instead of scatterplot, you'll have matshow/imshow

ntolley:
Great suggestion! I was actually just about to gripe that there is too much to plot for the default network, but filtering to a subset of connections might work here. This sort of plot would require fixing the `loc` and `receptor` which would make some plots a bit uninformative (invalid loc receptor combinations would be a big blob). 

In general though I would like to emphasize the list of connections (or dict of connections that allows efficient indexing) rather than a generic `net.connectivity` class. We can still add auxillary functions to `Network` that call the methods of `_Connection`.

My rationale is that the "canonical" neocortical circuit is characterized by a unique set of connectivity rules. I think the API should reflect this conserved circuit pattern, and allow users to interact with it. For me the motivation is to probe what makes these connection rules so special, not so much looking at how a singular cell class impacts cortical function.

jasmainak:
okay I see, makes sense. 

How have neuro papers made plots like this? Do you have any examples? Might be good to conform to the expectations of most users ...

ntolley:
Interestingly neither the [Blue Brain Project](https://www.sciencedirect.com/science/article/pii/S0092867415011915) or the [Brain Modeling Toolkit](https://www.cell.com/neuron/pdf/S0896-6273(20)30067-2.pdf) plot connectivity matrices as we've discussed. Instead they have matrices of connection probabilities, where the rows/columns are the different cell classes. While these are some the most closely related projects, their emphasis on 1:1 morphologic reconstructions makes the visualizations not very relevant.

Papers constructing functional networks with single unit recordings will likely be the best place mimic visualizations. Mainly because the method forces grouping cells into broad categories (and there's only 4 in HNN). This [Scherberger Paper](https://elifesciences.org/articles/15719#fig5) does a nice plot showing how single units from different areas are connected: 
image

Also [Olaf Sporns](https://pubmed.ncbi.nlm.nih.gov/30250388/) has done a lot of work applying graph theory to neural networks, and his papers may have more visualization examples. However, his focus is on inter-areal connectivitiy rather than single neurons.

jasmainak:
I like the subplots A and B in the figure you share above. Agree that we don't want to go the Blue Brain way ... 

But are we going to colorcode and show lamtha? I still have difficulty visualizing how the weights are chosen and how these weights drive the currents. Agree that some domain knowledge is missing on my part. I am thinking that there needs to be a diagram like one of these at some point:

image
image

where the thickness of the line indicates the weight/probability and color indicates if it's inhibitory/excitatory? And the populations are arranged spatially as they would be in a cortical column. That way you can literally take the diagram and explain to someone what the mechanism is ...

examples/plot_connectivity.py
```diff
@@ -50,19 +50,21 @@
...
+# Note that in the default network, the src_gids of each connection are
+# all the same cell type allowing for easy modifications.
+for conn in net.connectivity:
+    if conn['src_type'] == 'L2_basket':
+        conn.drop(0.1)
 
 net_remove = net.copy()
 dpl_remove = simulate_dipole(net_remove, n_trials=1)
 net_remove.cell_response.plot_spikes_raster()
+net.connectivity[10].plot()
```
jasmainak, line 67: can you change the sphinx gallery thumbnail to point to this plot?

examples/plot_connectivity.py
```diff
@@ -50,19 +50,21 @@
...
+# connectivity patterns. For example, we can remove 90% all layer 2 inhibitory
+# connections.
+# This can be achieved with ``net.connectivity[idx].drop``, a function that is
+# available for each element of ``net.connectivity``. We can also visualize
+# any connection using ``net.connectivity.plot``.
+# Note that in the default network, the src_gids of each connection are
+# all the same cell type allowing for easy modifications.
+for conn in net.connectivity:
+    if conn['src_type'] == 'L2_basket':
+        conn.drop(0.1)
```
jasmainak, line 62: `connectivity` is a private class. You shouldn't expose it (yet) ... could you perhaps expose `probability` using `add_connection` ? This feels like you're trying to modify the connection but we're still trying to settle on API for adding a connection

hnn_core/network.py
```diff
@@ -1170,6 +1196,74 @@ def __repr__(self):
...
+
+        Notes
+        -----
+        num_srcs and num_targets are not updated after pruning connections.
+        These variables are meant to describe the set of original connections
+        before they are randomly removed.
+
+        The probability attribute will store the most recent value passed to
+        this function. As such, this number does not accurately describe the
+        connections probability of the original set after successive calls.
```
jasmainak, line 1216: keep the complexity bounded, otherwise software will become unmaintainable. You should try to address 80 percent usecases. Look up [feature creep](https://en.wikipedia.org/wiki/Feature_creep). I'd suggest throwing an error if someone tries to drop connections more than once

hnn_core/network.py
```diff
@@ -1170,6 +1196,74 @@ def __repr__(self):
 
         return entr
 
+    def drop(self, probability):
```
jasmainak, line 1199: honestly I would keep this for future PR. I think you can add `probability` to `add_connection` with 1 or 2 lines and expose it in an example. This falls under the usecase of modifying existing connections which we haven't started looking into yet.

hnn_core/network.py
```diff
@@ -1170,6 +1196,74 @@ def __repr__(self):
...
+            self['gid_pairs'].pop(src_gid)
+
+        self['probability'] = probability
+
+    def plot(self, ax=None, show=True):
+        """Plot connectivity matrix for instance of _Connectivity object.
+
+        Parameters
+        ----------
+        conn : Instance of _Connectivity object
```
cjayb, line 1256: This is now a class method, document `ax` and `show` instead

hnn_core/network.py
cjayb: But we'll probably want to modify the "default" connections probabilistically, too?

cjayb:
How about adding an argument to `Network` that makes the call to `_set_default_connections` conditional. Then we could have a demo of creating the default connectivity using (many) consecutive calls to `add_ connection`, with and without the probabilistic angle. It would make the default connectivity explicit (not that users would want to do that), and allow exploring less rigid patterns with probability. The `drop`-method could indeed wait, though I do see why you've included it here, @ntolley

jasmainak:
related idea is to move `_set_default_connectivity` out of the `Network` class and into a separate function. I would do something like:

```py
from hnn_core import default_network

network = default_network(params)
```

That way the `Network` class has less baggage of the past. We should try to make the init thin. It is doing too much right now

jasmainak:
@ntolley you need a rebase here

ntolley:
Can do! I'm a bit underwater with research projects until mid next week but I can focus on wrapping up this PR afterwards.

ntolley:
Default network is now created with `net = default_network(params)` as suggested by @jasmainak.

I think all that is left to do is to fix the connectivity example. Since we are not supporting modifications of existing connectivity, I think this will have to be turned into the example suggested by @cjayb where we create a network (not necessarily the default!) with successive calls to `net.add_connection` in the notebook. 

In that case it will likely be useful to move away from instantiating the entire default network, but just add a few sparse connections and show off the plotting functions, as well as how they behave in simulation.

jasmainak:
Can you make the CIs happy so I can see how the example looks like?

Also @ntolley would you have some time to brainstorm the connection API over zoom sometime? I would like to make sure that we're not overseeing any critical usecases

jasmainak:
Just wanted to mention that as such the PR looks good. But overall, we need to think a bit about the connectivity API and what needs to be done to make sure users have the best experience

ntolley:
Just finished updating the example to be more appropriate for the current API. I've opted to create hand-made networks and compare their activity patterns. 

Also I have a super free schedule for zoom Wed-Fri next week!

ntolley:
In terms of critical use cases, two things come to mind. The first is implementing the src-target definition conventions in the drives API, allowing for drives to only a subset of cells. In my own research/reading. In my own research/reading this abililty seems fairly important for understanding circuit-EEG activity (how does the dipole look when subpopulations are driven from indepedent sources?). 

The second thing we should consider is adding connection probabilities that decay with distance. While a fixed connection probability is fine for now, distance dependent probability is much more biologically realistic topology. My understanding is that the motivation for Gaussian decaying synaptic strengths was to generally how net synaptic inputs from one location fall off with distance. In reality though the strength of a synaptic connection isn't really influenced by distance (as far as I know...).

jasmainak:
Cool, could you send me a zoom/calendar invite on one of those days? I'm pretty free. Everything works except Wednesday 9 am

examples/plot_simulate_gamma.py
```diff
@@ -41,7 +41,7 @@
 # simulate the dipole moment in a single trial (the default value used by
 # ``simulate_dipole`` is ``n_trials=params['N_trials']``).
 
-net = Network(params)
+net = default_network(params)
```
jasmainak, line 44: add to API documentation?

hnn_core/network.py
```diff
@@ -90,6 +91,147 @@ def _create_cell_coords(n_pyr_x, n_pyr_y, zdiff=1307.4):
     return pos_dict
 
 
+# connections:
+# this NODE is aware of its cells as targets
+# for each syn, return list of source GIDs.
+# for each item in the list, do a:
+# nc = pc.gid_connect(source_gid, target_syn), weight,delay
+# Both for synapses AND for external inputs
```
jasmainak, line 99: is this where the comment was or should be? Seems a bit out of place

examples/plot_connectivity.py
```diff
@@ -30,73 +33,99 @@
...
 # is a list of dictionaries which detail every cell-cell, and drive-cell
 # connection.
-print(len(net.connectivity))
-print(net.connectivity[0:2])
+print(len(net_erp.connectivity))
+print(net_erp.connectivity[0:2])
 
 ###############################################################################
 # Data recorded during simulations are stored under
 # :class:`~hnn_core.Cell_Response`. To test multiple network structures, we can
```
jasmainak, line 48: Need to change this line?

examples/plot_connectivity.py
```diff
@@ -30,73 +33,99 @@
...
-src_gid = net.gid_ranges['L2_basket'][0]
-target_gids = 'L2_pyramidal'
+
+net_all = Network(params, add_drives_from_params=True)
+
+# Pyramidal cell connections
+location, receptor = 'distal', 'ampa'
+weight, delay, lamtha = 1.0, 1.0, 70
+src = 'L5_pyramidal'
+for target in ['L5_pyramidal', 'L5_pyramidal']:
```
jasmainak, line 71: pyramidal twice? why so?

examples/plot_connectivity.py
```diff
@@ -30,73 +33,99 @@
...
+    net_all.add_connection(src, target, location, receptor,
+                           delay, weight, lamtha)
+
+dpl_all = simulate_dipole(net_all, n_trials=1)
+net_all.cell_response.plot_spikes_raster()
+
+###############################################################################
+# That's a lot of spiking! We can additionally use the ``probability``.
+# argument to create a sparse connectivity pattern instead of all-to-all. Let's
+# try creating the same network with a 10% change of cells connecting
```
jasmainak, line 89: 10% chance?

examples/plot_connectivity.py
```diff
@@ -30,73 +33,99 @@
...
+    net_sparse.add_connection(src, target, location, receptor,
+                              delay, weight, lamtha, probability)
 
-net_add = net.copy()
-dpl_add = simulate_dipole(net_add, n_trials=1)
-net_add.cell_response.plot_spikes_raster()
+dpl_sparse = simulate_dipole(net_sparse, n_trials=1)
+net_sparse.cell_response.plot_spikes_raster()
+net_sparse.connectivity[-2].plot()
+net_sparse.connectivity[-1].plot()
```
jasmainak, line 114: From this plot it's not clear that L2Basket are the rows. Could we instead say "L2_basket gids (0-34)" on x axis label?

examples/plot_connectivity.py
```diff
@@ -30,73 +33,99 @@
...
 
 ###############################################################################
-# Adding more inhibitory connections did not completely restore the normal
-# spiking. L2 basket and pyramidal cells rhythymically fire in the gamma
-# range (30-80 Hz). As a final step, we can see how this change in spiking
-# activity impacts the aggregate current dipole.
+# Using the sparse connectivity pattern led produced a lot more spiking in
```
jasmainak, line 117: "led produced" -> "produced"

examples/plot_connectivity.py
```diff
@@ -30,73 +33,99 @@
...
+# the L5 pyramidal cells. While there are less excitatory connections overall,
+# there was also a decrease in the inhibtory connections. This shift in
+# activity is often referred to as the excitatory/inhibitory (E/I) balance.
+# As a final step, we can see how this change in spiking activity impacts
+# the aggregate current dipole.
```
jasmainak, line 122: okay I am confused which way the E/I balance shifted?

Also, it might be more intuitive for the reader to find this hypothesis first before you try changing the connectivity, then what you tried, and finally your observation confirming the hypothesis

jasmainak:
> The first is implementing the src-target definition conventions in the drives API, allowing for drives to only a subset of cells.

yeah, this would be great to have!

examples/plot_connectivity.py
```diff
@@ -8,13 +8,16 @@
 
 # Author: Nick Tolley <nicholas_tolley@brown.edu>
 
+# sphinx_gallery_thumbnail_number = 4
```
jasmainak, line 11: should we move this example before the MPI example? I think that makes more sense logically

hnn_core/network.py
ntolley: Good point, this got dragged along when I was moving everything from `network_builder`, but it definitely doesn't make sense here any more. I'll put it back in an appropriate spot.

examples/plot_connectivity.py
ntolley: Changing "L2_basket source gids (0-34)"? I still think it's important to indicate directionality since the connectivity might look different when srcs and targets are flipped.

examples/plot_connectivity.py
ntolley: Definitely a mistake, fixing now. The goal is to just create connections within and between the different cells classes.

examples/plot_connectivity.py
ntolley: I was trying to draw attention to how E/I balance is not just a product of the raw number of connections, but also how the circuit is wired. Honestly this is a bit too much for a simple example so I am just adding descriptions of the resulting spiking activity.

ntolley:
@jasmainak I think I've addressed all of your comments. I'm pretty happy with this PR so merge if you dare 😉 

@cjayb @rythorpe definitely feel free to provide any suggestions. I've rewritten the `plot_connectivity` example to show off the new interface for creating and visualizing probabalistic cell-cell connections.

jasmainak:
:) Between this PR and #321 which one should be merged first? Think this PR is easier to rebase? I don't know ...

ntolley:
Go ahead and merge the other! I've got time tonight to submit a rebase.

hnn_core/tests/test_network.py
```diff
@@ -226,6 +228,24 @@ def test_network():
...
+    kwargs = kwargs_default.copy()
+    kwargs['probability'] = 0.5
+    net.add_connection(**kwargs)
+    n_connections = np.sum(
+        [len(t_gids) for
+         t_gids in net.connectivity[-2]['gid_pairs'].values()])
+    n_connections_new = np.sum(
+        [len(t_gids) for
+         t_gids in net.connectivity[-1]['gid_pairs'].values()])
+    assert n_connections_new == np.round(n_connections * 0.5).astype(int)
```
cjayb, line 242: Is this safe? Will it always be `==`, or could there be a rounding error (+/- 1)?

cjayb (approved):
Looks great! I have a few minor comments/questions, sorry about being last-minute.

examples/plot_connectivity.py
```diff
@@ -30,73 +33,101 @@
...
+# connectivity pattern between the L5 pyramidal cells, and L2 basket cells.
 # :meth:`hnn_core.Network.add_connection` allows connections to be specified
 # with either cell names, or the gids directly. If multiple gids are provided
 # for either the sources or the targets, they will be connected in an
 # all-to-all pattern.
-print(net.gid_ranges)
-src_gid = net.gid_ranges['L2_basket'][0]
-target_gids = 'L2_pyramidal'
+
+net_all = Network(params, add_drives_from_params=True)
```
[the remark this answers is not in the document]
ntolley: `net_all` refers to the default all-to-all connectivity pattern which is later compared with a `net_sparse` pattern.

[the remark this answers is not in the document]
ntolley: I see why the terminology is a bit confusing, but I am not sure how to get around it without reintroducing the `add_connection` steps back to the `__init__`. 

I think one way this could be improved is a more robust way to remove connections. Something like 
```py
net = default_network(params)
net.clear_connectivity(remove_drives=False)
```
This way the user never really has to deal with the `Network` constructor.

hnn_core/tests/test_network.py
ntolley: Good question. Under the hood, I am gathering all connections into a list. I then calculate how many connections should be retained with `n_connections * probability`, and then use `np.random.choice(conn_list, size=n_conn, replace=False)`. This guarantees that there will always be a predictable number of connections. 

Perhaps I can be more clear in the docs, as some people might expect a probabilistic yes/no for every possible connection. In that case there may be slight variation.

cjayb: > This guarantees that there will always be a predictable number of connections.

nice!

ntolley:
> they'll get notified every time you rebase and push that commit
😨 sorry chris! We'll try and get this merged soon haha

@jasmainak I am adding methods to clear connectivity and drives separately to avoid using `net = Network()` in any of the examples. Then I will call it quits until a followup PR.

examples/plot_connectivity.py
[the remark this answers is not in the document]
cjayb: Or how about

```python
net = default_network(params, connect=True, attach_drives=True)
```

I don't know if it should be a priority  to avoid `Network`? There's just something odd about the example right now in the way that `default_network` and `Network` are used.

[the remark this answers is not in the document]
ntolley: @jasmainak did you have a vision in mind for how `default_network()` would be used? Going down this path, it seems like we're just recreating the original `Network.__init__` in some other location.

In my mind instantiating an empty network should require some intention by the user as very few people have the background to create a realistic network _de novo_.

jasmainak:
@ntolley are you aware of this paper: https://www.biorxiv.org/content/biorxiv/early/2021/04/18/2021.04.16.440210.full.pdf ?

@stephanie-r-jones mentioned to me that the connectivity is changed in there. Could that be a good example for us?

ntolley:
@jasmainak regarding our previous discussion I was wrong. We can totally add that new connection with the current API. Here's the relevant line from the supplemental:
```
There were also some network structure differences between the 2009 model and current model. In the current model, we added
a  Martinotti-like recurrent tuft connections from the L5 interneuron to the L5 pyramidal neurons’ distal dendrites and removed
the L2/3 GABAA interneuron synapses onto those same dendrites.
```

I just checked and list is easily done in the model.

jasmainak:
Nice! Is there something from the paper that we can easily demo/illustrate due to the new connectivity? I'm fine leaving this for another PR though if you think it's outside the scope.

ntolley:
Let's leave it for another PR, I'll make it my next one to focus on. I agree that it will be great to show off as many examples as possible with published (almost) results. It's possible the influence of beta events on ERP's is partially present with just this new connection, but I'll need to play with the code some more. 

There were a lot of modified parameters in this paper so it may require a separate params file. Of course this can be more easily done with #321 merged!

hnn_core/network.py
```diff
@@ -143,6 +279,13 @@ class Network(object):
         Firing threshold of all cells.
     delay : float
         Synaptic delay in ms.
+
+    Notes
+    ----
+    `net = default_network(params)` is the reccomended path for creating a
+    network. Instantiating the network as `net = Network(params)` will
```
jasmainak, line 286: we should migrate the stuff that is specific to `params` specification to `default_network` slowly leaving a generalized `Network` class in place. That is to say, it should not take a `params` dictionary as input and should not have `add_drives_from_params`. Happy to delegate this to another PR though

examples/plot_connectivity.py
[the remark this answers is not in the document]
jasmainak: see my other comment. I think `Network` should be generalized a bit and `default_network` is where the heavy lifting of the `params` extraction should be done. I want the network class to go the same way as we did with the `Cell` class. I can't make up my mind whether we should have different `param` file specifications for every new network or if we can leverage existing specifications. I think the latter. However, once loaded into HNN-core, it should be represented with our `Network` class which is fairly generalizable.

hnn_core/network.py
```diff
@@ -201,8 +344,6 @@ def __init__(self, params, add_drives_from_params=False,
         self.n_cells = sum(len(self.pos_dict[src]) for src in
                            self.cell_types)
 
-        self._set_default_connections()
-
```
jasmainak, line 205: in a future PR, `add_drives_from_params` should be moved to `default_network` function

hnn_core/network.py
```diff
@@ -1070,11 +1101,87 @@ def add_connection(self, src_gids, target_gids, loc, receptor,
...
 
+        # Probabilistically define connections
+        if probability != 1.0:
+            self._connection_probability(conn, probability, seed)
+
+        conn['probability'] = probability
+
         self.connectivity.append(deepcopy(conn))
 
+    def _connection_probability(self, conn, probability, seed=0):
```
jasmainak, line 1112: prefer function over method where you can

hnn_core/network.py
```diff
@@ -1070,11 +1101,87 @@ def add_connection(self, src_gids, target_gids, loc, receptor,
...
+        ----------
+        conn : Instance of _Connectivity object
+            Object specifying the biophysical parameters and src target pairs
+            of a specific connection class. Function modifies conn in place.
+        probability : float
+            Probability of connection between any src-target pair.
+            Defaults to 1.0 producing an all-to-all pattern.
+        seed : int
+            Seed for the numpy random number generator.
+        Notes
```
jasmainak, line 1125:
```suggestion
            Seed for the numpy random number generator.

        Notes
```

hnn_core/network.py
```diff
@@ -1128,10 +1235,18 @@ class _Connectivity(dict):
...
+        dict indexed by src gids with the format:
+        {src_gid: [target_gids, ...], ...}
+        where each src_gid indexes a list of all its targets.
     num_srcs : int
         Number of unique source gids.
     num_targets : int
         Number of unique target gids.
+    src_range : range
+        Range of gids identified by src_type.
+    target_range : target_range
```
jasmainak, line 1248:
```suggestion
    target_range : range
```

the datatype should be after the column

hnn_core/network.py
```diff
@@ -1149,12 +1264,21 @@ class _Connectivity(dict):
             Synaptic delay in ms.
         lamtha : float
             Space constant.
+    probability : float
+        Probability of connection between any src-target pair.
+        Defaults to 1.0 producing an all-to-all pattern.
+
+    Notes
+    -----
+    The len() of src_range or target_range may not match
```
jasmainak, line 1273:
```suggestion
    The len() of src_range or target_range will not match
```

examples/plot_connectivity.py
```diff
@@ -30,73 +32,103 @@
...
+src = 'L2_basket'
+for target in ['L5_pyramidal', 'L2_basket']:
+    net_sparse.add_connection(src, target, location, receptor,
+                              delay, weight, lamtha, probability)
 
-net_add = net.copy()
-dpl_add = simulate_dipole(net_add, n_trials=1)
-net_add.cell_response.plot_spikes_raster()
+dpl_sparse = simulate_dipole(net_sparse, n_trials=1)
+net_sparse.cell_response.plot_spikes_raster()
```
jasmainak, line 116: could you add a tiny text line after this so that the connectivity plots are forced on the next line?

jasmainak (approved):
looks good other than nitpicks. Can you address them and I'll merge first thing tomorrow morning.

Is there anything to update in `whats_new` @ntolley ?

hnn_core/network.py
cjayb: I agree this is a great direction to go. Let's iterate on generalising `Network` and removing  `add_drives_from_params` together with the `legacy`-flag.

hnn_core/network.py
ntolley: Added this to #293. I'm planning to rewrite this function to work with the drives API and also allow for distant dependent connection probabilities

ntolley: nvm, just moved it to a function as the previous comment doesn't really impact that decision

ntolley:
I believe I've addressed the remaining comments! Let me know if anything else catches your eye, otherwise feel free to merge.

examples/plot_connectivity.py
jasmainak: I actually meant a line of text. But that's fine, we can fix it later

jasmainak:
Thanks @ntolley !

PR #321, opened 2021-04-26
[MRG] add method for updating cell biophysics from Network instance

### LINKED ISSUE #316: no way to modify cell properties
There is currently no mechanism for even an advanced user to set cell properties (e.g., GABAA decay time as demonstrated in the [gamma tutorial](https://jonescompneurolab.github.io/hnn-tutorials/gamma/gamma)). All cell properties are set to the default values within each cell instance when `network_builder._build()` calls `_create_cells_and_drives()`.

A short-term solution would be to set `override_params` when `_create_cells_and_drives()` instantiates its cell objects. Advanced users (or hnn-gui) can change cell properties in the `Params` dict.

rythorpe:
closes #316 

The goal here is to create an API for users to adjust biphysics and synapse properties as they do in HNN-GUI.

### DIFF
--- a/examples/plot_simulate_gamma.py
+++ b/examples/plot_simulate_gamma.py
@@ -82,7 +82,7 @@ freqs = np.arange(20., 100., 1.)
 dpls[trial_idx].plot_tfr_morlet(freqs, n_cycles=7, tmin=tmin, ax=axes[1])
 
 ###############################################################################
-# Now, let us try to re-run the simulation with a tonic bias
+# As a final exercise, let us try to re-run the simulation with a tonic bias
 # applied to the L5 Pyramidal cells. Notice that the oscillation waveform is
 # more regular, with less noise due to the fact that the tonic depolarization
 # dominates over the influence of the Poisson drive. By default, a tonic bias
@@ -107,21 +107,6 @@ net.cell_response.plot_spikes_raster()
 # frequency is lower compared with the "weak" PING simulation above.
 from hnn_core.viz import plot_psd
 plot_psd(dpls[trial_idx], fmin=20., fmax=100., tmin=tmin)
-
-###############################################################################
-# Finally, we demonstrate the mechanistic link between PING and the GABAA decay
-# time constant (`tau2`). Using the same network/drive configuration as before,
-# we decrease `tau2` from 5 to 2 ms. This will shorten the effective
-# refactory period between L5 pyramidal cell spikes and increase the PING
-# frequency from ~50 to ~65 Hz.
-net.cell_properties['L5_pyramidal']['synapses']['gabaa']['tau2'] = 2
-dpls = simulate_dipole(net, n_trials=1)
-
-fig, axes = plt.subplots(3, 1, sharex=True, figsize=(6, 6),
-                         constrained_layout=True)
-dpls[trial_idx].plot(ax=axes[0], show=False)
-net.cell_response.plot_spikes_raster(ax=axes[1], show=False)
-dpls[trial_idx].plot_tfr_morlet(freqs, n_cycles=7, tmin=tmin, ax=axes[2])
 
 ###############################################################################
 # References
--- a/examples/plot_simulate_gamma.py
+++ b/examples/plot_simulate_gamma.py
@@ -114,7 +114,8 @@ plot_psd(dpls[trial_idx], fmin=20., fmax=100., tmin=tmin)
 # we decrease `tau2` from 5 to 2 ms. This will shorten the effective
 # refactory period between L5 pyramidal cell spikes and increase the PING
 # frequency from ~50 to ~65 Hz.
-net.cell_properties['L5_pyramidal']['synapses']['gabaa']['tau2'] = 2
+net.cell_types['L5_pyramidal'].p_syn['gabaa']['tau2'] = 2
+net.update_cells()
 dpls = simulate_dipole(net, n_trials=1)
 
 fig, axes = plt.subplots(3, 1, sharex=True, figsize=(6, 6),
--- a/examples/plot_simulate_gamma.py
+++ b/examples/plot_simulate_gamma.py
@@ -115,7 +115,6 @@ plot_psd(dpls[trial_idx], fmin=20., fmax=100., tmin=tmin)
 # refactory period between L5 pyramidal cell spikes and increase the PING
 # frequency from ~50 to ~65 Hz.
 net.cell_types['L5_pyramidal'].p_syn['gabaa']['tau2'] = 2
-net.update_cells()
 dpls = simulate_dipole(net, n_trials=1)
 
 fig, axes = plt.subplots(3, 1, sharex=True, figsize=(6, 6),
--- a/examples/plot_simulate_gamma.py
+++ b/examples/plot_simulate_gamma.py
@@ -82,7 +82,7 @@ freqs = np.arange(20., 100., 1.)
 dpls[trial_idx].plot_tfr_morlet(freqs, n_cycles=7, tmin=tmin, ax=axes[1])
 
 ###############################################################################
-# As a final exercise, let us try to re-run the simulation with a tonic bias
+# Now, let us try to re-run the simulation with a tonic bias
 # applied to the L5 Pyramidal cells. Notice that the oscillation waveform is
 # more regular, with less noise due to the fact that the tonic depolarization
 # dominates over the influence of the Poisson drive. By default, a tonic bias
@@ -107,6 +107,21 @@ net.cell_response.plot_spikes_raster()
 # frequency is lower compared with the "weak" PING simulation above.
 from hnn_core.viz import plot_psd
 plot_psd(dpls[trial_idx], fmin=20., fmax=100., tmin=tmin)
+
+###############################################################################
+# Finally, we demonstrate the mechanistic link between PING and the GABAA decay
+# time constant (`tau2`). Using the same network/drive configuration as before,
+# we decrease `tau2` from 5 to 2 ms. This will shorten the effective
+# refactory period between L5 pyramidal cell spikes and increase the PING
+# frequency from ~50 to ~65 Hz.
+net.cell_properties['L5_pyramidal']['synapses']['gabaa']['tau2'] = 2
+dpls = simulate_dipole(net, n_trials=1)
+
+fig, axes = plt.subplots(3, 1, sharex=True, figsize=(6, 6),
+                         constrained_layout=True)
+dpls[trial_idx].plot(ax=axes[0], show=False)
+net.cell_response.plot_spikes_raster(ax=axes[1], show=False)
+dpls[trial_idx].plot_tfr_morlet(freqs, n_cycles=7, tmin=tmin, ax=axes[2])
 
 ###############################################################################
 # References
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -160,6 +160,7 @@ class Cell:
     def _set_biophysics(self, p_secs):
         "Set the biophysics for the cell."
 
+        from .cells_default import _set_variable_mech
         # neuron syntax is used to set values for mechanisms
         # sec.gbar_mech = x sets value of gbar for mech to x for all segs
         # in a section. This method is significantly faster than using
@@ -177,10 +178,11 @@ class Cell:
             for mech_name, p_mech in p_sec['mechs'].items():
                 sec.insert(mech_name)
                 for attr, val in p_mech.items():
-                    if hasattr(val, '__call__'):
+                    if val == 'variable':
                         sec.push()
                         for seg in sec:
-                            setattr(seg, attr, val(h.distance(seg.x)))
+                            val = _set_variable_mech(h.distance(seg.x))
+                            setattr(seg, attr, val)
                         h.pop_section()
                     else:
                         setattr(sec, attr, val)
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -277,12 +277,12 @@ class Cell:
             child_loc = connection[3]
             child_sec.connect(parent_sec, parent_loc, child_loc)
 
-    def build(self, align_dipole=None):
+    def build(self, sec_name_apical=None):
         """Build cell in Neuron and insert dipole if applicable.
 
         Parameters
         ----------
-        align_dipole : str | None
+        sec_name_apical : str | None
             If not None, a dipole will be inserted in this cell in alignment
             with this section. The section should belong to the apical dendrite
             of a pyramidal neuron.
@@ -292,11 +292,12 @@ class Cell:
         self._create_sections(self.p_secs, self.topology)
         self._create_synapses(self.p_secs, self.p_syn)
         self._set_biophysics(self.p_secs)
-        if align_dipole in self.sections:
-            self.insert_dipole(align_dipole)
-        elif align_dipole is not None:
-            raise ValueError(f'align_dipole must be an exiting section of'
-                             f'the current cell or None. Got {align_dipole}.')
+        if sec_name_apical in self.sections:
+            self._insert_dipole(sec_name_apical)
+        elif sec_name_apical is not None:
+            raise ValueError(f'sec_name_apical must be an existing '
+                             f'section of the current cell or None. '
+                             f'Got {sec_name_apical}.')
 
     def move_to_pos(self):
         """Move cell to position."""
@@ -317,7 +318,7 @@ class Cell:
     # 2. a list needs to be created with a Dipole (Point Process) in each
     #    section at position 1
     # In Cell() and not Pyr() for future possibilities
-    def insert_dipole(self, sec_name_apical):
+    def _insert_dipole(self, sec_name_apical):
         """Insert dipole into each section of this cell.
 
         Parameters
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -91,6 +91,29 @@ class Cell:
         The name of the cell.
     pos : tuple
         The (x, y, z) coordinates.
+    p_secs : nested dict
+        Dictionary with keys as section name.
+        p_secs[sec_name] is a dictionary with keys
+        L, diam, Ra, cm, syns and mech.
+        syns is a list specifying the synapses at that section.
+        The properties of syn are specified in p_syn.
+        mech is a dict with keys as the mechanism names. The
+        values are dictionaries with properties of the mechanism.
+    p_syn : dict of dict
+        Keys are name of synaptic mechanism. Each synaptic mechanism
+        has keys for parameters of the mechanism, e.g., 'e', 'tau1',
+        'tau2'.
+    topology : list of list
+        The topology of cell sections. Each element is a list of
+        4 items in the format
+        [parent_sec, parent_loc, child_sec, child_loc] where
+        parent_sec and parent_loc are float between 0 and 1
+        specifying the location in the section to connect and
+        parent_sec and child_sec are names of the connecting
+        sections.
+    sect_loc : dict of list
+        Can have keys 'proximal' or 'distal' each containing
+        names of section locations that are proximal or distal.
     gid : int or None (optional)
         Each cell in a network is uniquely identified by it's "global ID": GID.
         The GID is an integer from 0 to n_cells, or None if the cell is not
@@ -122,15 +145,37 @@ class Cell:
     sect_loc : dict of list
         Can have keys 'proximal' or 'distal' each containing
         names of section locations that are proximal or distal.
+
+    Examples
+    --------
+    >>> p_secs = {
+        'soma':
+        {
+            'L': 39,
+            'diam': 20,
+            'cm': 0.85,
+            'Ra': 200.,
+            'sec_pts': [[0, 0, 0], [0, 39., 0]],
+            'syns': ['ampa', 'gabaa', 'nmda'],
+            'mechs' : {
+                'ca': {
+                    'gbar_ca': 60
+                }
+            }
+        }
+    }
     """
 
-    def __init__(self, name, pos, gid=None):
+    def __init__(self, name, pos, p_secs, p_syn, topology, sect_loc, gid=None):
         self.name = name
         self.pos = pos
+        self.p_secs = p_secs
+        self.p_syn = p_syn
+        self.topology = topology
+        self.sect_loc = sect_loc
         self.sections = dict()
         self.synapses = dict()
         self.dipole_pp = list()
-        self.sect_loc = dict()
         self.rec_v = h.Vector()
         self.rec_i = dict()
         # insert iclamp
@@ -232,60 +277,26 @@ class Cell:
             child_loc = connection[3]
             child_sec.connect(parent_sec, parent_loc, child_loc)
 
-    def build(self, p_secs, p_syn, topology, sect_loc):
-        """Build cell in Neuron.
+    def build(self, align_dipole=None):
+        """Build cell in Neuron and insert dipole if applicable.
 
         Parameters
         ----------
-        p_secs : nested dict
-            Dictionary with keys as section name.
-            p_secs[sec_name] is a dictionary with keys
-            L, diam, Ra, cm, syns and mech.
-            syns is a list specifying the synapses at that section.
-            The properties of syn are specified in p_syn.
-            mech is a dict with keys as the mechanism names. The
-            values are dictionaries with properties of the mechanism.
-        p_syn : dict of dict
-            Keys are name of synaptic mechanism. Each synaptic mechanism
-            has keys for parameters of the mechanism, e.g., 'e', 'tau1',
-            'tau2'.
-        topology : list of list
-            The topology of cell sections. Each element is a list of
-            4 items in the format
-            [parent_sec, parent_loc, child_sec, child_loc] where
-            parent_sec and parent_loc are float between 0 and 1
-            specifying the location in the section to connect and
-            parent_sec and child_sec are names of the connecting
-            sections.
-        sect_loc : dict of list
-            Can have keys 'proximal' or 'distal' each containing
-            names of section locations that are proximal or distal.
-
-        Examples
-        --------
-        >>> p_secs = {
-            'soma':
-            {
-                'L': 39,
-                'diam': 20,
-                'cm': 0.85,
-                'Ra': 200.,
-                'sec_pts': [[0, 0, 0], [0, 39., 0]],
-                'syns': ['ampa', 'gabaa', 'nmda'],
-                'mechs' : {
-                    'ca': {
-                        'gbar_ca': 60
-                    }
-                }
-            }
-        }
+        align_dipole : str | None
+            If not None, a dipole will be inserted in this cell in alignment
+            with this section. The section should belong to the apical dendrite
+            of a pyramidal neuron.
         """
-        if 'soma' not in p_secs:
+        if 'soma' not in self.p_secs:
             raise KeyError('soma must be defined for cell')
-        self._create_sections(p_secs, topology)
-        self._create_synapses(p_secs, p_syn)
-        self._set_biophysics(p_secs)
-        self.sect_loc = sect_loc
+        self._create_sections(self.p_secs, self.topology)
+        self._create_synapses(self.p_secs, self.p_syn)
+        self._set_biophysics(self.p_secs)
+        if align_dipole in self.sections:
+            self.insert_dipole(align_dipole)
+        elif align_dipole is not None:
+            raise ValueError(f'align_dipole must be an exiting section of'
+                             f'the current cell or None. Got {align_dipole}.')
 
     def move_to_pos(self):
         """Move cell to position."""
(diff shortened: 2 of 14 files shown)


cjayb:
@rythorpe could you narrow down the scope of this PR? Is you intention to create an API for modifying the values of the existing/implemented biophysical parameters? In contrast to implementing new biophysics (e.g. calcium)?

rythorpe:
> @rythorpe could you narrow down the scope of this PR? Is you intention to create an API for modifying the values of the existing/implemented biophysical parameters? In contrast to implementing new biophysics (e.g. calcium)?

@cjayb This is just for modifying the values of previously existing biophysics/synapse parameters.

cjayb:
Perfect!

hnn_core/network.py
```diff
@@ -1066,6 +1066,10 @@ def plot_cells(self, ax=None, show=True):
         """
         return plot_cells(net=self, ax=ax, show=show)
 
+    def set_cell_biophysics(self, property, cell_type=None, synapse=None,
+                            receptor=None):
```
jasmainak, line 1070: I think this PR should now try to bring `cell` to the `Network` class like we discussed. And this function will just set the `cell.p_secs`

hnn_core/network.py
```diff
@@ -736,6 +758,20 @@ def gid_to_type(self, gid):
...
+        Should only be called after self.cells is populated
+        """
+        src_type = self.gid_to_type(gid)
+        if src_type not in self.cellname_list:
+            cell = None
+        else:
+            type_pos_ind = gid - self.gid_ranges[src_type][0]
+            cell = self.cells[src_type][type_pos_ind]
+        return cell
+
```
rythorpe, line 774: Ignore for now

hnn_core/network_builder.py
```diff
@@ -376,30 +375,24 @@ def _create_cells_and_drives(self, threshold, record_vsoma=False,
...
-                # figure out which cell type is assoc with the gid
-                # create cells based on loc property
+            if is_cell:
+                cell = Cell(name=_short_name(src_type), pos=src_pos, gid=gid)
+                p_secs = self.net.cell_properties[src_type]['sections']
+                p_syn = self.net.cell_properties[src_type]['synapses']
+                topology = self.net.cell_properties[src_type]['topology']
+                sect_loc = self.net.cell_properties[src_type]['sect_loc']
+                # instantiate NEURON object
+                cell.build(p_secs, p_syn, topology, sect_loc)
```
rythorpe, line 391: Note that `cell.build()` is called here instead of e.g. in `cells_default.pyramidal()`.

hnn_core/network.py
```diff
@@ -225,6 +229,24 @@ def copy(self):
...
 
+    def _set_cell_properties(self):
+        '''Set cell properties by cell type
+
+        This relies on self.cellname_list and must be updated when introducting
+        new cell types to HNN.
+        '''
+        for cell_type in self.cellname_list:
+            if cell_type in ('L2_pyramidal', 'L5_pyramidal'):
+                p_secs, p_syn, topology, sect_loc = pyramidal(
```
rythorpe, line 240: `cells_default.pyramidal` and `cells_default.basket` are only called from within `Network` now.

rythorpe:
This PR still needs some work, but I'd like some feedback before moving forward @jasmainak @cjayb.

I initially tried instantiating the population of `Cell` objects within `Network` and then building them in `network_builder`, but that was probably overkill for this PR and, more importantly, caused pickling errors. It'd be nice to have `Cell` objects accessible to the user, but idk if it's easily implementable or needed for most use cases. @cjayb would this be desirable for any of your work with LFP calculations?

jasmainak:
There shouldn't be pickling errors anymore. I think for the calcium dynamics ... see #333 ideal API would be:

```py
from hnn_core.default_cells import pyramidal_ca

net.cells['L5Pyr'] = pyramidal_ca()
```

You should first start by moving `p_secs`, `topology` etc as an attribute of `Cell` class. Check if the cell remains picklable once you do that (there is a [test](https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/tests/test_cell.py#L42) for that). Once you are able to do that, it's "just" a matter of exposing it in the `Network` class ... I'd really like us to go this way so the `Cell` class can evolve to have methods that are accessible to the user. 

For example, you can imagine a method `insert_mechanism` that updates `self.p_secs`. That way, users are free to try new mechanisms, add new synapses etc., and not just modify the existing cell.

rythorpe:
@jasmainak are you imagining that all the cells will be accessible from `Network`, or just a template cell that gets replicated downstream by `NetworkBuilder`? We had originally discussed the latter, but the former might be nice as well.

jasmainak:
template cell for now I would say. Making all the cells accessible might be helpful if we were to combine `Cell` and `CellResponse` class. See also #217 . Then the `Cell` class will begin to become a real workhorse, but we're not there yet.

ntolley:
Agreed with putting off making them all accessible until later. I really didn't see any immediate benefits while working to convert `CellResponse` to a list, the current data structure is pretty straightforward for spikes, voltages etc. If it simplifies things to merge CellResponse and Cell that's definitely great. 

Sadly that old PR is more or less just moving things around, but if there's a compelling argument to push it forward I'm all ears.

rythorpe:
I'm just finishing up running the tests, but I've left access to `Cell` objects from `Network` intact for simplicity. If we decide we'd rather wait to include this in master, it's pretty easy to migrate `Cell` population instantiation into `NetworkBuilder`.

examples/plot_simulate_gamma.py
```diff
@@ -108,6 +108,22 @@
...
 plot_psd(dpls[trial_idx], fmin=20., fmax=100., tmin=tmin)
 
+###############################################################################
+# Finally, we demonstrate the mechanistic link between PING and the GABAA decay
+# time constant (`tau2`). Using the same network/drive configuration as before,
+# we decrease `tau2` from 5 to 2 ms. This will shorten the effective
+# refactory period between L5 pyramidal cell spikes and increase the PING
+# frequency from ~50 to ~65 Hz.
+net.cell_types['L5_pyramidal'].p_syn['gabaa']['tau2'] = 2
+net.update_cells()
```
rythorpe, line 118: Here's what the API looks like for changing a biophysical synapse parameter. Note that it also produces the anticipated results before vs. after manual adjustment of `tau2`:
Screenshot from 2021-05-14 15-02-48

hnn_core/network.py
```diff
@@ -161,13 +165,13 @@ def __init__(self, params, add_drives_from_params=False,
...
-            'L5_basket',
-            'L5_pyramidal',
-        ]
+        # Source dict of names, first real ones only!
+        self.cell_types = {
+            'L2_basket': basket((0, 0, 0), cell_name=_short_name('L2_basket')),
+            'L2_pyramidal': pyramidal((0, 0, 0), _short_name('L2_pyramidal')),
+            'L5_basket': basket((0, 0, 0), cell_name=_short_name('L5_basket')),
+            'L5_pyramidal': pyramidal((0, 0, 0), _short_name('L5_pyramidal'))
+        }
```
rythorpe, line 174: The user can now also load a different cell type via

```python
net.cell_types['L5_pyramidal'] = pyramdal_ca((0, 0, 0), _short_name('L5_pyramidal'))
```
where `pyramidal_ca()` is e.g. a pyramidal cell with updated calcium dynamics imported from `cells_default.py`.

hnn_core/network.py
```diff
@@ -220,11 +227,26 @@ def copy(self):
...
+        for cell_type in self.pos_dict.keys():
+            if cell_type in self.cell_types:
+                cells = list()
+                for cell_idx, pos in enumerate(self.pos_dict[cell_type]):
+                    cell = deepcopy(self.cell_types[cell_type])
+                    cell.gid = self.gid_ranges[cell_type][cell_idx]
+                    cell.pos = pos
+                    cells.append(cell)
+                    self.cells[cell_type] = cells
+
```
rythorpe, line 249: This is where `Cell` instances are replicated to produce the cell population stored in `Network.cells`. We can leave this here for now or I can move it over to `NetworkBuilder`. Any preference @jasmainak @ntolley?

hnn_core/network.py
```diff
@@ -220,11 +227,26 @@ def copy(self):
...
         """
+        # reset cells to avoid pickling error
+        self.update_cells()
         net_copy = deepcopy(self)
         net_copy.cell_response = CellResponse(times=self.cell_response._times)
         net_copy._reset_drives()
         return net_copy
 
+    def update_cells(self):
+        '''Populate the network with cell objects'''
```
jasmainak, line 238:
```suggestion
        """Populate the network with cell objects"""
```

nitpick

examples/plot_simulate_gamma.py
jasmainak: yeaahhhh! this is the right direction to go :) Could we hide `update_cells` from the user? They shouldn't have to worry about it. Should be called during the build somewhere

jasmainak:
I didn't review properly but looks pretty clean on a superficial glance. @ntolley you want to take a first stab at the review?

hnn_core/cell.py
```diff
@@ -232,60 +277,26 @@ def _create_sections(self, p_secs, topology):
...
-                'sec_pts': [[0, 0, 0], [0, 39., 0]],
-                'syns': ['ampa', 'gabaa', 'nmda'],
-                'mechs' : {
-                    'ca': {
-                        'gbar_ca': 60
-                    }
-                }
-            }
-        }
+        align_dipole : str | None
```
jasmainak, line 285: use the same name as in `insert_dipole` ? Also not sure if `insert_dipole` should still be public method according to this change?

examples/plot_simulate_gamma.py
rythorpe: Oh good point, I think I can throw it in `NetworkBuilder._create_cells_and_drives()`.

hnn_core/cells_default.py
```diff
@@ -146,6 +146,11 @@ def _get_mechanisms(p_all, cell_type, section_names, mechanisms):
     return mech_props
 
 
+def _set_variable_mech(dist_from_soma):
+    """Set a cell mechanism based on its distance from the soma"""
+    return 1e-6 * np.exp(3e-3 * dist_from_soma)
+
+
 def basket(pos, cell_name='L2Basket', gid=None):
```
ntolley, line 154: Does this default name get used in the code anywhere?

hnn_core/cells_default.py
```diff
@@ -182,10 +187,7 @@ def basket(pos, cell_name='L2Basket', gid=None):
     for sec_name in p_secs:
         p_secs[sec_name]['sec_pts'] = sec_pts[sec_name]
 
-    cell = Cell(cell_name, pos=pos, gid=gid)
-    cell.build(p_secs, p_syn, topology, sect_loc)
-
-    return cell
+    return Cell(cell_name, pos, p_secs, p_syn, topology, sect_loc, gid)
```
jasmainak, line 190: In general, it's good practice to use keyword arguments when calling/instantiating so it's future proof if certain arguments are removed added.
```suggestion
    return Cell(cell_name, pos, 
                      p_secs=p_secs, 
                      p_syn=p_syn, 
                      topology=topology, 
                      sect_loc=sect_loc, 
                      gid=gid)
```

hnn_core/cells_default.py
```diff
@@ -182,10 +187,7 @@ def basket(pos, cell_name='L2Basket', gid=None):
...
 
 
 def pyramidal(pos, cell_name, override_params=None, gid=None):
```
ntolley, line 193: Similar to comment on basket cell, I am unsure what the preferred behavior is for `cell_name`. I think they should both have defaults or neither.

hnn_core/network.py
```diff
@@ -161,13 +165,13 @@ def __init__(self, params, add_drives_from_params=False,
...
-        # Source list of names, first real ones only!
-        self.cellname_list = [
-            'L2_basket',
-            'L2_pyramidal',
-            'L5_basket',
-            'L5_pyramidal',
-        ]
+        # Source dict of names, first real ones only!
+        self.cell_types = {
+            'L2_basket': basket((0, 0, 0), cell_name=_short_name('L2_basket')),
```
ntolley, line 170:
```suggestion
            'L2_basket': basket((0, 0, 0), _short_name('L2_basket')),
```
My comments above were related to making sure this won't break anything

hnn_core/network.py
```diff
@@ -161,13 +165,13 @@ def __init__(self, params, add_drives_from_params=False,
...
+            'L2_pyramidal': pyramidal((0, 0, 0), _short_name('L2_pyramidal')),
```
ntolley, line 171:
```suggestion
            'L2_pyramidal': pos=pyramidal((0, 0, 0), cell_name=_short_name('L2_pyramidal')),
```

hnn_core/network.py
```diff
@@ -161,13 +165,13 @@ def __init__(self, params, add_drives_from_params=False,
...
+            'L5_basket': basket((0, 0, 0), cell_name=_short_name('L5_basket')),
```
ntolley, line 172:
```suggestion
            'L5_basket': pos=basket((0, 0, 0), cell_name=_short_name('L5_basket')),
```

hnn_core/network.py
```diff
@@ -161,13 +165,13 @@ def __init__(self, params, add_drives_from_params=False,
...
+            'L5_pyramidal': pyramidal((0, 0, 0), _short_name('L5_pyramidal'))
```
ntolley, line 173:
```suggestion
            'L5_pyramidal': pos=pyramidal((0, 0, 0), cell_name=_short_name('L5_pyramidal'))
```

hnn_core/network.py
ntolley: Is there ever an instance where the position wouldn't start at `(0,0,0)`? If I understand correctly, the cell at this position is copied to the other locations, so the user doesn't really have a use for that argument.

hnn_core/viz.py
```diff
@@ -322,7 +322,7 @@ def plot_cells(net, ax=None, show=True):
     markers = {'L5_pyramidal': '^', 'L2_pyramidal': '^',
                'L5_basket': 'x', 'L2_basket': 'x'}
 
-    for cell_type in net.cellname_list:
+    for cell_type in net.cell_types:
```
ntolley, line 325: Great job catching this! I usually don't notice the viz is broken until way later..

hnn_core/cells_default.py
ntolley: On a related note, it's kind of weird that `cell_name` and `pos` get flipped in the argument list for `pyramidal` and `basket` functions.

examples/plot_simulate_gamma.py
```diff
@@ -108,6 +108,21 @@
 from hnn_core.viz import plot_psd
 plot_psd(dpls[trial_idx], fmin=20., fmax=100., tmin=tmin)
 
+###############################################################################
+# Finally, we demonstrate the mechanistic link between PING and the GABAA decay
+# time constant (`tau2`). Using the same network/drive configuration as before,
+# we decrease `tau2` from 5 to 2 ms. This will shorten the effective
+# refactory period between L5 pyramidal cell spikes and increase the PING
+# frequency from ~50 to ~65 Hz.
+net.cell_types['L5_pyramidal'].p_syn['gabaa']['tau2'] = 2
```
ntolley, line 117: @rythorpe what do you think about adding a few print statements (perhaps in another code block) to show off how you can inspect the biophysical parameter values?

rythorpe: You're thinking like a more verbose `Cell.__repr__()` method?

hnn_core/cells_default.py
rythorpe: is there a reason why we don't use keyword args for `cell_name` and `pos`?

hnn_core/cells_default.py
rythorpe: I don't think so. Let's remove it unless @jasmainak objects?

hnn_core/network.py
rythorpe: I made `pos=(0, 0, 0,)` default in https://github.com/jonescompneurolab/hnn-core/pull/321/commits/063f99eb813b2ea784dc88ac6ab8f3ee24a3e897.

hnn_core/cell.py
```diff
@@ -232,60 +277,26 @@ def _create_sections(self, p_secs, topology):
...
-        self._set_biophysics(p_secs)
-        self.sect_loc = sect_loc
+        self._create_sections(self.p_secs, self.topology)
+        self._create_synapses(self.p_secs, self.p_syn)
+        self._set_biophysics(self.p_secs)
+        if align_dipole in self.sections:
+            self.insert_dipole(align_dipole)
+        elif align_dipole is not None:
+            raise ValueError(f'align_dipole must be an exiting section of'
+                             f'the current cell or None. Got {align_dipole}.')
```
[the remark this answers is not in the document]
rythorpe: I was under the impression I had added a test earlier but turns out I just thought about it really hard. Addressed now in https://github.com/jonescompneurolab/hnn-core/pull/321/commits/4b561d8f9f3dbe34fb5e4f8fe63e3a1f651ecff1.

rythorpe:
I think I address all of your comments @jasmainak and @ntolley. LMK what you think.

examples/plot_simulate_gamma.py
```diff
@@ -108,6 +108,21 @@
...
+dpls = simulate_dipole(net, n_trials=1)
```
jasmainak, line 118: so this makes this example too long for CircleCI. Do you want to try adding '-v' to SPHINXOPTS and see if that fixes it:

https://github.com/jonescompneurolab/hnn-core/blob/master/doc/Makefile#L5

other option is to increase the [CircleCI timeout](https://circleci.com/docs/2.0/configuration-reference/#run). In general, we should think how to do continuous integration when the examples become longer and longer. 

MNE has the approach that it runs only the affected example. We could implement something more sophisticated like that if this starts turning out to be the bottleneck.

hnn_core/network_builder.py
```diff
@@ -409,7 +399,7 @@ def _create_cells_and_drives(self, threshold, record_vsoma=False,
                 # this call could belong in init of a _Cell (with threshold)?
                 nrn_netcon = cell.setup_source_netcon(threshold)
                 _PC.cell(cell.gid, nrn_netcon)
-                self.cells.append(cell)
+                self.cells.append(cell)  # XXX fix later
```
jasmainak, line 402: what is to be fixed here?

jasmainak:
PR looks great to me. Can you add a few tests, make all the CIs happy, update whats_new.rst and bring this to MRG?

hnn_core/network_builder.py
rythorpe: Right now there are identical lists of cell objects stored in `Network` and `NetworkBuilder`. If we like the cells staying in `Network`, the list in `NetworkBuilder` should be decommissioned.

hnn_core/network.py
```diff
@@ -107,12 +108,15 @@ class Network(object):
 
     Attributes
     ----------
-    cellname_list : list
-        The names of real cell types in the network (e.g. 'L2_basket')
+    cell_types : dict
```
jasmainak, line 111: there are [other places in the repo](https://github.com/jonescompneurolab/hnn-core/search?q=cell_types) where `cell_types` is a different kind of object. This could be potentially confusing

jasmainak: `cell_templates` maybe?

hnn_core/network_builder.py
jasmainak: Oh I see. Yeah, I would decommission the `cells` in `NetworkBuilder`. In terms of clean up, `NetworkBuilder` is my next target after a couple more things in `cell`, so this would definitely be up that alley. Can it be done in this PR?

hnn_core/network.py
rythorpe: Given that I converted `cell_types` from `cell_list`, this attribute now serves two purposes: to document an exhaustive list of cells that are represented in the network (e.g., as in [here](https://github.com/jonescompneurolab/hnn-core/blob/32d33cc96bd788f6c96b97339a98788b2cf3810e/hnn_core/network.py#L720) or [here](https://github.com/jonescompneurolab/hnn-core/blob/32d33cc96bd788f6c96b97339a98788b2cf3810e/hnn_core/network.py#L697)), and also to specify the cell (class) template that will be used to represent each cell type. So if a new user or developer wants to see which cells are implemented in an instance of `Network`, (e.g., in conjunction with functions like `plot_cell_morphology()`, they should look towards `Network.cell_types` for guidance.

Toward your point though, maybe what we want to do is have a `cell_types` attribute of type list that mirrors `Network.cell_templates.keys()`?

jasmainak: Right, and ignore `plot_cell_morphology` for now. I'm going to move that into the cell class, so that will become irrelevant. So, it's only in `cell_response.py` that we have a bit of mismatch. You could go either way -- rename that one instance of `cell_types` and/or add a `property` function that returns `cell_types` using `cell_templates.keys()`

rythorpe: So you're thinking just rename `cell_types` [here](https://github.com/jonescompneurolab/hnn-core/blob/70a38d6b43a0801a88b4d8c4742c644704eed828/hnn_core/cell_response.py#L284)?

jasmainak: yep!

examples/plot_simulate_gamma.py
ntolley: Yeah exactly! I'm getting of the mindset that any API we add to modify parameters will need to come with a really straightforward way to inspect those parameters.

rythorpe: That sounds great actually. I'll leave that for the next PR since there's talk of potentially combining the `Cell` and `CellResponse` classes.

hnn_core/network.py
rythorpe: What do you think of https://github.com/jonescompneurolab/hnn-core/pull/321/commits/1fff52c13c14a4d4feaee379afa786a6e427c19e @jasmainak ? My thinking here was to also reduce the amount of hardcoding we have for cell type names. cc @ntolley

jasmainak: I like it!

hnn_core/network_builder.py
rythorpe: Upon reviewing this further, I think it makes sense to leave `NetworkBuilder.cells` as an attribute since it is a clear way for a `NetworkBuilder` instance to keep track of the cells assigned to the current rank (e.g., as exemplified when [destroying Neuron objects](https://github.com/jonescompneurolab/hnn-core/blob/32d33cc96bd788f6c96b97339a98788b2cf3810e/hnn_core/network_builder.py#L543-L560) upon exiting a `NetworkBuilder` context. Note that I have now made this attribute private in https://github.com/jonescompneurolab/hnn-core/pull/321/commits/7b014072c80c859272f00295f17892fea89e0460 since `NetworkBuilder._cells` it is now solely for utility and should never be accessed by even an advanced user.

hnn_core/tests/test_cell_response.py
```diff
@@ -36,9 +36,10 @@ def test_cell_response(tmpdir):
     assert len(cell_response.times) == len(sim_times)
     sim_attributes = ['_spike_times', '_spike_gids', '_spike_types',
                       '_vsoma', '_isoma']
-    net_attributes = ['_times']  # `Network.__init__` creates these
-    # check that we always know which response attributes are simulated
-    # see #291 for discussion; objective is to keep cell_response size small
+    net_attributes = ['_times', '_cell_type_names']  # `Network.__init__`
```
jasmainak, line 39: did you investigate the impact on the size of the `network` object? You can use the code here

https://github.com/jonescompneurolab/hnn-core/issues/298#issue-843945983 ?

jasmainak:
@rythorpe you have a merge conflict

jasmainak:
probably need to rebase with latest master

hnn_core/tests/test_network.py
```diff
@@ -44,6 +44,21 @@ def test_network():
...
         assert len(net.external_drives[dn]['events']) == 1  # single trial!
 
+    # Test that a change in net.pos_dict influences the number of cells created
+    net_new = net.copy()
+    n_cells_orig = net_new.n_cells
+    new_cell_type = {'new_type': net_new.cell_types['L2_basket']}
+    net_new.pos_dict.update({'new_type': [(0, 0, 0)]})
+    net_new._update_gid_ranges()
+    net_new._update_cells()
+    # Without also updating net_new.cell_types, no new cells are created
```
jasmainak, line 54: this test is a bit confusing. Ideally, an error should have been thrown in this scenario, no?

jasmainak (approved):
Tiny nitpicks about test otherwise LGTM. @ntolley @cjayb feel free to merge when you are happy about this PR

hnn_core/tests/test_network.py
rythorpe: I went back and forth on how to implement this test quite a bit. My original intention was to validate that `Network.cell_type` acts as a gate for creating cells. In theory, the user should be able to delete a cell type from the network without throwing an error; however, the API isn't completely setup yet to accommodate such a scenario. I simplified the test in https://github.com/jonescompneurolab/hnn-core/pull/321/commits/e1bfa153f4cec7c71185afb9ade1a72d091ced83 by removing this assertion statement.

rythorpe: Does this work @jasmainak or would you prefer that we throw an error if/when the cell types in `pos_dict` and `cell_types` become incongruent?

rythorpe:
@jasmainak I tried running

```python
import os.path as op
from mne.utils import object_size
import hnn_core
from hnn_core import read_params, Network

net = Network(params)

size = object_size(net) / (1024 * 1024)
```
but get a runtime error reporting that type `Network` is unsupported. Am I doing this right?

hnn_core/network_builder.py
```diff
@@ -376,30 +373,21 @@ def _create_cells_and_drives(self, threshold, record_vsoma=False,
...
         """
-        type2class = {'L2_pyramidal': pyramidal, 'L5_pyramidal': pyramidal,
-                      'L2_basket': basket, 'L5_basket': basket}
+        self.net._update_cells()
         # loop through ALL gids
         # have to loop over self._gid_list, since this is what we got
         # on this rank (MPI)
-
         for gid in self._gid_list:
-            src_type, src_pos, is_cell = self.net._get_src_type_and_pos(gid)
```
jasmainak, line 386: if you're removing the call to this function, shouldn't you remove the function as well?

jasmainak:
> Am I doing this right?

humm ... indeed. I get this too. MNE probably supports only certain known datatypes etc for this. Anyhow, I deleted my comment because I realized that the `cell_type_names` list is only containing 4 items (right?), so it shouldn't bloat things too much. You could probably just measure `sizeof(cell_type_names)` for curiosity if you want

hnn_core/network_builder.py
rythorpe: Good catch! Addressed in https://github.com/jonescompneurolab/hnn-core/pull/321/commits/e9c79bf5cb3f9047ec463f00e9d9986965034121.

hnn_core/tests/test_network.py
jasmainak: In a follow up PR, I think it would be great to put in checks to ensure user does not inconsistently update `pos_dict` / `cell_types`. One way is to make this attributes private so user is not allowed to touch them and have one function/method which can be used to update them consistently. Stuff like [this](https://github.com/rythorpe/hnn-core/blob/e9c79bf5cb3f9047ec463f00e9d9986965034121/hnn_core/network.py#L526) scares me ...

jasmainak:
I was brave and went ahead and merged the PR. Thanks for the great effort @rythorpe !

What's the next PR? :)

rythorpe:
> humm ... indeed. I get this too. MNE probably supports only certain known datatypes etc for this. Anyhow, I deleted my comment because I realized that the `cell_type_names` list is only containing 4 items (right?), so it shouldn't bloat things too much. You could probably just measure `sizeof(cell_type_names)` for curiosity if you want

Oh I see. I do want to be conscious of how big `Network` gets though because it can bog things down. Hopefully the recent changes keep it slim....

rythorpe:
> I was brave and went ahead and merged the PR. Thanks for the great effort @rythorpe !
> 
> What's the next PR? :)

Woooo! Thanks for the reviews! Have we revitalized the updated calcium dynamics PR?

jasmainak:
Take a look at #333 to see what's left to do. We need to update the examples to make them look like before. How are we going to do the ground-truth tests once calcium dynamics come? Do we do #314 first to update the ground truth?

jasmainak:
>  I do want to be conscious of how big Network gets though because it can bog things down.

we probably want to deprecate `net._params` further down the line so there is no duplication of parameters through the Cell object. It's also cleaner structurally and code-wise.

PR #368, opened 2021-06-18
[MRG] Fix average_dipoles to allow for one trial

### LINKED ISSUE #306: average_dipoles should work with a single dipole

kenloi:
Closes #306
Fixed bug in average_dipoles() function in dipole.py, now allowing for computation of an average even for n_of_1 simulation. I essentially had the function just return the first and only dipole instance in the input parameter dipoles list.

For n_of_1 trials, I decided to also include a message to let users know that when calling an average on a sample size of 1, they should consider simulating at least two trials to properly produce an average.

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -41,6 +41,8 @@ Bug
 
 - Fix bug where depth of L5 and L2 cells were swapped, by `Christopher Bailey`_ in `#352 <https://github.com/jonescompneurolab/hnn-core/pull/352>`_
 
+- Fix bug where :func:`~hnn_core.average_dipole` failed when there were less than two dipoles in the input dipole list, by `Kenneth Loi`_ in `#368 <https://github.com/jonescompneurolab/hnn-core/pull/368>`_
+
 API
 ~~~
 - New API for defining cell-cell connections. Custom connections can be added with :func:`~hnn_core.Network.add_connection`, by `Nick Tolley`_ in `#276 <https://github.com/jonescompneurolab/hnn-core/pull/276>`_
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -41,7 +41,7 @@ Bug
 
 - Fix bug where depth of L5 and L2 cells were swapped, by `Christopher Bailey`_ in `#352 <https://github.com/jonescompneurolab/hnn-core/pull/352>`_
 
-- Fix bug where :func:`~hnn_core.average_dipole` failed when there were less than two dipoles in the input dipole list, by `Kenneth Loi`_ in `#368 <https://github.com/jonescompneurolab/hnn-core/pull/368>`_
+- Fix bug where average_dipole() failed when there were less than two dipoles in the input dipole list, by `Kenneth Loi`_ in `#368 <https://github.com/jonescompneurolab/hnn-core/pull/368>`_
 
 API
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -41,7 +41,7 @@ Bug
 
 - Fix bug where depth of L5 and L2 cells were swapped, by `Christopher Bailey`_ in `#352 <https://github.com/jonescompneurolab/hnn-core/pull/352>`_
 
-- Fix bug where average_dipole() failed when there were less than two dipoles in the input dipole list, by `Kenneth Loi`_ in `#368 <https://github.com/jonescompneurolab/hnn-core/pull/368>`_
+- Fix bug where :func:`~hnn_core.average_dipole` failed when there were less than two dipoles in the input dipole list, by `Kenneth Loi`_ in `#368 <https://github.com/jonescompneurolab/hnn-core/pull/368>`_
 
 API
 ~~~
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -113,7 +113,7 @@ def average_dipoles(dpls):
         average over the same components in the input list
     """
     for dpl_idx, dpl in enumerate(dpls):
-        if not dpls:
+        if not isinstance(dpl, Dipole):
             raise ValueError("Need at least one dipole object to return a"
                              " dipole")
         if dpl.nave > 1:
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -114,8 +114,9 @@ def average_dipoles(dpls):
     """
     for dpl_idx, dpl in enumerate(dpls):
         if not isinstance(dpl, Dipole):
-            raise ValueError("Need at least one dipole object to return a"
-                             " dipole")
+            raise ValueError(
+                f"All elements in the list should be instances of "
+                f"Dipole. Got {type(dpl)}")
         if dpl.nave > 1:
             raise ValueError("Dipole at index %d was already an average of %d"
                              " trials. Cannot reaverage" %
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -112,16 +112,10 @@ def average_dipoles(dpls):
         A new dipole object with each component of `dpl.data` representing the
         average over the same components in the input list
     """
-    # need at least one Dipole to get times
-    if len(dpls) == 1:
-        print('Only one dipole passed in. '
-              'Consider at least two dipole objects to compute an average.')
-        return dpls[0]
-    elif not dpls:
-        raise ValueError("Need at least one dipole object to return a"
-                         " dipole")
-
     for dpl_idx, dpl in enumerate(dpls):
+        if not dpls:
+            raise ValueError("Need at least one dipole object to return a"
+                             " dipole")
         if dpl.nave > 1:
             raise ValueError("Dipole at index %d was already an average of %d"
                              " trials. Cannot reaverage" %
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -113,9 +113,13 @@ def average_dipoles(dpls):
         average over the same components in the input list
     """
     # need at least one Dipole to get times
-    if len(dpls) < 2:
-        raise ValueError("Need at least two dipole object to compute an"
-                         " average")
+    if len(dpls) == 1:
+        print('Only one dipole passed in. '
+              'Consider at least two dipole objects to compute an average.')
+        return dpls[0]
+    elif not dpls:
+        raise ValueError("Need at least one dipole object to return a"
+                         " dipole")
 
     for dpl_idx, dpl in enumerate(dpls):
         if dpl.nave > 1:
--- a/hnn_core/tests/test_dipole.py
+++ b/hnn_core/tests/test_dipole.py
@@ -76,7 +76,6 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
     # test postproc
 
     # XXX all below to be deprecated in 0.3
-
     dpls_raw, net = run_hnn_core_fixture(backend='joblib', n_jobs=1,
                                          reduced=True, record_isoma=True,
                                          record_vsoma=True)
--- a/hnn_core/tests/test_dipole.py
+++ b/hnn_core/tests/test_dipole.py
@@ -73,8 +73,6 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
     for dpl_key in dpl_avg.data.keys():
         assert_allclose(dpl_1[0].data[dpl_key] / 2., dpl_avg.data[dpl_key])
 
-    # test postproc
-
     # XXX all below to be deprecated in 0.3
     dpls_raw, net = run_hnn_core_fixture(backend='joblib', n_jobs=1,
                                          reduced=True, record_isoma=True,
--- a/hnn_core/tests/test_dipole.py
+++ b/hnn_core/tests/test_dipole.py
@@ -64,6 +64,8 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
             single_dpl_avg.data[dpl_key],
             rtol=0,
             atol=0.000051)
+        np.testing.assert_allclose(dipole_read.data[dpl_key],
+                    single_dpl_avg.data[dpl_key], rtol=0, atol=0.000051)
 
     # average dipole list with one dipole object and a zero dipole object
     n_times = len(dipole_read.data['agg'])
@@ -71,7 +73,8 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
     dpl_1 = [dipole, dpl_null]
     dpl_avg = average_dipoles(dpl_1)
     for dpl_key in dpl_avg.data.keys():
-        assert_allclose(dpl_1[0].data[dpl_key] / 2., dpl_avg.data[dpl_key])
+        np.allclose(dpl_1[0].data[dpl_key],
+                    dpl_avg.data[dpl_key])
 
     # test postproc
 
--- a/hnn_core/tests/test_dipole.py
+++ b/hnn_core/tests/test_dipole.py
@@ -59,13 +59,11 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
     # average an n_of_1 dipole list
     single_dpl_avg = average_dipoles([dipole])
     for dpl_key in single_dpl_avg.data.keys():
-        assert_allclose((
+        np.testing.assert_allclose(
             dipole_read.data[dpl_key],
             single_dpl_avg.data[dpl_key],
             rtol=0,
             atol=0.000051)
-        np.testing.assert_allclose(dipole_read.data[dpl_key],
-                    single_dpl_avg.data[dpl_key], rtol=0, atol=0.000051)
 
     # average dipole list with one dipole object and a zero dipole object
     n_times = len(dipole_read.data['agg'])
--- a/hnn_core/tests/test_dipole.py
+++ b/hnn_core/tests/test_dipole.py
@@ -71,8 +71,8 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
     dpl_1 = [dipole, dpl_null]
     dpl_avg = average_dipoles(dpl_1)
     for dpl_key in dpl_avg.data.keys():
-        np.allclose(dpl_1[0].data[dpl_key],
-                    dpl_avg.data[dpl_key])
+        assert np.allclose([data / 2 for data in dpl_1[0].data[dpl_key]],
+                           dpl_avg.data[dpl_key])
 
     # test postproc
 
--- a/hnn_core/tests/test_dipole.py
+++ b/hnn_core/tests/test_dipole.py
@@ -56,7 +56,27 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
                        "average of 2 trials"):
         dipole_avg = average_dipoles([dipole_avg, dipole_read])
 
+    # average an n_of_1 dipole list
+    single_dpl_avg = average_dipoles([dipole])
+    for dpl_key in single_dpl_avg.data.keys():
+        assert_allclose((
+            dipole_read.data[dpl_key],
+            single_dpl_avg.data[dpl_key],
+            rtol=0,
+            atol=0.000051)
+
+    # average dipole list with one dipole object and a zero dipole object
+    n_times = len(dipole_read.data['agg'])
+    dpl_null = Dipole(np.zeros(n_times, ), np.zeros((n_times, 3)))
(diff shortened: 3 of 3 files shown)


hnn_core/network.py
```diff
@@ -887,7 +887,7 @@ def _instantiate_drives(self, n_trials=1):
                     drive['name']]['events'].append(event_times)
 
     def add_tonic_bias(self, *, cell_type=None, amplitude=None,
-                       t0=None, T=None):
+                       t0=None, tstop=None):
```
jasmainak, line 890: @kenloi it looks like there are changes from your old pull request in this pull request. Every time you make a new PR, you should fetch a fresh copy of master and branch from there:

```sh
$ git fetch upstream master:master
$ git checkout master
$ git branch fix_average_dipoles
$ git checkout fix_average_dipoles
```

now since you already started the work, I would recommend using `git cherry-pick` in the new branch. What you should do to fix the current situation is the following:

```sh
$ git branch -m fix_average_dipoles_old
$ git fetch upstream master:master
$ git checkout master
$ git branch fix_average_dipoles
$ git checkout fix_average_dipoles
$ git cherry-pick <commit1-hash>
$ git cherry-pick <commit2-hash>
$ git push -f origin fix_average_dipoles
```

does that make sense?

kenloi: Just read this^ @jasmainak. That makes sense. I'll try and fix my current branch. Thanks!

hnn_core/dipole.py
```diff
@@ -102,9 +102,13 @@ def average_dipoles(dpls):
...
-    if len(dpls) < 2:
-        raise ValueError("Need at least two dipole object to compute an"
-                         " average")
+    if len(dpls) == 1:
+        print('Only one dipole passed in. '
+              'Consider at least two dipole objects to compute an average.')
+        return dpls[0]
+    elif not dpls:
+        raise ValueError("Need at least one dipole object to return a"
+                         " dipole")
```
jasmainak, line 111: I would instead raise an error in the loop below if any of the objects in the list is not an instance of `Dipole`. Look around in the codebase to see how it's done

Could you also update a test after doing this? So you get familiar with the test suite and how it works?

hnn_core/dipole.py
```diff
@@ -102,9 +102,13 @@ def average_dipoles(dpls):
         average over the same components in the input list
     """
     # need at least one Dipole to get times
-    if len(dpls) < 2:
-        raise ValueError("Need at least two dipole object to compute an"
-                         " average")
+    if len(dpls) == 1:
```
jasmainak, line 105: I think this function shouldn't throw any warnings if:

1) You get a list (with len > 0)
2) Each element of the list is a `Dipole` object

Also, if you really want to throw a warning, you should use the `warnings` module:

```py
from warnings import warn

warn('important message')
```

hnn_core/dipole.py
kenloi: Yea, that makes sense. I'll put the error detection into the loop and update a test.

hnn_core/dipole.py
kenloi: @jasmainak This is me just nitpicking but I was thinking that it might help to let users know that when calling the average_dipoles() function when only one trial was run, they are effectively not averaging anything. I'm cool with removing the message if you think it doesn't help much.

cjayb: I also prefer to limit the number of warnings and only throw them when there's something to be _careful_ about. I think `average` should work silently for a single instance, makes it more flexible.

kenloi: Alrighty, I'll remove the message :)

hnn_core/dipole.py
kenloi: Updated both the loop and the test. Haven't created test cases before so I hope that what I did makes sense. I used the `# average  two identical dipole objects` as a guide for creating the single dipole object average test. Are there docs that explain the standards for developing test cases with examples?

hnn_core/tests/test_dipole.py
```diff
@@ -48,6 +48,12 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
                        "average of 2 trials"):
         dipole_avg = average_dipoles([dipole_avg, dipole_read])
 
+    # average single dipole object
+    dpl_avg1 = average_dipoles([dipole])
```
ntolley, line 52: What about this for a test. You can create a `Dipole` object with two simulations, but second the data stored under the second trials to `dpl[1].data = np.zeroes(len(dpl[0].times))`.

Then your test can be 
```py
assert np.all(dpl_avg == dpl[0].data / 2)
```
I think the current test is good and should remain, but it'd be good to make sure the average function is actually doing some averaging too.

hnn_core/dipole.py
```diff
@@ -101,12 +101,10 @@ def average_dipoles(dpls):
         A new dipole object with each component of `dpl.data` representing the
         average over the same components in the input list
     """
-    # need at least one Dipole to get times
-    if len(dpls) < 2:
-        raise ValueError("Need at least two dipole object to compute an"
-                         " average")
-
     for dpl_idx, dpl in enumerate(dpls):
+        if not dpls:
```
jasmainak, line 105: this is not a type check. If I give `dpls = True`, then this check will not raise any error. You need to use `isinstance` ... and for `dpl` not `dpls`

ntolley: Oops I'll take the heat for this, I think this was a suggestion I gave in a brief meeting with @kenloi. Misunderstood the purpose of the original check, `isinstance()` is definitely the way to go.

Alternatively, check out how [`_validate_type()`](https://github.com/jonescompneurolab/hnn-core/blob/9b95ae83918cf91352f7b342e81fd19e22ffdcc3/hnn_core/network.py#L257l) is used in the code. It provides a boilerplate error message as these sorts of checks come up frequently.

kenloi: @jasmainak That makes sense. I'll take care of it asap. 
@ntolley Thanks for the suggestion, will check it out.

hnn_core/tests/test_dipole.py
kenloi: @ntolley Sounds good! I'll give it a try.

kenloi: @ntolley I tried your implementation of setting the data for the second trial to zero. However the assertion
> `assert np.all(dpl_avg == dpl[0].data / 2)`

causes the entire array to be set to 0. floats. This raises an IndexError later when trying to compute the average because numpy isn't able to compute a mean for:
> `[0. 0. 0. ... 0. 0. 0.]`

I'll try and find a way around this in the meantime. I might have to rely on a different way to set the data to 0.

jasmainak: you should use `np.allclose` to compare arrays

hnn_core/tests/test_dipole.py
```diff
@@ -48,6 +48,23 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
                        "average of 2 trials"):
         dipole_avg = average_dipoles([dipole_avg, dipole_read])
 
+    # average an n_of_1 dipole list
+    single_dpl_avg = average_dipoles([dipole])
+    for dpl_key in single_dpl_avg.data.keys():
+        np.allclose(dipole_read.data[dpl_key],
+                    single_dpl_avg.data[dpl_key], rtol=0, atol=0.000051)
```
jasmainak, line 55: the test won't fail with this, sorry my bad. You have to use:

https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html

hnn_core/tests/test_dipole.py
```diff
@@ -48,6 +48,23 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
...
+    # average dipole list with one dipole object and a zero dipole object
+    dpl_null = dipole_read
+    dpl_null.data['agg'] = np.zeros(len(dpl_null.data['agg']))
+    dpl_null.data['L2'] = np.zeros(len(dpl_null.data['L2']))
+    dpl_null.data['L5'] = np.zeros(len(dpl_null.data['L5']))
+    dpl_1 = [dipole, dpl_null]
+    dpl_avg = average_dipoles(dpl_1)
+    for dpl_key in dpl_avg.data.keys():
+        np.allclose(dpl_1[0].data[dpl_key],
+                    dpl_avg.data[dpl_key], rtol=0, atol=0.000051)
```
jasmainak, line 66:
```suggestion
                    dpl_avg.data[dpl_key])
```

very strange these numbers are. I think we should just go with the defaults if it works

hnn_core/tests/test_dipole.py
```diff
@@ -48,6 +48,23 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
...
+    single_dpl_avg = average_dipoles([dipole])
+    for dpl_key in single_dpl_avg.data.keys():
+        np.allclose(dipole_read.data[dpl_key],
+                    single_dpl_avg.data[dpl_key], rtol=0, atol=0.000051)
+
+    # average dipole list with one dipole object and a zero dipole object
+    dpl_null = dipole_read
+    dpl_null.data['agg'] = np.zeros(len(dpl_null.data['agg']))
+    dpl_null.data['L2'] = np.zeros(len(dpl_null.data['L2']))
+    dpl_null.data['L5'] = np.zeros(len(dpl_null.data['L5']))
```
jasmainak, line 61: why not just use the constructor: https://jonescompneurolab.github.io/hnn-core/dev/generated/hnn_core.dipole.Dipole.html#hnn_core.dipole.Dipole

```py
n_times = len(dipole_read)
dpl_zeros = Dipole(np.zeros(n_times, ), np.zeros(n_times, 3))
```

hnn_core/dipole.py
```diff
@@ -101,12 +101,10 @@ def average_dipoles(dpls):
...
     """
-    # need at least one Dipole to get times
-    if len(dpls) < 2:
-        raise ValueError("Need at least two dipole object to compute an"
-                         " average")
-
     for dpl_idx, dpl in enumerate(dpls):
+        if not isinstance(dpl, Dipole):
+            raise ValueError("Need at least one dipole object to return a"
+                             " dipole")
```
jasmainak, line 107: there needs to be a test to check that this works. Or is there already a test? Also, I think the text should be changed to reflect the fact that all objects in the list must instances of `Dipole`. Also, when you raise an error, you should tell the user what you got. Something like:

```suggestion
            raise ValueError(f"All elements in the list should be instances of Dipole. Got {type(dpl)}")
```

kenloi: Yea, this test already exists. It's under line 51, the comment `# average an n_of_1 dipole list` which calls `average_dipoles()` on a list with only one dipole.

hnn_core/tests/test_dipole.py
kenloi: @jasmainak You're right, it works without the defaults. I set the numbers because I saw it used in the previous tests and figured that those were the accepted tolerances.

hnn_core/tests/test_dipole.py
kenloi: @jasmainak I like that. Much cleaner. Will go make the changes.

hnn_core/tests/test_dipole.py
```diff
@@ -48,6 +48,24 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
...
+            rtol=0,
+            atol=0.000051)
+
+    # average dipole list with one dipole object and a zero dipole object
+    n_times = len(dipole_read.data['agg'])
+    dpl_null = Dipole(np.zeros(n_times, ), np.zeros((n_times, 3)))
+    dpl_1 = [dipole, dpl_null]
+    dpl_avg = average_dipoles(dpl_1)
+    for dpl_key in dpl_avg.data.keys():
+        np.allclose(dpl_1[0].data[dpl_key],
```
ntolley, line 66: Maybe I am misunderstanding the code I think this should fail. Since you're averaging a `dipole` of random numbers with zeros, shouldn't every value be cut in half?

ntolley: And I see the tests passing so now I'm extra confused 😄

kenloi: @ntolley Yea, I see what you're talking about, I don't think it should be working either. I'll try debugging and see what's really happening.

jasmainak: it's because `np.allclose` returns a bool. You need an assert statement to make it fail

kenloi: Thanks @jasmainak. I think I was able to fix it with your suggestion. Thanks @ntolley for bringing it to my attention.

kenloi:
@ntolley @jasmainak @cjayb I think this issue is complete and ready for a final review. Took into account everyone's input. If anyone has the time, please take a look and let me know if there are any else to fix. I've also updated the whats_new.rst documentation.

ntolley (approved):
Looks great @kenloi, I think this is good to merge after a rebase (we'll make sure to merge this PR before any others since you've had to rebase a few times already...)

kenloi:
Thanks @ntolley . I rebased again about an hour ago. Hopefully that resolves some of the conflicts?

ntolley:
So it look like you have a merge conflict in `test_dipole.py`: 
image

Esentially if there have been any pull requests merge since your last rebase, you need to rebase again. Your IDE should have an interface to help resolve conflicts similar to this: https://code.visualstudio.com/docs/editor/versioncontrol#_merge-conflicts.

The last PR merged added some code to `test_dipole.py`. Git is asking you to edit your most recent commit as though you added your code to this new version of  `test_dipole.py`.

jasmainak:
@kenloi you'll get the hang of github soon :) Remember for contributing to public repositories, only rebase is allowed. Don't merge the `master` in your branch as it will create a merge commit that is unrelated to your pull request. You should do:

```sh
$ git fetch upstream master:master
$ git rebase master
```

hnn_core/tests/test_dipole.py
```diff
@@ -56,7 +56,28 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
                        "average of 2 trials"):
         dipole_avg = average_dipoles([dipole_avg, dipole_read])
 
+    # average an n_of_1 dipole list
+    single_dpl_avg = average_dipoles([dipole])
+    for dpl_key in single_dpl_avg.data.keys():
+        np.testing.assert_allclose(
```
jasmainak, line 62:
```suggestion
        assert_allclose(
```

hnn_core/tests/test_dipole.py
```diff
@@ -56,7 +56,28 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
...
+            rtol=0,
+            atol=0.000051)
+
+    # average dipole list with one dipole object and a zero dipole object
+    n_times = len(dipole_read.data['agg'])
+    dpl_null = Dipole(np.zeros(n_times, ), np.zeros((n_times, 3)))
+    dpl_1 = [dipole, dpl_null]
+    dpl_avg = average_dipoles(dpl_1)
+    for dpl_key in dpl_avg.data.keys():
+        assert np.allclose([data / 2 for data in dpl_1[0].data[dpl_key]],
```
jasmainak, line 74:
```suggestion
        assert_allclose(dpl_1[0].data[dpl_key] / 2.,
```

doc/whats_new.rst
```diff
@@ -41,6 +41,8 @@ Bug
 
 - Fix bug where depth of L5 and L2 cells were swapped, by `Christopher Bailey`_ in `#352 <https://github.com/jonescompneurolab/hnn-core/pull/352>`_
 
+-Fix bug where average_dipole() failed when there were less than two dipoles in the input dipole list, by `Kenneth Loi`_ in `#368 <https://github.com/jonescompneurolab/hnn-core/pull/368>`_
```
jasmainak, line 44:
```suggestion
-Fix bug where :func:`~hnn_core.average_dipole` failed when there were less than two dipoles in the input dipole list, by `Kenneth Loi`_ in `#368 <https://github.com/jonescompneurolab/hnn-core/pull/368>`_
```

doing this will link to the functions in the API page

kenloi:
I've rebased and implemented the changes suggested by @jasmainak. However, it looks like I've failed one of the tests which is reporting:
`FAILED hnn_core/tests/test_parallel_backends.py::TestParallelBackends::test_run_mpibackend`

I ran pytest on the `test_parallel_backends.py` and it passed all the test cases and returned a warning:
`UserWarning: No currently running process to terminate`

I'll try digging into the details of the report, but I'm not sure what this means.

ntolley:
No need to worry about the tests, if only one of the builds is failing that usually means it's a problem on the server side. I just restarted the jobs and expect them to pass (for some reason our tests can be pretty finicky, I'm mildly suspicious it's a neuron problem...)

Also from the code review suggestions, if it looks reasonable you can actually commit them directly with the `Commit suggestion` button, and then `git pull` from your local computer to get the updated code.

kenloi:
@ntolley Yea, I'm quickly realizing that my workflow is pretty inefficient. Thanks for the tip!

kenloi:
@jasmainak @cjayb I think this PR is ready for merging. @ntolley has reviewed and approved. Just to double check, if I think the PR is ready to be merged, at what point do I switch the tag to `[MRG]`? Do I wait until I get at least two review approvals or before that? Appreciate all the feedback so far.

ntolley:
You can go ahead and switch the tag! I'm not sure if we have an official system for [MRG], but in general we just use it to signal that we're ready for reviews (not that the PR is actually perfect). WIP is meant to indicate that we're not quite ready for input yet and are aware glaring errors.

jasmainak:
yeah, [MRG} is to indicate you are yourself satisfied with what you have done and would be okay if the PR was merged as it stands. [WIP] means you are just beginning to work on it, early feedback is welcome but not nitpicks ...

hnn_core/tests/test_dipole.py
```diff
@@ -56,7 +56,27 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
...
+    # average dipole list with one dipole object and a zero dipole object
+    n_times = len(dipole_read.data['agg'])
+    dpl_null = Dipole(np.zeros(n_times, ), np.zeros((n_times, 3)))
+    dpl_1 = [dipole, dpl_null]
+    dpl_avg = average_dipoles(dpl_1)
+    for dpl_key in dpl_avg.data.keys():
+        assert_allclose(dpl_1[0].data[dpl_key] / 2., dpl_avg.data[dpl_key])
+
+    # test postproc
+
```
jasmainak, line 77:
```suggestion
```

seems like a spurious comment not related to PR

hnn_core/tests/test_dipole.py
```diff
@@ -56,7 +56,27 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
...
     # XXX all below to be deprecated in 0.3
+
```
jasmainak, line 79:
```suggestion
```

while this is not technically an issue and a nitpick in this PR, I just want to get you used to good habits from the start. Don't touch anything that's not related to your PR: even blank lines. It can make the job of the reviewer hard in longer PRs

jasmainak:
@kenloi two final nitpicks. Once you fix them, it's +1 for merge from my end. 

Hope you're learning stuff and not losing motivation. Future PRs will definitely be easier and fun!

hnn_core/tests/test_dipole.py
kenloi: This comment was in the code before I had started my PR. While rebasing, I went back to an old commit to copy and paste the changes that I had made including the comment. I assumed that the comment was made by a previous commit and referenced the subsequent test cases. Should I keep this comment or just remove it?

ntolley: You can go ahead and remove it no longer exists on the master branch. Before #372 was merged it was here: https://github.com/jonescompneurolab/hnn-core/blob/5ade97af0d6da7ab223442ca7cd954b774156d3f/hnn_core/tests/test_dipole.py#L51
 but it's been replaced now

jasmainak:
Thanks @kenloi ! Looking forward to the next contribution :)

PR #329, opened 2021-05-10
[MRG] Improved LFP calculation

cjayb:
This PR builds on and supersedes #150 (commits by @ntolley and @jasmainak retained in squash)

- [x] account for all transmembrane currents (not just that through the mid point `sec(0.5)`; see [this gist](https://gist.github.com/cjayb/56a93385db003a009cac651516d299e2) for details)
- [x] validate units and shift conductance units to S/m (default: 0.3)
- [x] produces expected pattern of near-field current sinks close to active synapses (see [second figure in this gist](https://gist.github.com/cjayb/56a93385db003a009cac651516d299e2))
- [x] produces far-field dipolar field patterns ([see gist](https://gist.github.com/cjayb/3daec37d17f5767071029bfdc95230cf))
- [x] fix PSA and LSA implementations
- [x] illustrate laminar LFP ~and CSD~ calculation in (weak) gamma example
- [x] test that net transmembrane current is (very close to) zero (implemented by setting all transfer resistances to unity)
- [x] test that far-field LFPs is dipolar
- [x] test that transfer resistances are correct
- [x] consolidate NEURON and LFP electrode coordinates (NB y-coordinate)
- [x] create a suitable example (what do we want to demonstrate?)
- [x] update `whats_new.rst` & `api.rst`

* ~create demo using single instance of `Cell`~ (move to #323)
* ~add tests for validity of PSA and LSA values close to cells (how?)~ not needed as long as the above three tests pass
* ~optimise for speed~ future PR should improve implementation, possibly getting rid of `extra_gather_scatter`

Closes #68

PR #370, opened 2021-06-19
[MRG] specify in plane distance of cells

### LINKED ISSUE #342: Incompatible cell positions and morphological extents
As discussed in many-a PR, cell positions and orientations need to be formalised.

A related and more pressing matter is that the in-plane distances between pyramidal cells (10 x 10 grid with 1 um spacing) are inconsistent with the physical extents of the sections:

```python
        'L5Pyr_soma_diam': 28.9,
        'L5Pyr_apicaltrunk_diam': 10.2,
        'L5Pyr_apical1_diam': 7.48,   #etc
```

The packing density of cells in cortex is insane, but real cells don't intersect, as ours do ;)

I don't know of a principled way to deal with this, since our "cells" aren't really cells, but models of (multiple) cells. But let's say we defined the cell-to-cell lateral distance in HNN to be 30 um. The somas would be 1 um apart, which is quite realistic, and the cortical "column" we model would be 300 um a side. This is still _tiny_ in MEG-terms, but not totally outrageous.

The `lamtha`-parameter would have to be recast with a new physical identity, but in terms of implementation it would just amount to a scaling (from units of 1 um to 30 um). 

I don't think our simulated LFP-values will be physically meaningful until this matter has been resolved (even though qualitatively they might be reasonable). Certainly not for shanks inside the cortical column.

@stephanie-r-jones @jasmainak @ntolley @rythorpe

cjayb:
As discussed in #342, the current implementation has an in-plane cell separation of 1 um (soma-to-soma in x and y). This PR is intended to flesh out an API to modify that. Note that this won't resolve #342, which is of a more fundamental character.

- [x] add parameter to `Network` (and `default_network`)
- [x] modify cell creation and `lamtha` definition (also for drives!)
- [x] agree on API
- [x] tests (check sections created with desired distance)
- [x] make sure docstring of `Network` is informative and correct

Follow-up to #329


cjayb:
Mess with `plot_record...`-example in f39c7ca to see effects of distance. Note that when cells really close, the precise location of the electrodes matters more.

hnn_core/network.py
```diff
@@ -734,9 +750,11 @@ def _attach_drive(self, name, drive, weights_ampa, weights_nmda, location,
         drive['name'] = name  # for easier for-looping later
         drive['target_types'] = target_populations  # for _connect_celltypes
 
+        scaled_space_constant = space_constant * self._inplane_distance
```
jasmainak, line 753: I think this could be surprising to the user who provided a certain `lamtha` through the drives API and ultimately it was scaled and a different one was used.

hnn_core/network.py
```diff
@@ -52,20 +55,20 @@ def _create_cell_coords(n_pyr_x, n_pyr_y, zdiff=1307.4):
...
     pos_dict['L5_pyramidal'] = [
         pos for pos in it.product(xxrange, yyrange, [0])]
     pos_dict['L2_pyramidal'] = [
         pos for pos in it.product(xxrange, yyrange, [zdiff])]
 
     # BASKET CELLS
-    xzero = np.arange(0, n_pyr_x, 3)
-    xone = np.arange(1, n_pyr_x, 3)
+    xzero = np.arange(0, n_pyr_x, 3) * inplane_distance
+    xone = np.arange(1, n_pyr_x, 3) * inplane_distance
```
jasmainak, line 68: need a couple of simple tests to check that the inplane distance is indeed what's claimed

hnn_core/network.py
cjayb: Very good point!

cjayb:
The reference paper specifies the space constant in terms of the 'normalised' values (distance==1). I think this is a strong reason to keep the current implementation, where `lamtha` is essentially in units of: "steps on the grid". This would also keep legacy param files readable.

Now that we're at it perhaps it's time to get rid of the hard-coded value of L2-L5 soma distance (1307.4 um)?

```python
net = Network(params, inplane_distance=20., layer25_distance=1307.4)
```

jasmainak:
> lamtha is essentially in units of: "steps on the grid". 

okay understood, could we update the documentation of `space_constant` to reflect this?

> Now that we're at it perhaps it's time to get rid of the hard-coded value of L2-L5 soma distance (1307.4 um)?

what would you replace it with? or do you want to force users to specify it? I think a 1 mm distance is not too off, no?

hnn_core/network.py
```diff
@@ -104,6 +108,10 @@ def default_network(params, add_drives_from_params=False):
         If True, add drives as defined in the params-dict. NB this is mainly
         for backward-compatibility with HNN GUI, and will be deprecated in a
         future release. Default: False
+    inplane_distance : float
+        The in plane-distance (in um) between pyramidal cell somas in the
```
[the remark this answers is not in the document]
cjayb: That's an excellent question! As it is, they are placed _inside_ every third pyramidal cell soma... How's that for biological realism :) 

In our model, basket cells are 20 um-diameter, 39 um tall cylinders. That's a volume larger than the L2 pyramidals. I'm not sure why they need to be that big? They contain the `hh2` mechanism to produce sodium-spikes, which aren't going to contribute much to the LFP due to their biphasic nature.

cjayb:
> okay understood, could we update the documentation of space_constant to reflect this?

Like this be1e999f ?

> what would you replace it with? or do you want to force users to specify it? I think a 1 mm distance is not too off, no?

Oh I'd just make it a default value for the parameter in `Network`. Mostly just to document it, see a54268ed

hnn_core/network.py
```diff
@@ -500,9 +521,11 @@ def add_evoked_drive(self, name, *, mu, sigma, numspikes,
             Synaptic delay (in ms) at the column origin, dispersed laterally as
             a function of the space_constant. If float, applies to all target
             cell types. Use dict to create delay->cell mapping.
-        space_constant : float
+        space_constant : int | float
```
jasmainak, line 524: I would just keep it `float` ... don't give the users too many ideas :)

hnn_core/network.py
```diff
@@ -500,9 +521,11 @@ def add_evoked_drive(self, name, *, mu, sigma, numspikes,
...
             Describes lateral dispersion (from column origin) of synaptic
-            weights and delays within the simulated column
+            weights and delays within the simulated column. The constant is
+            measured in integer steps on the regular grid of the network
```
jasmainak, line 527:
```suggestion
            measured in units of inplane_distance. 
```

cjayb:
I moved basket cells out of the pyramidal soma plane. **NB Travis will likely fail because I haven't adjusted `lamtha`, but look at the CC rendering for this discussion**

There are tiny differences between the potentials in 
40c00fd (CC-rendering [here](https://229-265240816-gh.circle-artifacts.com/0/html/auto_examples/plot_record_extracellular_potentials.html#sphx-glr-auto-examples-plot-record-extracellular-potentials-py)) and bfcfe7a

The fact that basket cells appear not to contribute to the LFP makes sense: they only have the `hh2` mechanism, so they don't contribute with slower Ca-dependent currents. Based on this, I propose to change this

https://github.com/jonescompneurolab/hnn-core/blob/9b95ae83918cf91352f7b342e81fd19e22ffdcc3/hnn_core/extracellular.py#L476

back to `include_celltypes='Pyr'`, WDYT?

I still don't know what the 'correct' way to place the basket cells is though, but it seems to make very little difference.

jasmainak:
> NB Travis will likely fail because I haven't adjusted lamtha, but look at the CC rendering for this discussion

can it actually be adjusted to give the same result as before?

> The fact that basket cells appear not to contribute to the LFP makes sense: they only have the hh2 mechanism, so they don't contribute with slower Ca-dependent currents.

I'm lost :) could you point me to some resource? I tried searching for Sarah's thesis but couldn't find it

> back to include_celltypes='Pyr', WDYT?

yes, sounds reasonable! I still don't have a good sense of how the "validation" would proceed, so hard to comment further.

> I still don't know what the 'correct' way to place the basket cells is though,

looking at the code, I get the sense that the intention was to "spread the cells evenly" amongst the Pyramidal cells but it's not what has been achieved. For e.g., connectivity looks like this:

image

if you do:

```py
plot_cell_connectivity(net_erp, 28, 205)
```

in connectivity example. which is not exactly symmetric around the source cell

jasmainak:
> back to include_celltypes='Pyr'

between just to anticipate a little bit, do you think we should provide a list and it should be one of the types in `net.cell_types` ? Then we can add cell populations in the future which are "far away" and only impact the dynamics but not the measured LFP

jasmainak:
@cjayb I was playing around with this and I set `inplane_distance=60` hoping to see smaller LFPs but something really funky happened:

image

hnn_core/network.py
```diff
@@ -346,13 +368,14 @@ class Network(object):
...
+    `net = default_network(params)` is the recommended path for creating a
     network. Instantiating the network as `net = Network(params)` will
-    produce a network with no cell to cell connections. As such,
+    produce a network with no cell-to-cell connections. As such,
     connectivity information contained in `params` will be ignored.
     """
 
     def __init__(self, params, add_drives_from_params=False,
+                 inplane_distance=1., layer_separation=1307.4,
                  legacy_mode=True):
```
rythorpe, line 379: What if instead of creating an `inplane_distance` argument, we create a `cell_grid` argument that also exposes the number of cell replications in the network? The idea here would be to limit the arbitrary hard-coded assumptions (e.g., why are there only 100 L5 pyramidal cells in a 10-by-10 grid?) in a principled manner. For example, the `cell_grid=` argument might look like:

```python
grid_cell = {'L2_basket': [5, 7, 1],
             'L2_pyramidal': [10, 10, 1],
             'L5_basket': [10, 10, 1],
             'L5_pyramidal': [10, 10, 1]}
```
where the list assigned to each cell type has the form `[n_xdim, n_ydim, inplane_distance]`.

If this feels like over-engineering for this PR, we could reduce the `cell_grid` argument to a 3-element tuple (same structure as above) that only applies to pyramidal cells.

hnn_core/network.py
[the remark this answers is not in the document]
rythorpe: See my [comment below](https://github.com/jonescompneurolab/hnn-core/pull/370/files#r662490826).

hnn_core/network.py
cjayb: There's still the matter of how the basket cells are distributed (as noted by @jasmainak above). I'm not disagreeing with the objective to remove hard-coded assumptions, but when we do, we'll need quite a complicated API to support arbitrary arrangements. I'd say leave it for a later PR?

cjayb:
> @cjayb I was playing around with this and I set `inplane_distance=60` hoping to see smaller LFPs but something really funky happened:

That's very weird... I don't get the same:

LFPd60

Where did you place the electrode? Could you try again with the latest commit (after rebase)?

jasmainak:
@cjayb looks like you need to rebase this PR. I don't have the strange plot anymore 🥳

hnn_core/network.py
```diff
@@ -52,31 +55,31 @@ def _create_cell_coords(n_pyr_x, n_pyr_y, zdiff=1307.4):
...
     coords = [pos for pos in it.product(
         xzero, yeven)] + [pos for pos in it.product(xone, yodd)]
     coords_sorted = sorted(coords, key=lambda pos: pos[1])
     # append the z value for position for L2 and L5
     # print(len(coords_sorted))
 
-    pos_dict['L5_basket'] = [pos_xy + (0,) for
+    pos_dict['L5_basket'] = [(pos_xy[0], pos_xy[1], zdiff / 2) for
+                             pos_xy in coords_sorted]
+    pos_dict['L2_basket'] = [(pos_xy[0], pos_xy[1], 1.3 * zdiff) for
```
cjayb, line 81: Note that the 1.3 here is completely _ad hoc_, as is placing the L5 baskets halfway between the pyramidal layers. I'd like to hear @stephanie-r-jones take on where we should "place" the basket cells? Or more to the point: what sort of documentation is appropriate to describe the biological function modelled by them? Note that the position in Z (depth) has no impact on synaptic transmission, as 'lamtha' is purely in-plane.

cjayb:
> @cjayb looks like you need to rebase this PR. I don't have the strange plot anymore 🥳

Great! Rebase wasn't too hard. With the new `network_models`, we should have a discussion on the API we want here (and how to implement it in a way that's easy to maintain when new models are added). Then there's the broader question on what might be a "realistic" distance to set as the default. And what to do about basket cells (locations, contribution to LFP).

I'll have more time in the coming weeks to work on this.

jasmainak:
@cjayb a more pressing concern. I get segmentation faults when I use the code:

```py

import os.path as op

# We will use the default network with three evoked drives; see
# :ref:`evoked example <sphx_glr_auto_examples_plot_simulate_evoked.py>` for
# details. We'll go ahead and use the drive features defined in the parameter
# file.

import hnn_core
from hnn_core import read_params, default_network, simulate_dipole

voltages = dict()
for inp_dist in [30, 40, 50, 60]:
    print(f'Inplane distance of {inp_dist}')
    net = default_network(params, add_drives_from_params=True,
                          inplane_distance=inp_dist)

    depths = list(range(-325, 2150, 25))
    electrode_pos = [(135, 135, dep) for dep in depths]
    net.add_electrode_array('shank1', electrode_pos)

    dpl = simulate_dipole(net)
    trial_idx = 0
    voltages[inp_dist] = net.rec_arrays['shank1'][trial_idx].voltages.copy()
```

are you able to reproduce? It happens after the first iteration in the loop for me ...

cjayb:
I can reproduce the segfault in `master` :( Note that `default_network` no longer exists, so I replaced it with `jones_2009_model`.

I've traced the problem to the line `self._nrn_imem_ptrvec.gather(self._nrn_imem_vec)` in `extracellular.py`. On the first ODE iteration in the second loop, the acquisition of current values of `sec(seg.x)._ref_i_membrane_` (initialised on line 489) seems to lead to some sort of memory error. Maybe `h.allsec()` returns bogus sections?  I'd imagine crashes earlier, though...

jasmainak:
I know we talked about this earlier, but have you considered explicitly deleting any references to neuron objects after the simulation. You'd have to update this function:

https://github.com/jonescompneurolab/hnn-core/blob/78ce7cf73607135475a19718c79fa41304aaff80/hnn_core/network_builder.py#L539

I'm not sure we can rely on garbage collector ...

cjayb:
Went digging in NetPyNe and elsewhere, decided to give this `ptr_update_callback` a shot... and it seems to work (#401). Let's see what the CIs say before I rebase, tests pass locally.

hnn_core/network.py
```diff
@@ -36,6 +37,8 @@ def _create_cell_coords(n_pyr_x, n_pyr_y, zdiff=1307.4):
         Expressed as a positive DEPTH of L2 relative to L5, where L5 is defined
         to lie at z==0. Interlaminar weight/delay calculations (lamtha) are not
         affected.
+    inplane_distance : float
+        The grid spacing in um
```
ntolley, line 41: Perhaps add an entry to `Notes` that the basket cells don't lie in a true grid?

hnn_core/network.py
```diff
@@ -162,6 +165,14 @@ class Network(object):
         If True, add drives as defined in the params-dict. NB this is mainly
         for backward-compatibility with HNN GUI, and will be deprecated in a
         future release. Default: False
+    inplane_distance : float
```
jasmainak, line 168: would it make more sense to have a method called `set_cell_positions` rather than bake it into the `__init__` ? That way, say we read a network and want to change the in-plane separation later, we can do that.

cjayb: Hmm... I think the positions _have_ to be set in `init`, right? So would the method really help?

cjayb: > would it make more sense to have a method called `set_cell_positions`

I don't think so. Semantically, changing cell positions is equivalent to creating a new net, so I think the API should be to instantiate a new `net` with the spacing wanted. Note that in the future there may be other topologies with more parameters than just in-plane distance.

Pragmatically, `_create_cell_coords` in `init` returns the positions of the cells. These are used (still in `init`) by  `_add_cell_type` to set `net.pos_dict`.

Maybe I don't understand the API you're envisioning @jasmainak ?

jasmainak: I guess I'm wondering how this would fit in with https://github.com/jonescompneurolab/hnn-core/issues/402

if you get an already instantiated network from the reader function, how do you modify the cell spacing afterwards?

cjayb:
> if you get an already instantiated network from the reader function, how do you modify the cell spacing afterwards?

The latest commit (d273f68) introduces `net._reset_cell_positions()`, which trivially deals with `N_x` and `N_y`, too (#402). I haven't included them here because I want to hear your thoughts on this API first. Check out the new test in `test_network` and the example.

Note that #401 botched up the actual calculation of LFP, but I'm pushed d273f68 in order to keep the API discussion going here...

hnn_core/network.py
```diff
@@ -372,13 +388,15 @@ def __init__(self, params, add_drives_from_params=False,
...
-        pos = _create_cell_coords(n_pyr_x=self._params['N_pyr_x'],
-                                  n_pyr_y=self._params['N_pyr_y'],
-                                  zdiff=1307.4)
-        self.pos_dict['origin'] = pos['origin']
+
+        self._inplane_distance = inplane_distance
+        self._layer_separation = layer_separation
+        self._N_pyr_x = self._params['N_pyr_x']
+        self._N_pyr_y = self._params['N_pyr_y']
+        self._reset_cell_positions()
```
jasmainak, line 396:
```suggestion
        self._set_cell_positions()
```

hnn_core/network.py
```diff
@@ -399,6 +417,51 @@ def __repr__(self):
                  len(self.pos_dict['L5_basket'])))
         return '<%s | %s>' % (class_name, s)
 
+    def _reset_cell_positions(self):
```
jasmainak, line 420: but wouldn't you need to adjust the `lamtha` too ?

hnn_core/network.py
```diff
@@ -399,6 +417,51 @@ def __repr__(self):
...
+    def layer_separation(self, new_sep):
+        if len(self.external_drives) > 0:
+            raise RuntimeError('Layer separation between pyramidal cells can '
+                               'only be adjusted before external drives are '
+                               f'attached; have {len(self.external_drives)} '
+                               'drives.')
+        _validate_type(new_sep, (float, int), 'new separation')
+        if not new_sep > 0.:
+            raise ValueError('Layer separation must be positive definite, '
+                             f'got: {new_sep}')
```
jasmainak, line 460: I would just give this as an argument to `set_cell_position` and set it inside there

hnn_core/network.py
```diff
@@ -399,6 +417,51 @@ def __repr__(self):
...
+                               'only be adjusted before external drives are '
+                               f'attached; have {len(self.external_drives)} '
+                               'drives.')
+        _validate_type(new_dist, (float, int), 'new distance')
+        if not new_dist > 0.:
+            raise ValueError('In-plane distance must be positive definite, '
+                             f'got: {new_dist}')
+
+        self._inplane_distance = new_dist
+        self._reset_cell_positions()
```
jasmainak, line 444: KISS principle. Just have a `set_cell_position` with these arguments and be done with it

jasmainak:
rebase time? :)

hnn_core/network.py
cjayb: indeed, 4517a47 now adjusts `nc_dict['lamtha']` for all connections.

cjayb:
I found a way to keep `network_models` independent of `inplane_distance`. Right now this is implemented using a `@setter`, which isn't KISS, but makes for quite a nice API?

```python
net = any_current_or_future_net(params, ...)
net.inplane_distance = 30.
add_default_drives(net)
```

NB currently drives are assumed to be defined in the same 'unitless' scale as the net is (unit cell distance). This could be a little confusing: maybe after setting the `inplane_distance`, the user assumes `space_constant` to also be in these units?

We should talk about this PR at an upcoming meeting...

hnn_core/network.py
```diff
@@ -400,6 +420,57 @@ def __repr__(self):
...
+        _validate_type(new_dist, (float, int), 'new distance')
+        if not new_dist > 0.:
+            raise ValueError('In-plane distance must be positive definite, '
+                             f'got: {new_dist}')
+
+        # Adjust the lamtha of all cells; note that this would have no effect
+        # on external drives, since the event times there are calculated when
+        # the drives are attached.
+        for conn in self.connectivity:
+            conn['nc_dict']['lamtha'] *= new_dist / self._inplane_distance
```
jasmainak, line 450: feels a bit like a hack. If a user inspected the `net.connectivity` they would be surprised that `lamtha` changed. And also it doesn't work for drives. I think we should pass `inplane_distance` here so it's applied uniformly and doesn't change `lamtha`:

https://github.com/jonescompneurolab/hnn-core/blob/0d40aae54abc313277089b750654637a4c5826d9/hnn_core/network_builder.py#L501

cjayb:
Before I rebase, take a look at 0b30d33 . Now `inplane_distance` is applied in `parconnect_from_src`. For some reason, I still need to set the distance before attaching drives. I need to dig deeper into why this is the case (it shouldn't be, since I've realised that only event times are pre-computed, and these are independent of `lamtha`).

But how does the API feel? All `lamtha`s are kept unitless, each `net` is initialised with unit distances, but the user can set the attribute before simulating.

And what about the specific choices for basket cell depths? The are currently totally _ad hoc_, since the `hh2` mechanisms produce neither 'dipole' current nor LFP.

hnn_core/network.py
```diff
@@ -400,6 +420,51 @@ def __repr__(self):
...
+    def inplane_distance(self):
+        return self._inplane_distance
+
+    @inplane_distance.setter
+    def inplane_distance(self, new_dist):
+        if len(self.external_drives) > 0:
+            raise RuntimeError('In-plane distance between network cells can '
+                               'only be adjusted before external drives are '
+                               f'attached; have {len(self.external_drives)} '
+                               'drives.')
```
jasmainak, line 440: is this still needed now that we moved `inplane_distance` to `parconnect_from_src`?

hnn_core/tests/test_network.py
```diff
@@ -82,6 +82,48 @@ def test_network_models():
...
+    for cell_name, coords in pos_dict.items():
+        if cell_name == 'origin':
+            continue
+        for xyz_orig, xyz_new in zip(coords, net.pos_dict[cell_name]):
+            for orig, new in zip(xyz_orig, xyz_new):
+                assert np.isclose(orig, new)
+
+    net.inplane_distance = 2.
+    # check that in-plane distance has doubled
+    assert np.isclose(2. * pos_dict['L5_pyramidal'][1][1],
```
jasmainak, line 103: but this is not inplane distance. It's the absolute position, no?

jasmainak:
> But how does the API feel? All lamthas are kept unitless, each net is initialised with unit distances, but the user can set the attribute before simulating.

Not totally obvious to me what's the right API. I guess I still prefer the `set_cell_positions` method. That way you avoid an intermediate function (`_reset_cell_positions`) that does the heavy-lifting. And you can set all the variables using a single function:

```py
def set_cell_positions(inplane_distance=1., layer_separation=1307, N_Pyr_x=10, N_Pyr_y=10)
```

probably a bias from MNE where `set_xxx` function is more common than `setters`. Think @ntolley and @rythorpe should pitch in as they are the real users.

jasmainak:
> And what about the specific choices for basket cell depths?

Maybe don't allow users to set the layer separation? Specially if there is no clear use case yet ... The more features we add, the more work we create for ourselves.

hnn_core/network.py
cjayb: It shouldn't be, but I haven't yet found the bug...

cjayb:
So you prefer this?

```python
net.set_cell_positions(...)
```

I can live with that. 

> Maybe don't allow users to set the layer separation? Specially if there is no clear use case yet

In my mind the layer separation is a part of the discussion in #342 As soon as anyone wants to compare dipole currents to sink-source distributions in cortex (and calculate current dipole moments), this is going to pop up. I think it's worth incorporating and documenting here.

I was previously miffed by the fact that with 1 um in-plane spacing, our pyramidal cells would overlap due to their oversized dendrites. This also makes the concept of "extracellular potentials" dubious: any electrode inside the net would be inside multiple cells simultaneously. So here I'm moving the cells laterally to make room for some extracellular space. But the interneurons will still be located _inside_ the pyramidal cells, whatever in-plane distance we choose. I imagine this will start to stir up confusion once people begin visualising the network and inserting electrode arrays. I just don't know what the solution should be.

rythorpe:
Maybe we can discuss the API in our HNN meeting tomorrow? I suspect it will be easier to hash out in real-time.

jasmainak:
Okay, fair enough about exposing `layer_separation`. If we can change it to `set_cell_positions` and rebase, the PR is good to go on my end

hnn_core/tests/test_network.py
cjayb: hmm, not sure I understand. I guess it's more accurately a 'step size'?

cjayb:
> Maybe we can discuss the API in our HNN meeting tomorrow?

I'll do my best to join today. In terms of code I've decided to submit this for MRG, so go ahead and review away!

examples/howto/plot_record_extracellular_potentials.py
```diff
@@ -23,18 +23,31 @@
...
+
+###############################################################################
+# The default network model defined in Jones et al. (2009) [1]_ consists of a
+# square grid of pyramidal cells. The in-plane distance between pyramidal cell
+# somas on the grid can be set by the user, which will have an influence on the
+# extracellular potentials (but not on the calculated net intracellular dipole
+# moment). In this example, we'll simulate a network of model cells spaced
+# 30 um apart. To drive the network dynamics, we'll use three evoked 'ERP'
+# drives; see
+# :ref:`evoked example <sphx_glr_auto_examples_plot_simulate_evoked.py>`
```
jasmainak, line 44: I think this will not render as a link, you might have to change this because the folder hierarchy changed and this is now under tutorials

hnn_core/network.py
```diff
@@ -400,6 +406,61 @@ def __repr__(self):
                  len(self.pos_dict['L5_basket'])))
         return '<%s | %s>' % (class_name, s)
 
+    def set_cell_positions(self, *, n_pyr_x=10, n_pyr_y=10,
+                           inplane_distance=1., layer_separation=1307.4):
+        """Set relative positions of cells arranged in a square grid
+
+        All ``hnn_core``-models are organised in a square grid of
+        ``n_pyr_x X n_pyr_y` pyramidal cells. Basket cells are included at a
```
jasmainak, line 414: missing one backtick. In general, I suggest being conservative with fancy formatting because many users tend to check the docstrings on ipython where it is not rendered in HTML

hnn_core/tests/test_network.py
```diff
@@ -83,6 +83,48 @@ def test_network_models():
...
+
+    net = jones_2009_model()
+    assert np.isclose(net._inplane_distance, 1.)
+    assert np.isclose(net._layer_separation, 1307.4)
+    pos_dict_nodrives = deepcopy(net.pos_dict)
+
+    net.set_cell_positions(inplane_distance=2.)
+    # check that in-plane distance has doubled from the default 1.0
+    assert np.isclose(2. * pos_dict_nodrives['L5_pyramidal'][1][1],
+                      net.pos_dict['L5_pyramidal'][1][1])
```
jasmainak, line 97: sorry to ask again. Let me try to rephrase. Isn't `pos_dict` the position not the in-plane distance? For checking the inplane distance you will have to actually do `np.diff` or some such thing

jasmainak:
Some minor comments but PR looks in great shape to me! @ntolley and @rythorpe if you guys want to take a look, feel free

jasmainak:
Don't forget to update `whats_new.rst` !

hnn_core/tests/test_network.py
cjayb: I see the confusion now, the test is formulated weirdly. I've made a better one using `diff`

hnn_core/tests/test_network.py
```diff
@@ -83,6 +83,49 @@ def test_network_models():
...
+    net.set_cell_positions(layer_separation=1000.)
+    # check that layer separation has changed (L5 is zero)
+    assert np.isclose(net.pos_dict['L2_pyramidal'][0][2], 1000.)
+
+    with pytest.raises(ValueError,
+                       match='Number of pyramidal cells in each direction'):
+        net.set_cell_positions(n_pyr_y=0)
+    with pytest.raises(ValueError,
+                       match='Number of pyramidal cells in each direction'):
+        net.set_cell_positions(n_pyr_y=0)
```
jasmainak, line 109: isn't this same as the test in the last line?

hnn_core/tests/test_network.py
```diff
@@ -83,6 +83,49 @@ def test_network_models():
...
+    with pytest.raises(ValueError,
+                       match='In-plane distance must be positive'):
+        net.set_cell_positions(inplane_distance=0.)
+    with pytest.raises(ValueError,
+                       match='Layer separation must be positive'):
+        net.set_cell_positions(layer_separation=0.)
+
+    # check that the origin of the drive cells matches the new 'origin'
```
jasmainak, line 117: why would the origin change when `set_cell_positions` is called?

hnn_core/network.py
```diff
@@ -327,11 +332,12 @@ class Network(object):
     ----
     `net = jones_2009_model(params)` is the reccomended path for creating a
     network. Instantiating the network as `net = Network(params)` will
-    produce a network with no cell to cell connections. As such,
+    produce a network with no cell-to-cell connections. As such,
     connectivity information contained in `params` will be ignored.
     """
 
     def __init__(self, params, add_drives_from_params=False,
+                 inplane_distance=1., layer_separation=1307.4,
```
jasmainak, line 340: these two parameters are not documented in public class

examples/howto/plot_record_extracellular_potentials.py
```diff
@@ -23,18 +23,29 @@
...
+###############################################################################
+# The default network model defined in Jones et al. (2009) [1]_ consists of a
+# square grid of pyramidal cells. The in-plane distance between pyramidal cell
+# somas on the grid can be set by the user, which will have an influence on the
+# extracellular potentials (but not on the calculated net intracellular dipole
+# moment). In this example, we'll simulate a network of model cells spaced
+# 30 um apart. To drive the network dynamics, we'll use three evoked 'ERP'
+# drives; see the event-related potential (ERP) example for details.
+
+net = jones_2009_model(params)
```
jasmainak, line 45:
```suggestion
net = jones_2009_model()
```

like in #426

jasmainak (approved):
Sorry I noticed a couple of last minute issues. I will let @rythorpe or @ntolley press the green button

hnn_core/network.py
```diff
@@ -373,13 +379,13 @@ def __init__(self, params, add_drives_from_params=False,
         # cell counts, real and artificial
         self.pos_dict = dict()
         self.cell_types = dict()
-        pos = _create_cell_coords(n_pyr_x=self._params['N_pyr_x'],
-                                  n_pyr_y=self._params['N_pyr_y'],
-                                  zdiff=1307.4)
-        self.pos_dict['origin'] = pos['origin']
+
+        self._N_pyr_x = self._params['N_pyr_x']
+        self._N_pyr_y = self._params['N_pyr_y']
```
rythorpe, line 384: Is there a particular reason why these are attributes of `Network`?

hnn_core/network.py
```diff
@@ -484,8 +545,14 @@ def add_evoked_drive(self, name, *, mu, sigma, numspikes, location,
...
             cell types. Use dict to create delay->cell mapping.
         space_constant : float
-            Describes lateral dispersion (from column origin) of synaptic
-            weights and delays within the simulated column
+            Describes lateral dispersion (from the column origin) of synaptic
+            weights and delays within the simulated column. The constant is
+            measured in the units of ``inplane_distance`` of
+            :class:`~hnn_core.Network`. For example, for ``space_constant=3``,
+            the weights and delays are modulated by the factor
+            ``exp(-(x / (3 * inplane_distance)) ** 2)``, where ``x`` is the
```
rythorpe, line 553:
```suggestion
            the weights are modulated by the factor
            ``exp(-(x / (3 * inplane_distance)) ** 2)``, where ``x`` is the
```
delays are modulated by a factor that is the inverse of that value

rythorpe:
Aside from the few remaining questions and nitpicks, this looks great @cjayb!

hnn_core/network.py
jasmainak: also, not sure if it's updated after calling `set_cell_positions` ...

hnn_core/tests/test_network.py
cjayb:
```python
    # As the network dimensions increase, so does the center-of-mass of the
    # grid points, which is where all hnn drives should be located. The lamtha-
    # dependent weights and delays of the drives are calculated with respect to
    # this origin.
```

jasmainak: humm ... I guess this means the neurons are not symmetrically distributed around origin. Side effect of having the origin defined based on the positions of the neurons rather than having the positions based on the origin. I find this a bit weird but can live with it for now ... let's have users complain about it

hnn_core/network.py
cjayb: These were removed, more discussion is needed.

hnn_core/network.py
cjayb: Good questions, which revealed a bug (calling `set_cell_positions` would inadvertently set all non-specified parameters to their default values). This is fixed in ff7fb202

hnn_core/network.py
```diff
@@ -394,12 +409,76 @@ def __init__(self, params, add_drives_from_params=False,
...
         class_name = self.__class__.__name__
         s = ("%d x %d Pyramidal cells (L2, L5)"
-             % (self._params['N_pyr_x'], self._params['N_pyr_y']))
+             % (self._N_pyr_x, self._N_pyr_y))
         s += ("\n%d L2 basket cells\n%d L5 basket cells"
               % (len(self.pos_dict['L2_basket']),
                  len(self.pos_dict['L5_basket'])))
         return '<%s | %s>' % (class_name, s)
 
+    def set_cell_positions(self, *, n_pyr_x=None, n_pyr_y=None,
```
jasmainak, line 418: thinking a bit more about this don't think `n_pyr_x` or `n_pyr_y` should be exposed here. Because `n_pyr_x` and `n_pyr_y` also dictate the number of cells. Once you have created cells using [`_add_cell_type`](https://github.com/jonescompneurolab/hnn-core/blob/d3e51aee3f4a2aea63b411448f1aea8fa80ae44f/hnn_core/network.py#L382-L383), it should not be possible to change the number of cells. I think we can expose these arguments as part of `add_cell_type` when we make it public. The `pos_dict` would get created inside that method. Then the `repr` of the `Network` class will tell you how many cells per layer per cell type and if users are more curious they should inspect `net.pos_dict` or the visualization.

A second issue is that this will lead to inconsistencies in the API. For e.g., why would you want to control number of pyramidal cells but not the number basket cells?

hnn_core/network.py
```diff
@@ -268,6 +273,12 @@ def pick_connection(net, src_gids=None, target_gids=None,
 class Network(object):
     """The Network class.
 
+    All hnn_core-models are organised in a square grid of pyramidal cells.
+    Basket cells are included at a ratio of 1:3. The separation between L2 and
+    L5 layers and the distance between grid points (i.e., pyramidal cell somas)
+    and the number of cells in the x- and y-directions are adjusted using the
+    set_cell_positions-method.
```
jasmainak, line 280: I would document this in `jones_2009_model` ...

cjayb: OK to `jones_2009_model`, but `Network` will need major changes to be generic. For example, `self._params['N_pyr_x']` (and `_y`) is used, which assumes a grid. Another PR!

jasmainak: agree, but one PR at a time :) Little strokes fell great oaks ...

cjayb:
I've now removed the ability to change `self._N_pyr_x`, only in-plane and layer separation can be changed. The two latter ones are hard-coded into `Network.__init__`, as I couldn't figure out a more appropriate place for them! I considered re-inserting them as parameters of `__init__`, but since we discourage the use of `Network`, it didn't seem right.

Perhaps properties of `Network`, like this:

```python
net = jones_2009_model()
print(net.inplane_distance)  # 1.
print(net.layer_separation)  # 1307.4
net.inplane_distance = 30.  # ERROR, no @setter
net.set_cell_positions(inplane_distance=30.)
```

Should I make `net.inplane_distance` and `net.layer_separation` public properties?

jasmainak:
Oops, I just saw your comment @cjayb 

I need to run to a meeting now, hopefully I didn't do anything catastrophic :)

cjayb:
> Oops, I just saw your comment @cjayb
> 
> I need to run to a meeting now, hopefully I didn't do anything catastrophic :)

No worries :) I did change to MRG! Take a look at my final proposal, it's just cosmetics. A follow-up will be good

jasmainak:
Okay just read your comment quickly. I was wondering the same. I would say let's keep it private unless we think it will need to be used in user code. You can put that info in `repr`

rythorpe:
Thanks @cjayb! :tada::tada: This PR was a slow burn given the sheer number of issues you encountered along the way but it's finally merged!

PR #475, opened 2022-03-22
[MRG] Add average dpl plot

chenghuzi:
To address #473 , we add an extra line plot when the `dpl` in `plot_dipole` is a list. All individual trial lines are colored gray while the average one is green. And if the dpl is a `Dipole` instance, the single trial will be green. 

Below is the current behavior:
image

Also, I changed the way to turn `dpl` into a list, as a function should not modify its parameters especially when the purpose is to draw something. Because the user may expect the `dpl` to stay the original type after multiple calls of the visualization function (for reasons like refining styles).

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -34,6 +34,9 @@ Bug
 - Subsets of trials can be indexed when using :func:`~hnn_core.viz.plot_spikes_raster`
   and :func:`~hnn_core.viz.plot_spikes_hist`, by `Nick Tolley`_ in :gh:`472`.
 
+- Add option to plot the averaged dipole in `~hnn_core.viz.plot_dipole` when `dpl` is a list of dipoles, 
+  by `Huzi Cheng`_ in :gh:`475`.
+
 API
 ~~~
 - Optimization of the evoked drives can be conducted on any :class:`~hnn_core.Network`
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -303,6 +303,7 @@ People who contributed to this release (in alphabetical order):
 .. _Christopher Bailey: https://github.com/cjayb
 .. _Carmen Kohl: https://github.com/kohl-carmen
 .. _Dylan Daniels: https://github.com/dylansdaniels
+.. _Huzi Cheng: https://github.com/chenghuzi
 .. _Kenneth Loi: https://github.com/kenloi
 .. _Mainak Jas: http://jasmainak.github.io/
 .. _Nick Tolley: https://github.com/ntolley
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -34,8 +34,8 @@ Bug
 - Subsets of trials can be indexed when using :func:`~hnn_core.viz.plot_spikes_raster`
   and :func:`~hnn_core.viz.plot_spikes_hist`, by `Nick Tolley`_ in :gh:`472`.
 
-- Add option to plot the averaged dipole in `~hnn_core.viz.plot_dipole` when `dpl` is a list of dipoles, 
-  by `Huzi Cheng`_ in :gh:`475`.
+- Add option to plot the averaged dipole in `~hnn_core.viz.plot_dipole` when `dpl`
+  is a list of dipoles, by `Huzi Cheng`_ in :gh:`475`.
 
 API
 ~~~
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -151,6 +151,8 @@ def average_dipoles(dpls):
         )
     avg_data = np.c_[avg_data].T
     avg_dpl = Dipole(dpls[0].times, avg_data)
+    # The averaged scale should equal all scals in the input dpl list.
+    avg_dpl.scale_applied = dpls[0].scale_applied
 
     # set nave to the number of trials averaged in this dipole
     avg_dpl.nave = len(dpls)
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -133,7 +133,10 @@ def average_dipoles(dpls):
         A new dipole object with each component of `dpl.data` representing the
         average over the same components in the input list
     """
+    scale_applied = dpls[0].scale_applied
     for dpl_idx, dpl in enumerate(dpls):
+        if dpl.scale_applied != scale_applied:
+            raise RuntimeError('All dipoles must be scaled equally!')
         if not isinstance(dpl, Dipole):
             raise ValueError(
                 f"All elements in the list should be instances of "
@@ -152,7 +155,7 @@ def average_dipoles(dpls):
     avg_data = np.c_[avg_data].T
     avg_dpl = Dipole(dpls[0].times, avg_data)
     # The averaged scale should equal all scals in the input dpl list.
-    avg_dpl.scale_applied = dpls[0].scale_applied
+    avg_dpl.scale_applied = scale_applied
 
     # set nave to the number of trials averaged in this dipole
     avg_dpl.nave = len(dpls)
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -392,7 +392,7 @@ class Dipole(object):
         return self
 
     def plot(self, tmin=None, tmax=None, layer='agg', decim=None, ax=None,
-             color=None, show=True):
+             color='k', show=True):
         """Simple layer-specific plot function.
 
         Parameters
@@ -408,7 +408,7 @@ class Dipole(object):
         ax : instance of matplotlib figure | None
             The matplotlib axis
         color : tuple of float
-            RGBA value to use for plotting (optional)
+            RGBA value to use for plotting. By default, 'k' (black)
         show : bool
             If True, show the figure
 
--- a/hnn_core/tests/test_dipole.py
+++ b/hnn_core/tests/test_dipole.py
@@ -49,6 +49,12 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
         assert_allclose(dipole_read.data[dpl_key],
                         dipole.data[dpl_key], rtol=0, atol=0.000051)
 
+    # dpls with different scale_applied should not be averaged.
+    with pytest.raises(RuntimeError, match="All dipoles must be scaled equally"
+                       "!"):
+        dipole_avg = average_dipoles([dipole, dipole_read])
+    # force the scale_applied to be identical across dpls to allow averaging.
+    dipole.scale_applied = dipole_read.scale_applied
     # average two identical dipole objects
     dipole_avg = average_dipoles([dipole, dipole_read])
     for dpl_key in dipole_avg.data.keys():
--- a/hnn_core/tests/test_dipole.py
+++ b/hnn_core/tests/test_dipole.py
@@ -50,8 +50,7 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
                         dipole.data[dpl_key], rtol=0, atol=0.000051)
 
     # dpls with different scale_applied should not be averaged.
-    with pytest.raises(RuntimeError, match="All dipoles must be scaled equally"
-                       "!"):
+    with pytest.raises(RuntimeError, match="All dipoles must be scaled equally"):
         dipole_avg = average_dipoles([dipole, dipole_read])
     # force the scale_applied to be identical across dpls to allow averaging.
     dipole.scale_applied = dipole_read.scale_applied
--- a/hnn_core/tests/test_dipole.py
+++ b/hnn_core/tests/test_dipole.py
@@ -50,7 +50,8 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
                         dipole.data[dpl_key], rtol=0, atol=0.000051)
 
     # dpls with different scale_applied should not be averaged.
-    with pytest.raises(RuntimeError, match="All dipoles must be scaled equally"):
+    with pytest.raises(RuntimeError,
+                       match="All dipoles must be scaled equally"):
         dipole_avg = average_dipoles([dipole, dipole_read])
     # force the scale_applied to be identical across dpls to allow averaging.
     dipole.scale_applied = dipole_read.scale_applied
--- a/hnn_core/viz.py
+++ b/hnn_core/viz.py
@@ -256,7 +256,6 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
     if isinstance(dpl, Dipole):
         dpl = [dpl]
         average = False
-        alpha = 1
     else:
         # add average plot when we have dpl>1 and `average` is specified.
         if average:
--- a/hnn_core/viz.py
+++ b/hnn_core/viz.py
@@ -209,8 +209,17 @@ def plot_extracellular(times, data, tmin=None, tmax=None, ax=None,
     return ax.get_figure()
 
 
-def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
-                color=None, show=True):
+def plot_dipole(dpl,
+                tmin=None,
+                tmax=None,
+                ax=None,
+                layer='agg',
+                decim=None,
+                color=None,
+                average=True,
+                individual_alpha=0.5,
+                linewidth=1.5,
+                show=True):
     """Simple layer-specific plot function.
 
     Parameters
@@ -233,6 +242,12 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
         ints can be provided. These are applied successively.
     color : tuple of float
         RGBA value to use for plotting (optional)
+    average : bool
+        If True, render the average across all dpls.
+    individual_alpha : float
+        The alpha value of individual dpls.
+    linewidth : float
+        The width of dpl lines.
     show : bool
         If True, show the figure
 
@@ -248,13 +263,16 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
         _, ax = plt.subplots(1, 1, constrained_layout=True)
 
     if isinstance(dpl, Dipole):
-        dpls = [dpl]
+        dpl = [dpl]
+        average = False
+        individual_alpha = 1
     else:
-        # add average plot only when we have  trials>1.
-        dpls = dpl + [average_dipoles(dpl)]
+        # add average plot when we have dpl>1 and `average` is specified.
+        if average:
+            dpl = dpl + [average_dipoles(dpl)]
 
-    scale_applied = dpls[0].scale_applied
-    for i, dpl_trial in enumerate(dpls):
+    scale_applied = dpl[0].scale_applied
+    for idx, dpl_trial in enumerate(dpl):
         if dpl_trial.scale_applied != scale_applied:
             raise RuntimeError('All dipoles must be scaled equally!')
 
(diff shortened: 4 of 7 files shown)


chenghuzi:
Some questions:

* Do we need to change the `alpha` of those individual trials?
* Should we increase the line width for the average dpl data like [this one](https://github.com/jonescompneurolab/hnn/blob/4bc3e4d51fcf8ae601191861d8b415205fbcaf7f/hnn/qt_dipole.py#L97 )?

hnn_core/viz.py
```diff
@@ -242,29 +242,36 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
...
     import matplotlib.pyplot as plt
-    from .dipole import Dipole
+    from .dipole import Dipole, average_dipoles
 
     if ax is None:
         _, ax = plt.subplots(1, 1, constrained_layout=True)
 
     if isinstance(dpl, Dipole):
-        dpl = [dpl]
+        dpls = [dpl]
```
ntolley, line 251: No need to rename to `dpls`, changing the `dpl` variable will only persist within the scope of this function.

If the user passes a `Dipole` object into the function, the original variable won't be turned into a `list()`.

ntolley:
This is great! I think the gray for the individual trials looks good, but perhaps a slightly thicker line for the average? I'm thinking 1.25 - 1.5x compared to the gray lines

jasmainak:
I would indeed add an alpha to the individual trials if there are multiple trials.

hnn_core/viz.py
jasmainak: I think the reason is the following:

"Also, I changed the way to turn dpl into a list, as a function should not modify its parameters especially when the purpose is to draw something. Because the user may expect the dpl to stay the original type after multiple calls of the visualization function (for reasons like refining styles)."

which makes sense to me

jasmainak:
@chenghuzi could you add/update some tests? Smoke tests are fine for visualization.

hnn_core/viz.py
```diff
@@ -242,29 +242,36 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
...
+    else:
+        # add average plot only when we have  trials>1.
+        dpls = dpl + [average_dipoles(dpl)]
 
-    scale_applied = dpl[0].scale_applied
-    for dpl_trial in dpl:
+    scale_applied = dpls[0].scale_applied
+    for i, dpl_trial in enumerate(dpls):
```
jasmainak, line 257: this is a nitpick, but we make a lot of those :) I would rename the variable `i` to be `idx` as short for "index". Single letter variable names are not recommended typically because they are a bit hard to Ctr+F (find) and replace.

hnn_core/viz.py
chenghuzi: I see, previously I was using the following form:


```Python
    if isinstance(dpl, Dipole):
        dpl = [dpl]
    else:
        # add average plot only when we have  trials>1.
        # dpl = dpl + [average_dipoles(dpl)]
        dpl.append(average_dipoles(dpl))
```
So this leads to 

image

as the `append` modified the `dpl` in an inplace manner. But `dpl = dpl + [average_dipoles(dpl)]` should work so I'll change it back

hnn_core/viz.py
chenghuzi: OK

chenghuzi:
> @chenghuzi could you add/update some tests? Smoke tests are fine for visualization.

Sure, will add them once the implementation looks good to us.

chenghuzi:
> looks good! also I'd call it `alpha` instead of `individual_alpha` and clarify in the docstring that it's applied only to individual trials.
> […](#)

Sure, fixed in [8aad9cb](https://github.com/jonescompneurolab/hnn-core/pull/475/commits/8aad9cb9a3ab06cc95713226b3f1a5f6635c70f6)

hnn_core/viz.py
```diff
@@ -209,8 +209,17 @@ def plot_extracellular(times, data, tmin=None, tmax=None, ax=None,
...
+                tmin=None,
+                tmax=None,
+                ax=None,
+                layer='agg',
+                decim=None,
+                color=None,
+                average=True,
+                alpha=0.5,
+                linewidth=1.5,
+                show=True):
```
jasmainak, line 222: try to refrain from stylistic changes in PRs and keep the diff minimal. If necessary, we should discuss and address in a separate PR

jasmainak:
tests remain, and after that you can update `whats_new.rst`. Getting close!

hnn_core/dipole.py
```diff
@@ -151,6 +151,8 @@ def average_dipoles(dpls):
         )
     avg_data = np.c_[avg_data].T
     avg_dpl = Dipole(dpls[0].times, avg_data)
+    # The averaged scale should equal all scals in the input dpl list.
+    avg_dpl.scale_applied = dpls[0].scale_applied
```
jasmainak, line 155: We should also throw a `ValueError` error if the `scale_applied` is not the same for all `dpls`

chenghuzi: agreed

chenghuzi: It seems doing so requires changes in `read_dipole` as it ignores the `scale_applied` and only read data from the npy file. That causes the https://github.com/jonescompneurolab/hnn-core/blob/52620b2f05d8ba6c1192a80cac64e2deb0591700/hnn_core/tests/test_dipole.py#L53

to fail the test. My suggestion is: for now we change the `scale_applied` manually for the above test to let it pass the test. But in the future we may include `scale_applied` in the reading/writing of dipole.

hnn_core/viz.py
```diff
@@ -242,29 +248,45 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
...
-    from .dipole import Dipole
+    from .dipole import Dipole, average_dipoles
 
     if ax is None:
         _, ax = plt.subplots(1, 1, constrained_layout=True)
 
     if isinstance(dpl, Dipole):
         dpl = [dpl]
+        average = False
+        alpha = 1
```
rythorpe, line 259: This all looks really good @chenghuzi! Sorry I'm late to the game. This is super minor, but I would avoid overwriting `alpha` here so that the user can still choose to set it's value if desired.

chenghuzi: Got it. just fixed that

rythorpe: Actually, what happens if the user sets the `color` parameter but only supplies a single `Dipole` object to `plot_dipole`? IMO, they should be able set the color of individual trials if `average=False`, or the color of the average dipole if `average=True`.

rythorpe: Once again, my only objective here is to avoid silently overwriting user-specified parameters to avoid confusion.

chenghuzi: Then the same parameter `color` has different meanings in that case, which I think may create even more confusion. In our current implementation, `color` is just for the averaged dpl line but not for individual dpls. As for the case of `average=False`, I don't think manually assigning different colors for different trials will be meaningful. Of course, a straightforward solution is to add another parameter like `individual_colors` as a list of color values.

jasmainak: what one might want to do is display dipoles from two different conditions in the same figure by passing the `ax` parameter:

```py
fig = dpl_cond1.plot(color='r')
dpl_cond2.plot(color='b', ax=fig.gca())
```

I think it's fine to fix the color of the averaged dipole and trust that the user does not select this color? Alternatively if you want more control do something like:

```py
color : float | dict
   If dict, keys can be 'average' and 'trial'
```

But I think we can build this in later if there is a need. For now, I'd say YAGNI. Just use color for setting the color of individual trials

hnn_core/viz.py
```diff
@@ -233,6 +233,12 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
         ints can be provided. These are applied successively.
     color : tuple of float
         RGBA value to use for plotting (optional)
+    average : bool
+        If True, render the average across all dpls.
+    alpha : float
+        The alpha value of individual dpls.
+    linewidth : float
+        The width of dpl lines.
```
jasmainak, line 241: let's not expose these. The more you expose parameters, the more you have to maintain.

jasmainak, line 241: let's not expose these. The more you expose parameters, the more you have to maintain.

hnn_core/tests/test_dipole.py
```diff
@@ -49,6 +49,12 @@ def test_dipole(tmpdir, run_hnn_core_fixture):
         assert_allclose(dipole_read.data[dpl_key],
                         dipole.data[dpl_key], rtol=0, atol=0.000051)
 
+    # dpls with different scale_applied should not be averaged.
+    with pytest.raises(RuntimeError, match="All dipoles must be scaled equally"
+                       "!"):
```
jasmainak, line 54:
```suggestion
    with pytest.raises(RuntimeError, match="All dipoles must be scaled equally"):
```

just a cosmetic suggestion. Hopefully it fits in 80 chars :)

jasmainak:
I don't quite get why circleCI is not working on your pull request @chenghuzi. I tried to make a dummy pull request on #476 and that did run circleCI:

image

Does this also happen in your other pull request?

hnn_core/viz.py
rythorpe: ^that works. There have also been times when I’ve needed to compute the trial-average dipole outside `plot_dipole` and then wished to plot it as a certain color within a preexisting matplotlib `axes` instance.

chenghuzi: A little bit confused. Are we going to stay with the current implementation? If we change to use `color` for individual trials, then a default color must be set as the matplotlib will assign different colors for each line if `color=None`. And to avoid conflicts, a default color for averaged line should also be set. And this brings us back to the `color : float | dict` solution I guess

chenghuzi:
> I don't quite get why circleCI is not working on your pull request @chenghuzi. I tried to make a dummy pull request on #476 and that did run circleCI:
> 
> image
> 
> Does this also happen in your other pull request? Maybe try pushing a new commit to see if the problem goes away?

I noticed that sometimes `neuron.yale.edu` is not accessible and probably that might be the reason it fails tests

doc/whats_new.rst
```diff
@@ -34,6 +34,9 @@ Bug
 - Subsets of trials can be indexed when using :func:`~hnn_core.viz.plot_spikes_raster`
   and :func:`~hnn_core.viz.plot_spikes_hist`, by `Nick Tolley`_ in :gh:`472`.
 
+- Add option to plot the averaged dipole in `~hnn_core.viz.plot_dipole` when `dpl` is a list of dipoles, 
+  by `Huzi Cheng`_ in :gh:`475`.
```
jasmainak, line 38: you will need to put your name at the bottom with a URL/homepage so this renders as a hyperlink. You can check the rendering by doing:

```sh
$ cd doc/
$ make html-noplot
```

hnn_core/viz.py
```diff
@@ -242,29 +244,45 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
...
-                                                dpl_trial.data[layer],
-                                                tmin, tmax)
+                                                dpl_trial.data[layer], tmin,
+                                                tmax)
             if decim is not None:
                 data, times = _decimate_plot_data(decim, data, times)
 
-            ax.plot(times, data, color=color)
+            if idx == len(dpl) - 1 and average:
+                # the last one (average dpl) or the single dpl
```
jasmainak, line 276:
```suggestion
                # the last one (average dpl)
```

hnn_core/viz.py
```diff
@@ -242,29 +244,45 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
...
+                ax.plot(times,
+                        data,
+                        color=color,
+                        label="average",
+                        lw=linewidth * 1.5)
+            else:
+                ax.plot(times, data, color="gray", alpha=0.5, lw=linewidth)
```
jasmainak, line 283:
```suggestion
                alpha = 0.5 if average else 1.
                ax.plot(times, data, color=color, alpha=alpha, lw=linewidth)
```

hnn_core/viz.py
```diff
@@ -242,29 +244,45 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
...
+                                                tmax)
             if decim is not None:
                 data, times = _decimate_plot_data(decim, data, times)
 
-            ax.plot(times, data, color=color)
+            if idx == len(dpl) - 1 and average:
+                # the last one (average dpl) or the single dpl
+                ax.plot(times,
+                        data,
+                        color=color,
```
jasmainak, line 279:
```suggestion
                        color='g',
```

jasmainak: or any other aesthetically pleasing color you like :)

hnn_core/viz.py
```diff
@@ -210,7 +210,7 @@ def plot_extracellular(times, data, tmin=None, tmax=None, ax=None,
 
 
 def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
-                color=None, show=True):
+                color=None, average=True, show=True):
```
jasmainak, line 213:
```suggestion
                color=None, average=False, show=True):
```

I think this is a better default for two reasons: 1) it keeps the existing plots the same with the same color etc., 2) We probably don't want to see the average unless there are at least a few dipoles

hnn_core/viz.py
```diff
@@ -242,29 +244,45 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
...
     import matplotlib.pyplot as plt
-    from .dipole import Dipole
+    from .dipole import Dipole, average_dipoles
 
     if ax is None:
         _, ax = plt.subplots(1, 1, constrained_layout=True)
 
     if isinstance(dpl, Dipole):
         dpl = [dpl]
+        average = False
```
jasmainak, line 254:
```suggestion
```

Now with `average = False` as default you don't need to override the user-provided setting. Also, this avoids the different behavior between `dpl` and `[dpl]`. Both should give the same plot in my opinion.

hnn_core/viz.py
```diff
@@ -242,29 +244,45 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
...
+    else:
+        # add average plot when we have dpl>1 and `average` is specified.
```
jasmainak, line 256:
```suggestion
```

Now in docstring. Also it should have been `len(dpl) > 1`

jasmainak (approved):
Last round of reviews I hope :)

I let @rythorpe or @ntolley merge once these last comments are addressed and if they're happy.

hnn_core/viz.py
```diff
@@ -242,29 +244,45 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
...
+        if average:
```
jasmainak, line 257:
```suggestion
```

Otherwise, this will give buggy result I think

hnn_core/viz.py
jasmainak: mixing the two will cause problems. Hence, see what I [suggested above](https://github.com/jonescompneurolab/hnn-core/pull/475/files#r835728105)

hnn_core/viz.py
```diff
@@ -242,29 +244,45 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
...
+            dpl = dpl + [average_dipoles(dpl)]
 
+    linewidth = 1.5
     scale_applied = dpl[0].scale_applied
-    for dpl_trial in dpl:
+    for idx, dpl_trial in enumerate(dpl):
```
jasmainak, line 262: I'm trying to resist the temptation to avoid micro-optimizing your code but it's hard for me ;-)
```suggestion
    for dpl_trial in dpl[:-1]:
```
now if the last element of the `dpl` list is always the average, then you can move the:

```py
if idx == len(dpl) - 1 and average:
```

outside of the loop and into the:

```py
  if average:
        ax.legend()
```

chenghuzi: You mean also to move

```Python
            # extract scaled data and times
            data, times = _get_plot_data_trange(dpl_trial.times,
                                                dpl_trial.data[layer], tmin,
                                                tmax)
            if decim is not None:
                data, times = _decimate_plot_data(decim, data, times)
```

out of the loop and put into the if average?

jasmainak: indeed, sorry I didn't realize that would also need to change. I think it's fine as it is. I pushed some cosmetic changes. Made the default color to be "black" rather than do a color cycle and tweaked the linewidth a bit.

doc/whats_new.rst
jasmainak: this still needs fixing @chenghuzi

chenghuzi: If you mean the doc build error, I guess it's caused by my indentation error, which has been fixed in 85442df6ce6a056ff6cd39a5c0f7d2d16bf7babf.

chenghuzi: Fixed in 68a2761ee19d1cc06d86546c936f6bd7ec41b320

jasmainak:
For completeness: CircleCI does not trigger builds for first-time contributors to avoid passing the secrets ... probably a security hazard. See [here](https://github.com/corona-warn-app/cwa-app-android/issues/695#issuecomment-651738590). 

I'll push a dummy commit here to trigger the build and check if everything is alright.

jasmainak:
something wrong in this example here @chenghuzi : https://2173-168215891-gh.circle-artifacts.com/0/html/auto_examples/workflows/plot_simulate_evoked.html#sphx-glr-auto-examples-workflows-plot-simulate-evoked-py

There are only 2 dipoles but 3 are being plotted. If `average=False`, the average dipole shouldn't be plotted at all. I think the `if-else` condition needs fixing

jasmainak:
also why is the default color of the plot not black when only one dipole is plotted?

chenghuzi:
> something wrong in this example here @chenghuzi : https://2173-168215891-gh.circle-artifacts.com/0/html/auto_examples/workflows/plot_simulate_evoked.html#sphx-glr-auto-examples-workflows-plot-simulate-evoked-py
> 
> There are only 2 dipoles but 3 are being plotted. If `average=False`, the average dipole shouldn't be plotted at all. I think the `if-else` condition needs fixing

Oh sorry I forgot that. Fixed in d7ec7837a39998ccf32852b7af78aa8c60599912

chenghuzi:
> also why is the default color of the plot not black when only one dipole is plotted?

I thinks it's because the `color` of `plot` in `Dipole` is `None` and is passed to `plot_dipole`.

jasmainak:
I think all corner cases are addressed now, merging. Thanks @chenghuzi ! Congratulations on your first pull request 🎉

rythorpe:
What I’m proposing is that the color argument apply a uniform color across all trials if no average is to be computed and plotted. If it’s the case that we want individual trials to always be plotted in grey, than we should clarify in the docstring that the color parameter only applies to an average dipole if/when it’s computed within the plot_dipole function. I’m not sure this is the case, as there have been times when e.g. I’ve computed the trial-average dipole outside the plot_dipole function and then wished to plot it as a certain color within a preexisting matplotlib axes instance, but I’ll let @jasmainak and @ntolley weigh in here as well.

jasmainak:
@rythorpe old comment?

rythorpe:
Woah. Yeah something in the interwebs broke - I wrote this last spring.

PR #478, opened 2022-03-29
[MRG] specifiying which evoked drive to optimize over

mohdsherif:
Hi - please check the changes and let me know your thoughts. Thanks.


jasmainak:
@rythorpe is the real expert here, so I asked him for a review.

Is there a test we could add in the meanwhile?

mohdsherif:
@test - thinking about something. In the meantime, I think the last commit resulted in the three dipoles in the optimization plotted all in black. Might need to modify the code for plotting the tutorial result for that (not related to my changes).

In colors:
2022mar28_dipoles_firstEvoked_Optimization_a


In black:
2022mar29_dipoles_firstEvoked_Optimization_allBlack_a

jasmainak:
oops, sorry my bad. You can update the tutorial.

Now, for the test, you should update this function probably: https://github.com/jonescompneurolab/hnn-core/blob/b974fed4317b9b03ec732850829d9c17841c6ee2/hnn_core/tests/test_optimization.py#L79

and to run the test locally you can do:

```sh
$ cd hnn-core
$ pytest ./hnn_core/tests/test_optimization.py::test_optimize_evoked
```

We use the [`pytest` framework](https://docs.pytest.org/en/7.1.x/). Normally after every change, I run all the tests to make sure nothing was broken by doing:

```sh
$ pytest .
```

hnn_core/optimization.py
```diff
@@ -480,8 +483,12 @@ def optimize_evoked(net, tstop, n_trials, target_dpl, initial_dpl, maxiter=50,
...
     net = net.copy()
 
-    drive_names = [key for key in net.external_drives.keys()
-                   if net.external_drives[key]['type'] == 'evoked']
+    if which_drives == 'all':
+        drive_names = [key for key in net.external_drives.keys()
+                       if net.external_drives[key]['type'] == 'evoked']
+    else:
+        drive_names = [key for key in net.external_drives.keys()
+                       if net.external_drives[key]['type'] == 'evoked' and key in which_drives]
```
rythorpe, line 491: This looks good @mohdsherif! My only recommendation would be to throw a `ValueError` if the items in `which_drives` are not evoked drive names defined in `net.external_drives[key]['type']` so that the user doesn't accidentally define a drive that is silently ignored during optimization.

Then, you should add a test that raises this error. You can probably just add such a test [here](https://github.com/jonescompneurolab/hnn-core/blob/b974fed4317b9b03ec732850829d9c17841c6ee2/hnn_core/tests/test_optimization.py#L112) to test for all expected user errors that might emerge when `which_drives` is `'all'` and when it is of type list.

rythorpe:
I think there's a flake8 error as well. Make sure you have the `flake8` linter installed (either that runs automatically in your text editor or called explicitly via `$ flake8` from within your `~/hnn-core/hnn_core` directory).

rythorpe:
Also @mohdsherif, have you run the optimization example where only a subset of drives is selected to ensure that it's doing what we think its doing? When you get around to it, it might be nice to post a few screenshots just so we have it documented that drive selection works. Either that or we could consider modifying the optimization example to only optimize e.g. `evdist1` in order to demonstrate that all the other drives remain the same, but the 2nd proximal drive is modified during optimization.

rythorpe:
I see, that makes sense now. Thanks @mohdsherif!

mohdsherif:
@rythorpe and @jasmainak - I am running into an issue with pytest - it seems to be unable to see the change I made to the function. I can run the test function from python, but with pytest, it is failing and I am getting the following error message:

E       TypeError: optimize_evoked() got an unexpected keyword argument 'which_drives'

When I load test_optimization.py into python, I am able to run the function while passing arguments to the keyword 'which_drives'. I thought it might have something to do with pytest cache and downloaded pytest-cache package to explicitly clean pytest cache, but still I am getting the same error. Any thoughts?

jasmainak:
Did you make an editable install of `hnn-core` as detailed in the [contributing guide](https://github.com/jonescompneurolab/hnn-core/blob/master/CONTRIBUTING.rst)?

mohdsherif:
Yes, except for --depth 1 when cloning the repo.

jasmainak:
humm ... what do you get when you do:

```sh
$ ipython --matplotlib
```

and then 

```py
>>> import hnn_core
>>> hnn_core.__version__
```

jasmainak:
maybe you need to do:

```py
$ python -m pytest ./hnn_core/tests/test_optimization.py::test_optimize_evoked
```

as this invokes the `pytest` in the current environment (and hence hnn-core in the current environment)

rythorpe:
> maybe you need to do:
> 
> ```python
> $ python -m pytest ./hnn_core/tests/test_optimization.py::test_optimize_evoked
> ```
> 
> as this invokes the `pytest` in the current environment (and hence hnn-core in the current environment)

Yeah, this sounds almost like a python environment issue. Could it be calling a pytest library that exists outside your current environment?

hnn_core/tests/test_optimization.py
```diff
@@ -136,3 +136,91 @@ def test_optimize_evoked():
 
     # static drive params should remain constant
     assert drive_static_params_opt == drive_static_params_orig
+
+
+def test_optimize_evoked_selectedDrive():
+    """Test running optimzing a selected drive while parameters of other drives
+    stay the same in a reduced network."""
```
jasmainak, line 143: nitpick but [python style conventions](https://peps.python.org/pep-0257/#one-line-docstrings) dictate that this should be one line.

hnn_core/tests/test_optimization.py
```diff
@@ -136,3 +136,91 @@ def test_optimize_evoked():
...
+    n_trials = 1
+
+    # simulate a dipole to establish ground-truth drive parameters
+    mu_orig = 6.
+    params.update({'N_pyr_x': 3,
+                   'N_pyr_y': 3,
+                   't_evprox_1': mu_orig,
+                   'sigma_t_evprox_1': 2.,
+                   't_evdist_1': mu_orig + 2,
+                   'sigma_t_evdist_1': 2.})
```
jasmainak, line 158: there seems to be quite a bit of code duplication with `test_optimize_evoked`. If we add these tests to that function instead of creating a new function, could we reduce this duplication?

hnn_core/tests/test_optimization.py
```diff
@@ -136,3 +136,91 @@ def test_optimize_evoked():
 
     # static drive params should remain constant
     assert drive_static_params_opt == drive_static_params_orig
+
+
+def test_optimize_evoked_selectedDrive():
```
jasmainak, line 141: another [style convention](https://peps.python.org/pep-0008/#function-and-variable-names). In Python, function names are usually snake_case rather than CamelCase. This is more Matlab style :)

jasmainak:
Looking great @mohdsherif . Once you're done, don't forget to update `whats_new.rst` !

mohdsherif:
- Appreciate the "nitpicking" - I understand it is important to keep the style consistent.
- Re/ duplicate code - yes - I will rewrite parts of the test_optimize_evoked to include the testing for selected drives instead.

mohdsherif:
@rythorpe  and @jasmainak - let me know if you have any other comments.

doc/whats_new.rst
```diff
@@ -55,7 +55,8 @@ Notable Changes
 - Local field potentials can now be recorded during simulations 
   :ref:`[Example] <sphx_glr_auto_examples_howto_plot_record_extracellular_potentials.py>`
 
-- Ability to optimize parameters to reproduce event related potentials from real data 
+- Ability to optimize parameters to reproduce event related potentials from real data,
```
jasmainak, line 58: this is a change from previous version 0.2. You should add yours under "Current" version under the category "Changelog". 

I just realized that some folks have added theirs to the "Bug" section when it actually belongs to "Changelog" ... we'll fix that separately

hnn_core/tests/test_optimization.py
```diff
@@ -90,21 +90,25 @@ def test_optimize_evoked():
...
                    'N_pyr_y': 3,
                    't_evprox_1': mu_orig,
-                   'sigma_t_evprox_1': 2.})
+                   'sigma_t_evprox_1': 2.,
+                   't_evdist_1': mu_orig + 2,
+                   'sigma_t_evdist_1': 2.})
     net_orig = jones_2009_model(params, add_drives_from_params=True)
     del net_orig.external_drives['evprox2']
-    del net_orig.external_drives['evdist1']
+    # del net_orig.external_drives['evdist1']
```
jasmainak, line 98: don't leave commented code unless it's to explain something.

@rythorpe perhaps you want to take a look here since you wrote this originally?

hnn_core/tests/test_optimization.py
```diff
@@ -113,15 +117,26 @@ def test_optimize_evoked():
...
         del net_empty.external_drives['evprox1']
+        del net_empty.external_drives['evdist1']
         net_opt = optimize_evoked(net_empty, tstop=tstop, n_trials=n_trials,
                                   target_dpl=dpl_orig, initial_dpl=dpl_offset)
 
+    with pytest.raises(ValueError, match='The drives selected to be optimized '
+                       'are not evoked drives'):
+        net_testBursty = net_offset.copy()
+        which_drives = ['bursty1']
+        net_opt = optimize_evoked(net_testBursty, tstop=tstop,
```
jasmainak, line 128:
```suggestion
        net_opt = optimize_evoked(net_test_bursty, tstop=tstop,
```

snake_case, snake_case :)

jasmainak:
looking good except some minor comments! Can you set the title of the PR with prefix [MRG] when it's ready from your end?

mohdsherif:
thank you. Let me know of any final edits (hopefully none ;-) )

doc/whats_new.rst
```diff
@@ -12,6 +12,9 @@ Current
 
 Changelog
 ~~~~~~~~~
+- Ability to select which drives to optimize for event related potentials. 
+  :ref:`[Example] <sphx_glr_auto_examples_howto_plot_optimize_evoked.py>`
```
jasmainak, line 16:
```suggestion
- Add option to select drives using argument `which_drives` in :func:`~hnn_core.optimization.optimize_evoked`, by `Mohamed Sherif`_ in :gh:`478`.
```

and then down below add your name with a link. Here is a rendering of the page:

https://jonescompneurolab.github.io/hnn-core/dev/whats_new.html

hnn_core/optimization.py
```diff
@@ -480,12 +484,22 @@ def optimize_evoked(net, tstop, n_trials, target_dpl, initial_dpl, maxiter=50,
...
                          'drives. Consider adding drives using '
                          'net.add_evoked_drive')
+    elif which_drives == 'all':
+        drive_names = evoked_drive_names
+    else:
+        drive_names = [mydrive for mydrive in np.unique(which_drives)
+                       if mydrive in evoked_drive_names]
+    if len(drive_names) == 0:
+        raise ValueError('The drives selected to be optimized are not evoked '
+                         'drives. Optimization works only evoked drives.')
```
rythorpe, line 501: This looks good! For brevity's sake, we could probably combine the initial check for `len(evoked_drive_names == 0` into the final check for `len(drive_names) == 0` and raise a more general error about no available drives being selected, but that is completely a stylistic suggestion. I'll let @mohdsherif and @jasmainak have the final say here.

rythorpe:
Aside from my very minor comment above and @jasmainak's comment on `whats_new.rst`, I think this is ready!

hnn_core/optimization.py
jasmainak: I actually like the separate messages. It's a bit more informative for the user :)

mohdsherif:
Thanks to both of your comments. Please take a look.

jasmainak:
rebased and merged after cosmit 1221bfca134733da531b6b81f0334b509fe89b20 

thanks @mohdsherif ! 🎉  looking forward to more pull requests. I promise it gets less painful ;-) The first one is always the hardest

mohdsherif:
thanks a lot @jasmainak and @rythorpe - wasn't as painful as I was worried about ;-) - looking forward to continuing working together.

PR #500, opened 2022-07-13
[MRG] GUI refactoring

chenghuzi:
Try to refactor the GUI code in an OP manner.

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -960,7 +960,7 @@ def on_upload_change(change, params, tstop, tstep, log_out, variables,
 
 
 def _init_network_from_widgets(params, tstep, tstop, variables, drive_widgets,
-                               connectivity_sliders):
+                               connectivity_sliders, add_drive=True):
     """Construct network and add drives."""
     print("init network")
     params['dt'] = tstep.value
@@ -986,6 +986,8 @@ def _init_network_from_widgets(params, tstep, tstop, variables, drive_widgets,
                 variables['net'].connectivity[conn_idx][
                     'probability'] = vbox.children[3].value
 
+    if add_drive is False:
+        return
     # add drives to network
     for drive in drive_widgets:
         logging.debug(f"add drive type to network: {drive['type']}")
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -96,7 +96,7 @@ class HNNGUI:
                             description='MPI cmd:',
                             disabled=False)
 
-        self.joblib_cores = BoundedIntText(value=1,
+        self.n_jobs = BoundedIntText(value=1,
                                            min=1,
                                            max=multiprocessing.cpu_count(),
                                            description='Cores:',
@@ -216,7 +216,7 @@ class HNNGUI:
         # link callbacks
         def _handle_backend_change(backend_type):
             return handle_backend_change(backend_type.new, self.backend_config,
-                                         self.mpi_cmd, self.joblib_cores)
+                                         self.mpi_cmd, self.n_jobs)
 
         def _add_drive_button_clicked(b):
             return add_drive_widget(self.drive_type_selection.value,
@@ -243,7 +243,7 @@ class HNNGUI:
             return run_button_clicked(
                 self.log_out, self.drive_widgets, self.variables, self.tstep,
                 self.tstop, self.ntrials, self.backend_selection, self.mpi_cmd,
-                self.joblib_cores, self.params, self.plot_outputs_list,
+                self.n_jobs, self.params, self.plot_outputs_list,
                 self.plot_dropdowns_list, self.simulation_status, b)
 
         def _handle_viz_layout_change(layout_option):
@@ -904,7 +904,7 @@ def init_network_from_widgets(params, tstep, tstop, variables, drive_widgets):
 
 
 def run_button_clicked(log_out, drive_widgets, variables, tstep, tstop,
-                       ntrials, backend_selection, mpi_cmd, joblib_cores,
+                       ntrials, backend_selection, mpi_cmd, n_jobs,
                        params, plot_outputs_list, plot_dropdowns_list,
                        simulation_status, b):
     """Run the simulation and plot outputs."""
@@ -918,8 +918,8 @@ def run_button_clicked(log_out, drive_widgets, variables, tstep, tstop,
             variables['backend'] = MPIBackend(
                 n_procs=multiprocessing.cpu_count() - 1, mpi_cmd=mpi_cmd.value)
         else:
-            variables['backend'] = JoblibBackend(n_jobs=joblib_cores.value)
-            print(f"Using Joblib with {joblib_cores.value} core(s).")
+            variables['backend'] = JoblibBackend(n_jobs=n_jobs.value)
+            print(f"Using Joblib with {n_jobs.value} core(s).")
         with variables['backend']:
             simulation_status.value = """<div
             style='background:orange;padding-left:10px;color:white;'>
@@ -947,13 +947,13 @@ def run_button_clicked(log_out, drive_widgets, variables, tstep, tstop,
             })
 
 
-def handle_backend_change(backend_type, backend_config, mpi_cmd, joblib_cores):
+def handle_backend_change(backend_type, backend_config, mpi_cmd, n_jobs):
     backend_config.clear_output()
     with backend_config:
         if backend_type == "MPI":
             display(mpi_cmd)
         elif backend_type == "Joblib":
-            display(joblib_cores)
+            display(n_jobs)
 
 
 def init_left_right_viz_layout(plot_outputs,
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -16,8 +16,8 @@ from IPython.display import display
 from ipywidgets import (HTML, Accordion, AppLayout, BoundedFloatText,
                         BoundedIntText, Button, Dropdown, FileUpload,
                         FloatLogSlider, FloatText, GridspecLayout, HBox,
-                        IntText, Layout, Output, RadioButtons, Tab,
-                        Text, VBox, interactive_output)
+                        IntText, Layout, Output, RadioButtons, Tab, Text, VBox,
+                        interactive_output)
 
 import hnn_core
 from hnn_core import (JoblibBackend, MPIBackend, jones_2009_model, read_params,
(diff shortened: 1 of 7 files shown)


hnn_core/gui/gui.py
```diff
@@ -29,6 +29,263 @@
 THEMECOLOR = "#8A2BE2"
 
 
+class HNNGUI:
```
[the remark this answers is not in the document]
chenghuzi: Sure, it's still in progress. Will add later

hnn_core/tests/test_gui.py
```diff
@@ -1,8 +1,9 @@
...
+from hnn_core.gui.gui import HNNGUI
 
 
 def test_run_gui():
     """Test if main gui function gives proper ipywidget"""
-    gui = run_hnn_gui()
-    assert gui is not None
+    gui = HNNGUI()
+    app_layout = gui.run()
+    assert app_layout is not None
```
jasmainak, line 9: can you add a couple of tests using the new class so we can see how the new structure helps?

hnn_core/gui/gui.py
```diff
@@ -26,15 +26,307 @@
...
+        The theme color of the whole dashboard.
+    log_window_height: str
+        The height of the log output window (in pixel).
+    visualization_window_width: str
+        The width of the visualization window (in pixel).
+    visualization_window_height: str
+        The height of the visualization window (in pixel).
+    left_sidebar_width: str
+        The width of the left side bar (in pixel).
+    TODO: add more parameters.
```
rythorpe, line 45: I love the idea of having access to these parameters in the constructor rather than just hard-coding them, but I'm not clear on how/when they'll be used. Out of curiosity, what's your high-level vision here @chenghuzi?

hnn_core/gui/gui.py
```diff
@@ -26,15 +26,307 @@
...
+                                                      ('Vertical', 'U-D')],
+                                             value='L-R',
+                                             description='Layout:')
+
+        self.mpi_cmd = Text(value='mpiexec',
+                            placeholder='Fill if applies',
+                            description='MPI cmd:',
+                            disabled=False)
+
+        self.joblib_cores = BoundedIntText(value=1,
```
jasmainak, line 99: self.n_jobs

as in the rest of the codebase

jasmainak:
see how the `Makefile` is used to make the test commands consistent across different CIs and developers: https://github.com/tqdm/tqdm/blob/4f208e72552c4d916aa4fe6a955349ee8b2ed353/Makefile#L65

chenghuzi:
> @chenghuzi I was looking at `tqdm` and it seems they also have a notebook for the tests of the notebook: https://github.com/tqdm/tqdm/blob/4591083884c8ca079ccd1192c079113c7baeadc8/tests_notebook.ipynb. I think that might be the simplest solution for us to avoid complicated frameworks ...

I think probably we can directly call this in a `. py` file. I'll submit a commit about this soon

hnn_core/gui/gui.py
chenghuzi: You can check that now in our latest gui tests

hnn_core/tests/test_gui.py
chenghuzi: check the latest commit

hnn_core/gui/gui.py
```diff
@@ -26,15 +26,307 @@
...
+            # by default load default.json
+            hnn_core_root = op.join(op.dirname(hnn_core.__file__))
+            params_fname = op.join(hnn_core_root, 'param', 'default.json')
+
+        return read_params(params_fname)
+
+    @staticmethod
+    def init_cell_connectivity(params):
+        return [
+            _get_sliders(params, [
```
ntolley, line 196: Relevant to https://github.com/jonescompneurolab/hnn-core/issues/497#issuecomment-1186600445

hnn_core/tests/test_gui.py
```diff
@@ -1,8 +1,114 @@
...
+    assert isinstance(dpls, list)
+    assert all([isinstance(dpl, Dipole) for dpl in dpls])
+
+# For other tests, we can follow this paradigm:
+# 1. Initialize the gui
+# 2. Change some parameters from the widget interface.
+# 3. Use modified parameters to construct the Network model
+# 4. (Optional) run simulation
+# 5. Check if the simulation results or network with modified parameters
+#    match.
```
jasmainak, line 95: This makes sense to me!

hnn_core/tests/test_gui.py
```diff
@@ -1,8 +1,114 @@
...
+
+# def test_gui_run_simulation():
+#     """Test if gui can reproduce the same results as using hnn-core"""
+#     pass
+
+# def test_gui_update_simulation_parameters():
```
jasmainak, line 101: Go ahead and finish these tests!

hnn_core/gui/gui.py
```diff
@@ -26,15 +26,307 @@
...
+        The height of the visualization window (in pixel).
+    left_sidebar_width: str
+        The width of the left side bar (in pixel).
+    TODO: add more parameters.
+
+    Attributes
+    ----------
+    params: dict
+        The parameters to use for constructing the network.
+    TODO: add more attributes
```
jasmainak, line 51: Add them all! Otherwise it's hard to evaluate if the class needs restructuring or not

hnn_core/gui/gui.py
jasmainak, line 45: add them all!

jasmainak:
How do you plan to use the class and the methods in the documentation? Let's say you are describing the part where drives must be added? You would want to: 

a) activate the drive tab
b) click "add drive" 
c) select drive parameters
d) show the current state of GUI (i.e., re-paint the layout/GUI)

is that something that can be done with the current interface?

jasmainak:
@chenghuzi you need to update `api.rst`

hnn_core/gui/gui.py
```diff
@@ -536,91 +842,97 @@ def on_upload_change(change, sliders, params, tstop, tstep, log_out, variables,
                 drive_boxes, tstop)
 
 
+def init_network_from_widgets(params, tstep, tstop, variables, drive_widgets):
+    """Construct network and add drives."""
```
jasmainak, line 846: Document all public functions and methods!

hnn_core/gui/gui.py
```diff
@@ -26,15 +26,307 @@
...
+        self.drive_widget_width = drive_widget_width
+
+        # load default parameters
+        self.params = self.load_parameters()
+
+        # In-memory storage of all simulation related variables
+        self.variables = dict(net=None, dpls=list())
+
+        # Simulation parameters
+        self.tstop = FloatText(value=170,
```
jasmainak, line 74: I think you should have `self.widgets` which is a dictionary and similarly `self.callbacks` which is a dictionary. The dictionaries have matching key names ... that way it will be easy to inspect the object

hnn_core/gui/gui.py
chenghuzi: I think I'll turn this into a private one... as no other part of hnn-core should call this

hnn_core/gui/gui.py
chenghuzi: `self.layouts` is a good point, but I'd like to keep all widgets and callbacks as direct member of the HNNGUI class, which makes documentation and tests easier

jasmainak: The object should not have > 10 attributes ... otherwise it will be a mess

hnn_core/gui/gui.py
```diff
@@ -26,15 +26,307 @@
...
+        self._init_ui_components()
+
+    def _init_ui_components(self):
+        """Initialize larger UI components and dynamical output windows. It's
+        not encouraged for users to modify or access attributes in this part.
+        """
+        self.connectivity_sliders = self.init_cell_connectivity(self.params)
+        # dynamic larger components
+        self.drives_out = Output()  # window to add new drives
+        self.log_out = Output(
```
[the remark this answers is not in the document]
chenghuzi: I tend to not to do so because almost all attributes in this method are  not supposed to be accessible to users. I'll mark then as private with `_`prefix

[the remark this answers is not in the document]
jasmainak: Just declare everything in init, whether private or not. Also init should have minimal logic

See: https://betterprogramming.pub/6-best-practices-for-defining-the-initialization-method-in-python-9173808bfb8f.

chenghuzi:
We have dynamic connectivity tab now in 8c920a2b19e5d44a902ea9d618800d861d943ae3

hnn_core/gui/gui.py
```diff
@@ -504,129 +921,162 @@ def load_drives(variables, params, log_out, drives_out, drive_widgets,
...
+    for connectivity_slider in connectivity_sliders:
+        for vbox in connectivity_slider:
+            conn_indices = pick_connection(
+                net=variables['net'],
+                src_gids=vbox._belongsto['src_gids'],
+                target_gids=vbox._belongsto['target_gids'],
+                loc=vbox._belongsto['location'],
+                receptor=vbox._belongsto['receptor'])
+
+            if len(conn_indices) > 0:
```
ntolley, line 982: There should be a check to ensure that there is only one element in `conn_indices`

jasmainak:
@chenghuzi I really like how the tests are shaping up. It looks really nice. I just tried running them and noticed that plots start popping up. Do we need to set the matplotlib "Agg" backend in the test? You can see how it's done in the other tests

chenghuzi:
> @chenghuzi I really like how the tests are shaping up. It looks really nice. I just tried running them and noticed that plots start popping up. Do we need to set the matplotlib "Agg" backend in the test? You can see how it's done in the other tests

Oh I didn't notice this as my local tests are on the server. I think you're right and I'll add it soon.

jasmainak:
Also the tests could be faster if you use a smaller network? Maybe the class `HNNGUI` could have an option of passing a params dictionary? Then you can do:

```py
params['N_Pyr_x'] = 3
params['N_Pyr_y'] = 3

HNNGUI(params=params)
```

hnn_core/tests/test_gui.py
```diff
@@ -1,8 +1,187 @@
...
+    _ = gui.run()
+
+    params_name = 'default.json'
+    hnn_core_root = op.join(op.dirname(hnn_core.__file__))
+    params_fname = op.join(hnn_core_root, 'param', params_name)
+
+    content = b""
+    with open(params_fname, "rb") as f:
+        for line in f:
+            pass
```
jasmainak, line 38: is this intentional or an artifact of debugging?

chenghuzi: It's used to load the byte string from the parameter file so that we can simulate the uploading event. But I need to remove the `pass`

chenghuzi:
> Also the tests could be faster if you use a smaller network? Maybe the class `HNNGUI` could have an option of passing a params dictionary? Then you can do:
> 
> ```python
> params['N_Pyr_x'] = 3
> params['N_Pyr_y'] = 3
> 
> HNNGUI(params=params)
> ```

For tests we can do that manually

chenghuzi:
> @chenghuzi you need to update `api.rst`

Updated in c39810a254fd568a32b0a113b99608c025fb7099

jasmainak:
should the code logic for finding possible connectivity pairs between cell types be in a separate function? I can imagine this logic could be useful outside of the GUI as well.

chenghuzi:
> should the code logic for finding possible connectivity pairs between cell types be in a separate function? I can imagine this logic could be useful outside of the GUI as well.

Agreed. But we may need another PR for this because the current way we use to extract connections is hierarchical while a general method (possibly in the form of a method in `Network`) should be plain. And then we can further simplify the existing code for extracting possible connectivity boxes using that method.

hnn_core/gui/gui.py
```diff
@@ -9,58 +9,427 @@
...
+
+        # In-memory storage of all simulation related variables
+        self.variables = dict(net=None, dpls=list())
+
+        # Simulation parameters
+        self.tstop = FloatText(value=170,
+                               description='tstop (ms):',
+                               disabled=False)
+        self.tstep = FloatText(value=0.025,
+                               description='tstep (ms):',
```
ntolley, line 123:
```suggestion
                               description='dt (ms):',
```

chenghuzi:
`variables` is now `simulation_data`.  We keep that because it allows modifying the item in the dict easily, otherwise we need to return the value to modify them

jasmainak:
Great. I'm doing a pass over the code right now. Report back in an hour :-)

jasmainak:
okay back to you. Finish this PR: https://github.com/chenghuzi/hnn-core/pull/2/files

and I think we're good to go here. 

I also had some trouble following the logic of loading the param files. But let's discuss that later.

hnn_core/gui/gui.py
```diff
@@ -504,167 +922,199 @@ def load_drives(variables, params, log_out, drives_out, drive_widgets,
...
-    params_fname = change['owner'].metadata[0]['name']
+    # params_fname = change['owner'].metadata[0]['name']
+    params_fname = list(change['owner'].value.keys())[0]
     file_uploaded = change['owner'].value
     param_data = list(file_uploaded.values())[0]['content']
     param_data = codecs.decode(param_data, encoding="utf-8")
 
+    if load_info['prev_param_data'] == param_data:
+        with log_out:
+            print("Same param. No reloading. \
```
jasmainak, line 939:
```suggestion
            print("Same param. No reloading."
```
nitpick ... backslashes are not needed within parenthesis. They make the code look ugly. Better to use parenthesis and align.

hnn_core/tests/test_gui.py
```diff
@@ -1,8 +1,198 @@
...
+
+    # check if parameter is reloaded.
+    assert gui.widget_tstop.value == original_tstop
+    assert gui.widget_dt.value == original_tstep
+    assert len(gui.drive_widgets) == original_drive_count
+
+
+def test_gui_change_connectivity():
+    """Test if GUI properly changes cell connectivity parameters.
+    """
```
jasmainak, line 74: The quotes should be in the previous line. Fix everywhere else. One line docstrings should really be one line.

chenghuzi:
> okay back to you. Finish this PR: https://github.com/chenghuzi/hnn-core/pull/2/files
> 
> and I think we're good to go here.
> 
> I also had some trouble following the logic of loading the param files. But let's discuss that later.

All format issues fixed

hnn_core/gui/gui.py
```diff
@@ -943,7 +943,8 @@ def on_upload_change(change, params, tstop, dt, log_out, simulation_data,
         with log_out:
             print(
                 "Same param. No reloading."
-                "To force reloading, hit \"clear uploaded parameters\" button")
+                "To force reloading, hit \"clear uploaded parameters\" button",
```
jasmainak, line 959: see: https://peps.python.org/pep-0008/#string-quotes

"When a string contains single or double quote characters, however, use the other one to avoid backslashes in the string. It improves readability."

Just avoid backslashes when you can :)

jasmainak:
Never merge your own PR :) But I think it's time to move on to the next steps! Thanks @chenghuzi . This is a great improvement 🥳 🥳 🥳 . Next step is to try and figure out automatic building of documentation using `html2canvas` and [jupyter-require](https://pypi.org/project/jupyter-require/) ?

chenghuzi:
> > okay back to you. Finish this PR: https://github.com/chenghuzi/hnn-core/pull/2/files
> > and I think we're good to go here.
> > I also had some trouble following the logic of loading the param files. But let's discuss that later.
> 
> All format issues fixed



> Never merge your own PR :) But I think it's time to move on to the next steps! Thanks @chenghuzi . This is a great improvement 🥳 🥳 🥳 . Next step is to try and figure out automatic building of documentation using `html2canvas` and [jupyter-require](https://pypi.org/project/jupyter-require/) ?

I figured out a way to do that, though not every elegant. Will open a PR for that soon.

PR #529, opened 2022-08-02
[MRG] hnn-core GUI ERP partial tutorial

chenghuzi:
* Added first ERP tutorial (unfinished). 
* Exposed few methods used to manipulate the GUI

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -34,7 +34,7 @@ _plot_options = {
 }
 
 
-def update_plot_window(simulation_data, _plot_out, plot_type):
+def _update_plot_window(simulation_data, _plot_out, plot_type):
     """Refresh plots with data from simulation_data."""
     _plot_out.clear_output()
     if not (plot_type['type'] == 'change' and plot_type['name'] == 'value'):
@@ -98,9 +98,9 @@ def update_plot_window(simulation_data, _plot_out, plot_type):
                 print("No network data")
 
 
-def init_left_right_viz_layout(plot_outputs, plot_dropdowns,
-                               simulation_data, plot_options, previous_outputs,
-                               layout, init=False):
+def _init_left_right_viz_layout(plot_outputs, plot_dropdowns, simulation_data,
+                                plot_options, previous_outputs, layout,
+                                init=False):
     layout_LR = Layout(width=f"{int(layout.width[:-2])/2-10}px",
                        height=layout.height,
                        border=layout.border)
@@ -152,12 +152,12 @@ def init_left_right_viz_layout(plot_outputs, plot_dropdowns,
     plot_dropdowns.append(plot_dropdown_R)
 
     if not init:
-        update_plot_window(simulation_data, plot_outputs_L, {
+        _update_plot_window(simulation_data, plot_outputs_L, {
             "type": "change",
             "name": "value",
             "new": default_plot_types[0]
         })
-        update_plot_window(simulation_data, plot_outputs_R, {
+        _update_plot_window(simulation_data, plot_outputs_R, {
             "type": "change",
             "name": "value",
             "new": default_plot_types[1]
@@ -168,9 +168,9 @@ def init_left_right_viz_layout(plot_outputs, plot_dropdowns,
     return grid
 
 
-def init_upper_down_viz_layout(plot_outputs, plot_dropdowns, simulation_data,
-                               plot_options, previous_outputs,
-                               layout, init=False):
+def _init_upper_down_viz_layout(plot_outputs, plot_dropdowns, simulation_data,
+                                plot_options, previous_outputs, layout,
+                                init=False):
 
     default_plot_types = [plot_options[0], plot_options[1]]
     for idx, plot_type in enumerate(previous_outputs[:2]):
@@ -226,12 +226,12 @@ def init_upper_down_viz_layout(plot_outputs, plot_dropdowns, simulation_data,
     plot_dropdowns.append(plot_dropdown_D)
 
     if not init:
-        update_plot_window(simulation_data, plot_outputs_U, {
+        _update_plot_window(simulation_data, plot_outputs_U, {
             "type": "change",
             "name": "value",
             "new": default_plot_types[0]
         })
-        update_plot_window(simulation_data, plot_outputs_D, {
+        _update_plot_window(simulation_data, plot_outputs_D, {
             "type": "change",
             "name": "value",
             "new": default_plot_types[1]
@@ -243,7 +243,7 @@ def init_upper_down_viz_layout(plot_outputs, plot_dropdowns, simulation_data,
     return grid
 
 
-def change_plot_type(grid, layout, plot_idx, plot_type):
+def _change_plot_type(grid, layout, plot_idx, plot_type):
     assert layout in _plot_options.values(), ValueError("Illegal layout type")
     if layout == "L-R":
         assert 0 <= plot_idx < 2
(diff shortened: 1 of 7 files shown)


hnn_core/gui/gui.py
```diff
@@ -518,6 +796,42 @@ def run_notebook_cells(self):
         """
         return js_string
 
+    # below are a series of methods that are used to manipulate the GUI
+    def _simulate_upload_file(self, file_url):
+        import urllib.request
```
jasmainak, line 801: import should be at the top of the file

hnn_core/gui/gui.py
```diff
@@ -518,6 +796,42 @@ def run_notebook_cells(self):
...
+        tab_index = None
+        for idx in self.app_layout.left_sidebar._titles.keys():
+            if tab_title == self.app_layout.left_sidebar._titles[idx]:
+                tab_index = int(idx)
+                break
+        if tab_index is None:
+            raise ValueError("Incorrect tab title")
+        self.app_layout.left_sidebar.selected_index = tab_index
+
+    def _simulate_switch_plot_type(self, plot_idx, plot_type):
```
jasmainak, line 830: why are these private methods instead of public?

hnn_core/gui/gui.py
```diff
@@ -28,6 +28,271 @@
...
+        grid[layout].children[plot_idx].children[0].value = plot_type
+        pass
+    elif layout == "U-D":
+        assert 0 <= plot_idx < 2
+        grid[layout].children[plot_idx].children[0].value = plot_type
+    else:
+        pass
+
+
+def initialize_viz_window(viz_grid, viz_window, simulation_data, plot_outputs,
```
jasmainak, line 259: This should be private, no?

hnn_core/gui/gui.py
```diff
@@ -28,6 +28,271 @@
...
+            "new": default_plot_types[1]
+        })
+
+    grid = GridspecLayout(2, 1, height=layout.height)
+    grid[0, 0] = VBox([plot_dropdown_U, plot_outputs_U])
+    grid[1, 0] = VBox([plot_dropdown_D, plot_outputs_D])
+    return grid
+
+
+def change_plot_type(grid, layout, plot_idx, plot_type):
```
jasmainak, line 246: private

hnn_core/gui/gui.py
```diff
@@ -518,6 +796,42 @@ def run_notebook_cells(self):
         """
         return js_string
 
+    # below are a series of methods that are used to manipulate the GUI
+    def _simulate_upload_file(self, file_url):
```
jasmainak, line 800: use this also in `test_gui.py` ?

jasmainak:
@chenghuzi any updates here? When can we merge this tutorial?

ntolley:
Isn't the goal to implement [this](https://jonescompneurolab.github.io/hnn-tutorials/erp/erp) tutorial verbatim? 

@jasmainak I think we need to walk through the tutorials ourselves before moving to merge this

jasmainak:
I agree, I am waiting for @chenghuzi to tell us when it is ready from his end ... we need to really ramp up the pace. Still left to do:

- finish at least one tutorials
- changing the figure layout
- resizing GUI size
- miscellaneous improvements such as legacy_mode=False as default

jasmainak:
@chenghuzi let's add the "analysis" tab in a separate PR please ...

chenghuzi:
> @chenghuzi let's add the "analysis" tab in a separate PR please ...

But we need this for tutorial
image

jasmainak:
The visualizations should not refresh any time a widget in the "analysis" tab changes, this is very distracting. We should discuss this in a separate PR

jasmainak:
make a separate PR, get that merged and then rebase this PR :) Otherwise diff becomes hard to read

jasmainak:
Also, if I want to start my tutorial, I won't have the latest GUI changes if they are not in master.

chenghuzi:
> The visualizations should not refresh any time a widget in the "analysis" tab changes, this is very distracting. We should discuss this in a separate PR

I don't think so, because some parameters under analysis change the visualization results. For example,  smoothing window size changes the shape of dipole curve.

chenghuzi:
> make a separate PR, get that merged and then rebase this PR :) Otherwise diff becomes hard to read

OK

jasmainak:
I think a button called "update plot" should be used

chenghuzi:
> I think a button called "update plot" should be used

That's not reactive

jasmainak:
@chenghuzi @ntolley @rythorpe let's chat here: https://hnn-core.cloud.mattermost.com/main/channels/gsoc ... 

we need a chatroom to iterate fast

chenghuzi:
> @chenghuzi @ntolley @rythorpe let's chat here: https://hnn-core.cloud.mattermost.com/main/channels/gsoc ...
> 
> we need a chatroom to iterate fast

It seems I can't sign up...

jasmainak:
> That's not reactive

If you want to make it reactive, it should only update the line of the plot, not the entire visualization area. Right now you have an in-between solution. Doing a proper reactive solution will add complexity, which I'm not sure if it's worth it

chenghuzi:
> > That's not reactive
> 
> If you want to make it reactive, it should only update the line of the plot, not the entire visualization area. Right now you have an in-between solution. Doing a proper reactive solution will add complexity, which I'm not sure if it's worth it

If we want to control the granularity of reactive update, that will be extremely complex. And now, it's just a single flash and everything will be updated, so I don't think it's very annoying compared with hitting a "update plot" button many times.

jasmainak:
> It seems I can't sign up...

sign up with github? what do you see?

chenghuzi:
> > It seems I can't sign up...
> 
> sign up with github? what do you see?
This is what I got:
image

I assume this is something like slack so we need username and password for each organization?

jasmainak:
just sent an invite, does that help?

chenghuzi:
I think it's ok now. I also changed several methods into private ones @jasmainak

jasmainak:
I'll merge once the CIs come green!

jasmainak:
The linkcheck is failing because of missing `nbsphinx` ...

hnn_core/gui/gui.py
jasmainak: did you address this @chenghuzi ?

chenghuzi:
> The linkcheck is failing because of missing `nbsphinx` ...

Just have it fixed. But we still got error as neuron.yale.edu is inaccessible. It seems this happened many times before and I tried to visit their website with safari and it does not work. Should we do something to this?

jasmainak:
We should add that URL here:

https://github.com/jonescompneurolab/hnn-core/blob/1170bbc1e1c046128e4dfc8b3ae09ba1feb1d138/doc/conf.py#L205

hnn_core/gui/gui.py
jasmainak: here is the part of the test I am imagining will become simpler: https://github.com/jonescompneurolab/hnn-core/blob/1170bbc1e1c046128e4dfc8b3ae09ba1feb1d138/hnn_core/tests/test_gui.py#L33-L50

jasmainak:
@chenghuzi I left another comment inline with the code in case you missed it.

hnn_core/gui/gui.py
chenghuzi: It's now fixed in our latest commit

chenghuzi:
We've passed all tests, @jasmainak could you help to review and merge if it looks ok?

jasmainak:
Perfect, thank you @chenghuzi 🥳 🥳 🥳 

What is the next step?

chenghuzi:
> Perfect, thank you @chenghuzi 🥳 🥳 🥳
> 
> What is the next step?

To finish the analysis tab as well as the simulation loading part 😄

PR #541, opened 2022-08-11
[MRG] Add multi-simulation support

chenghuzi:
Users now can run multiple simulations and compare them. Also simplified some layout init functions

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -1478,6 +1478,12 @@ def run_button_clicked(widget_simulation_name, log_out, drive_widgets,
     log_out.clear_output()
 
     with log_out:
+        # clear empty trash simulations
+        for _name in tuple(simulation_data.keys()):
+            if simulation_data[_name]['net'] is None or len(
+                    simulation_data[_name]['dpls']) == 0:
+                del simulation_data[_name]
+
         _sim_name = widget_simulation_name.value
         if simulation_data[_sim_name]['net'] is not None:
             print("Simulation with the same name exists!")
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -11,6 +11,7 @@ import os.path as op
 import sys
 import urllib.parse
 import urllib.request
+from collections import defaultdict
 
 import hnn_core
 import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


hnn_core/gui/gui.py
```diff
@@ -300,32 +251,24 @@ def _initialize_viz_window(simulation_data, analysis_config, init=False):
...
-        viz_grid[layout_option] = grid
-        display(grid)
+    with viz_window:
+        viz_grid[layout_option] = _init_viz_layout(
+            simulation_data,
+            plot_options,
+            previous_plot_types,
+            analysis_config,
+            layout,
+            layout_option,
```
jasmainak, line 269: use keyword arguments to be clearer when there are many parameters to the function ... is it `layout` or `style`? Seems like you call it `style` inside the function

chenghuzi: Yeah there's some inconsistency... will fix later. Now I need to fix the GUI test first 😂

chenghuzi:
Now it looks like this: 

image

hnn_core/gui/gui.py
chenghuzi: Fixed in commit 0eb29c7a3118f0b2302abed83c3a2e567c4f15af

hnn_core/gui/gui.py
```diff
@@ -43,36 +45,51 @@
...
+        A dict of simulation data
+    analysis_config: dict
+        A dict of visualization configs
+    plot_key: str
+        A string used to identify the plot and data to update.
+    """
     # Make sure that visualization does not change the original data
-    dpls_copied = copy.deepcopy(simulation_data['dpls'])
-    net_copied = copy.deepcopy(simulation_data['net'])
+    _plot_out = analysis_config['plot_outputs'][plot_key]
```
jasmainak, line 61: usually leading underscore in the variable name means that it is "private" or for internal use. Is that what you mean here? If not, maybe worth just removing the underscores.

jasmainak:
I'll go ahead and merge, so you can iterate with the new design @chenghuzi . In the future, try to avoid the leading underscores unless you really mean it's a private function or variable.

jasmainak:
Thanks @chenghuzi !

hnn_core/gui/gui.py
chenghuzi: You're right I should remove the underscore there as it's safe to do so as it's already in a function. Will clean these sort of stuff later

PR #565, opened 2022-09-12
[MRG] Allow subplots overlay

chenghuzi:
image


hnn_core/viz.py
```diff
@@ -233,8 +246,10 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
         The SciPy function :func:`~scipy.signal.decimate` is used, which
         recommends values <13. To achieve higher decimation factors, a list of
         ints can be provided. These are applied successively.
-    color : tuple of float
+    color : tuple of float | str
         RGBA value to use for plotting. By default, 'k' (black)
+    sim_name : str | None
+        Prefix of dipole labels. Used when overlaying multiple simulations.
```
jasmainak, line 252: could we just call it `label` to adhere to matplotlib naming convention? It could default to `'average'`

jasmainak:
Let us know when ready!

chenghuzi:
> Let us know when ready!

The functionality is now finished. I'll add some tests soon to complete it

chenghuzi:
@jasmainak could you help to check what's going on in circleci? I tried to build the doc locally and it works fine while circleci keeps reporting errors. Once we have this figured out I think this pr will be ready to be merged

hnn_core/viz.py
```diff
@@ -6,9 +6,22 @@
 
 import numpy as np
 from itertools import cycle
+
+import matplotlib.colors as mc
```
jasmainak, line 10: This import is causing the trouble. I'm not sure why it doesn't get triggered in the other examples. You should "re-run job with ssh" and then try to figure what's going on

jasmainak: Actually, aren't matplotlib imports always nested? I think that's how we have done it in MNE in the past because of known issues. Probably it's not thread safe:

https://matplotlib.org/stable/users/faq/howto_faq.html#work-with-threads

chenghuzi: Bumping the matplotlib version up to >=3.5.3 seems solved the problem

hnn_core/gui/_viz_manager.py
```diff
@@ -149,19 +159,19 @@ def _update_ax(fig, ax, single_simulation, plot_type, plot_config):
...
-            io_buf = io.BytesIO()
-            _fig.savefig(io_buf, format='raw')
-            io_buf.seek(0)
-            img_arr = np.reshape(np.frombuffer(io_buf.getvalue(),
-                                               dtype=np.uint8),
-                                 newshape=(int(_fig.bbox.bounds[3]),
-                                           int(_fig.bbox.bounds[2]), -1))
-            io_buf.close()
-            ax.imshow(img_arr)
+            with plt.ioff():
```
jasmainak, line 162: why was this necessary?

hnn_core/viz.py
```diff
@@ -6,9 +6,23 @@
 
 import numpy as np
 from itertools import cycle
+
+import matplotlib
```
jasmainak, line 10: I would still nest it inside the function to avoid unexpected issues.

jasmainak (approved):
Other than minor nitpicks, looks good to me. I did not try but I trust that you did :)

hnn_core/gui/_viz_manager.py
chenghuzi: yes otherwise there will be another figure of the network object under the config tabs like this one:

image

hnn_core/viz.py
```diff
@@ -6,9 +6,23 @@
 
 import numpy as np
 from itertools import cycle
+
```
ntolley, line 9: 2 blank lines doesn't throw a flake8?

hnn_core/viz.py
```diff
@@ -6,9 +6,23 @@
...
+
+import colorsys
+
 from .externals.mne import _validate_type
 
 
+def _lighten_color(color, amount=0.5):
```
ntolley, line 16: this is pretty clever!

ntolley (approved):
Looks great @chenghuzi! Just played around with the GUI a bit and everything seems to be working as desired. The color cycling and label updates are quite nice too :)

hnn_core/dipole.py
```diff
@@ -448,6 +448,10 @@ def plot_psd(self, fmin=0, fmax=None, tmin=None, tmax=None, layer='agg',
             simulation.
         layer : str, default 'agg'
             The layer to plot. Can be one of 'agg', 'L2', and 'L5'
+        color : str or tuple or None
+            The line color of PSD
+        label : str or None
+            Line label for PSD
```
rythorpe, line 454: Just to maintain style consistency...
```suggestion
        color : str | tuple | None
            The line color of PSD
        label : str | None
            Line label for PSD
```

rythorpe (approved):
This all looks great @chenghuzi! Minus my one nitpicky suggestion, I'm ready to merge this.

hnn_core/viz.py
chenghuzi: yeah that's a bit weird...

hnn_core/gui/_viz_manager.py
```diff
@@ -114,7 +114,9 @@ def _update_ax(fig, ax, single_simulation, plot_type, plot_config):
 
     elif plot_type == 'PSD':
```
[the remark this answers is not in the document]
chenghuzi: Sorry I didn't get it. Do you mean to add docstrings about types that are allowed to be overlayed?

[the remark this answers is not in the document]
ntolley: Actually I'm realizing that this isn't necessary since you disable the add plot button. I originally thought there was logic in the code here specific to overlaid/non-overlaid plots but that isn't the case.

PR #570, opened 2022-11-21
[MRG] Adaptive colorbar for spectrogram plots & dynamic x-axis alignment

chenghuzi:
image

Now one can safely delete a spectrogram plot without changing the figure layout.
Notice this involves some changes to the original spectrogram viz function and we may have to make it non-default (or not?)

### DIFF
--- a/hnn_core/cell_response.py
+++ b/hnn_core/cell_response.py
@@ -322,7 +322,8 @@ class CellResponse(object):
 
         return spike_rates
 
-    def plot_spikes_raster(self, trial_idx=None, ax=None, show=True):
+    def plot_spikes_raster(self, trial_idx=None, ax=None, start_from_t0=True,
+                           show=True):
         """Plot the aggregate spiking activity according to cell type.
 
         Parameters
@@ -331,6 +332,8 @@ class CellResponse(object):
             Index of trials to be plotted. If None, all trials plotted.
         ax : instance of matplotlib axis | None
             An axis object from matplotlib. If None, a new figure is created.
+        start_from_t0 : bool
+            If True, hard set xlim left to 0.
         show : bool
             If True, show the figure.
 
@@ -340,7 +343,8 @@ class CellResponse(object):
             The matplotlib figure object.
         """
         return plot_spikes_raster(
-            cell_response=self, trial_idx=trial_idx, ax=ax, show=show)
+            cell_response=self, trial_idx=trial_idx, ax=ax,
+            start_from_t0=start_from_t0, show=show)
 
     def plot_spikes_hist(self, trial_idx=None, ax=None, spike_types=None,
                          show=True):
--- a/hnn_core/cell_response.py
+++ b/hnn_core/cell_response.py
@@ -322,8 +322,7 @@ class CellResponse(object):
 
         return spike_rates
 
-    def plot_spikes_raster(self, trial_idx=None, ax=None, start_from_t0=True,
-                           show=True):
+    def plot_spikes_raster(self, trial_idx=None, ax=None, show=True):
         """Plot the aggregate spiking activity according to cell type.
 
         Parameters
@@ -332,8 +331,6 @@ class CellResponse(object):
             Index of trials to be plotted. If None, all trials plotted.
         ax : instance of matplotlib axis | None
             An axis object from matplotlib. If None, a new figure is created.
-        start_from_t0 : bool
-            If True, hard set xlim left to 0.
         show : bool
             If True, show the figure.
 
@@ -343,8 +340,7 @@ class CellResponse(object):
             The matplotlib figure object.
         """
         return plot_spikes_raster(
-            cell_response=self, trial_idx=trial_idx, ax=ax,
-            start_from_t0=start_from_t0, show=show)
+            cell_response=self, trial_idx=trial_idx, ax=ax, show=show)
 
     def plot_spikes_hist(self, trial_idx=None, ax=None, spike_types=None,
                          show=True):
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -68,24 +68,6 @@ fig_templates = {
 }
 
 
-def _align_fig_xaxis(fig):
-    # align x axis for all plots whose x axis are time.
-    ax_align_x = []
-    for ax in fig.get_axes():
-        if hasattr(ax, 'align_x'):
-            if getattr(ax, 'align_x'):
-                ax_align_x.append(ax)
-
-    if len(ax_align_x) > 1:
-        ax_align_x[0].get_shared_x_axes().join(*ax_align_x)
-
-
-def _disable_align_fig_xaxis(fig):
-    for ax in fig.get_axes():
-        for aax in fig.get_axes():
-            ax.get_shared_x_axes().remove(aax)
-
-
 def _idx2figname(idx):
     return f"Figure {idx}"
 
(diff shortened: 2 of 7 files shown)


jasmainak:
Looks great! What remains to be done?

chenghuzi:
> Looks great! What remains to be done?

I'm not sure if we should make this mini colorbar the new default colorbar for the spectrogram.

jasmainak:
I'm okay with mini colorbar only for GUI and let's not touch the current default for hnn-core unless there is a compelling reason.

Can you match the x-axis of the two subplots? Otherwise the full benefit of the mini-colorbar is not realized ...

chenghuzi:
> I'm okay with mini colorbar only for GUI and let's not touch the current default for hnn-core unless there is a compelling reason.
> 
> Can you match the x-axis of the two subplots? Otherwise the full benefit of the mini-colorbar is not realized ...

Yeah I know this issue... the problem is that some plot types such as PSD could appear and alignment then will be broken. But I guess we can use some similar tricks to address this problem.

jasmainak:
Just use xlim = (0, tstop) in all plots that are relevant ... instead of sharex=True

chenghuzi:
> Just use xlim = (0, tstop) in all plots that are relevant ... instead of sharex=True

That may lead to some semi-hidden dots, e.g., in spikes, and aesthetically looks a little narrow... The latest commits provide another solution you can check that out

jasmainak:
@chenghuzi your solution is fine as such but it's not a [KISS](https://en.wikipedia.org/wiki/KISS_principle) solution. I worry about future developers ... @rythorpe already complained that the GUI code was hard to follow. Plus it only works for GUI.

Why not just `(0, tstop)` or `(-delta, tstop + delta)` if you need a little buffer to show spikes near tstop.

hnn_core/cell_response.py
```diff
@@ -322,7 +322,8 @@ def mean_rates(self, tstart, tstop, gid_ranges, mean_type='all'):
 
         return spike_rates
 
-    def plot_spikes_raster(self, trial_idx=None, ax=None, show=True):
+    def plot_spikes_raster(self, trial_idx=None, ax=None, start_from_t0=True,
```
jasmainak, line 325: I don't think you need a new parameter for this ... if users want, they can just do `ax.set_xlim` to change the limits.

rythorpe (approved):
I definitely think that adding the `colorbar_loc` feature at the level of `viz.plot_tfr_morlet()` will serve us well in the future. It *always* frustrates me how much space is wasted by the colorbar in a tight layout.

On the other hand, I'm not convinced of the `start_from_t0` feature. It seems like unnecessary complexity to me. Happy to be convinced otherwise though...

hnn_core/cell_response.py
rythorpe: Another option would be to use ticks instead of dots for the raster plot so that they are more visible if/when the x-axis is cut off at 0. I actually prefer ticks generally speaking.

hnn_core/cell_response.py
```diff
@@ -331,6 +332,8 @@ def plot_spikes_raster(self, trial_idx=None, ax=None, show=True):
             Index of trials to be plotted. If None, all trials plotted.
         ax : instance of matplotlib axis | None
             An axis object from matplotlib. If None, a new figure is created.
+        start_from_t0 : bool
```
ntolley, line 335: I suppose I'm not sure I understand when this shouldn't be true, do the plots look weird or get cut off?

hnn_core/gui/_viz_manager.py
```diff
@@ -68,6 +68,24 @@
...
+def _align_fig_xaxis(fig):
+    # align x axis for all plots whose x axis are time.
+    ax_align_x = []
+    for ax in fig.get_axes():
+        if hasattr(ax, 'align_x'):
+            if getattr(ax, 'align_x'):
+                ax_align_x.append(ax)
+
+    if len(ax_align_x) > 1:
+        ax_align_x[0].get_shared_x_axes().join(*ax_align_x)
```
ntolley, line 80: To make sure I understand, this queries the xlimits from the first axis in the figure?

hnn_core/gui/_viz_manager.py
```diff
@@ -229,7 +256,16 @@ def _plot_on_axes(b, widgets_simulation, widgets_plot_type,
...
                 existing_plots, add_plot_button):
+    # clear all potential linked axes before any modifications.
+    _disable_align_fig_xaxis(fig)
     ax.clear()
+    delattr(ax, 'align_x')
+
+    # remove attached colorbar if exists
+    if hasattr(fig, f'_cbar-ax-{id(ax)}'):
+        getattr(fig, f'_cbar-ax-{id(ax)}').ax.remove()
+        delattr(fig, f'_cbar-ax-{id(ax)}')
```
ntolley, line 267: Kind of curious what happens if you don't delete it, does it just maintain a reference to the missing colorbar?

hnn_core/viz.py
```diff
@@ -610,6 +615,8 @@ def plot_tfr_morlet(dpl, freqs, *, n_cycles=7., tmin=None, tmax=None,
         The name of a matplotlib colormap, e.g., 'viridis'. Default: 'inferno'
     colorbar : bool
         If True (default), adjust figure to include colorbar.
+    colorbar_loc: str or None
+        The location of colorbar. Set to right side if None.
```
ntolley, line 619: Are the only options `'inner'` and `None`? If that's the case it seems like this should just be a boolean for placing it inside or outside.

hnn_core/gui/_viz_manager.py
chenghuzi: I guess so. Deleting is more for clearness here and the chance that some other variable takes that memory block should be neglectable.

hnn_core/gui/_viz_manager.py
chenghuzi: yes

hnn_core/cell_response.py
chenghuzi: Suppose the first spike occurs at t=50, matplotlib will probably set the xlim to somewhere close but less than 50 like 40. But one may want to plot the whole range from the very beginning of the simulation even if it's not aligned with other subplots, which I guess is the original motivation of `ax.set_xlim(left=0)`. And if we always turn on this option, the alignment will be influenced and all other subplots will start from t=0. Since I personally tend to leave some space for both sides for all plots so I added this behavior.  It indeed looks a bit weird and we can safely remove it if we adopt Mainak's proposal.

rythorpe:
What's the status of this PR? Are we ready to rebase and merge?

chenghuzi:
> What's the status of this PR? Are we ready to rebase and merge?

We're almost there... Just one last piece: the raster marker type. Here's the new one with vertical lines

image

ntolley:
love the change @chenghuzi !

Much more in line with actual published raster plots

hnn_core/viz.py
```diff
@@ -495,23 +495,23 @@ def plot_spikes_raster(cell_response, trial_idx=None, ax=None, show=True):
...
             gid_time = spike_times[spike_gids == gid]
             cell_type_times.append(gid_time)
-            cell_type_ypos.append(np.repeat(ypos, len(gid_time)))
+            cell_type_ypos.append(ypos)
             ypos = ypos - 1
 
         if cell_type_times:
-            cell_type_times = np.concatenate(cell_type_times)
-            cell_type_ypos = np.concatenate(cell_type_ypos)
+            events.append(
```
jasmainak, line 509: is `events` used later?

jasmainak:
Looks clean, do you want to update `whats_new.rst` @chenghuzi ? Otherwise happy to merge.

chenghuzi:
> Looks clean, do you want to update `whats_new.rst` @chenghuzi ? Otherwise happy to merge.

I think we're good! `whats_new.rst` will be updated with other features together

jasmainak:
Thanks @chenghuzi !

PR #599, opened 2023-02-19
[MRG] Fix wrong slider type

### LINKED ISSUE #598: Error in the new GUI
We used both `FloatText` and `FloatLogSlider` for connectivities. However, we didn't notice that some connectivity values of 0 would be forced to the minimal value of `FloatLogSlider`, which is not actually 0 due to its logarithm nature. See the attached screenshot for an example.

Xnip2023-02-19_18-37-25
I'll do a quick fix to this based on our master branch by replacing the `FloatLogSlider` to `FloatSlider`

chenghuzi:
Closes #598

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -781,9 +781,9 @@ def _get_rhythmic_widget(name, tstop_widget, layout, style, location,
     burst_std = BoundedFloatText(
         value=default_data['burst_std'], description='Burst std dev (Hz)',
         min=0, max=1e6, **kwargs)
-    repeats = BoundedFloatText(
-        value=default_data['repeats'], description='Repeats', min=0, max=1e6,
-        **kwargs)
+    repeats = BoundedIntText(
+        value=default_data['repeats'], description='Repeats', min=0,
+        max=int(1e6), **kwargs)
     seedcore = IntText(value=default_data['seedcore'],
                        description='Seed',
                        **kwargs)
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -14,9 +14,8 @@ from pathlib import Path
 
 from IPython.display import IFrame, display
 from ipywidgets import (HTML, Accordion, AppLayout, BoundedFloatText,
-                        BoundedIntText, Button, Dropdown, FileUpload,
-                        FloatText, HBox, IntText, Layout, Output, RadioButtons,
-                        Tab, Text, VBox)
+                        BoundedIntText, Button, Dropdown, FileUpload, VBox,
+                        HBox, IntText, Layout, Output, RadioButtons, Tab, Text)
 from ipywidgets.embed import embed_minimal_html
 
 import hnn_core
@@ -206,10 +205,11 @@ class HNNGUI:
         self.simulation_data = defaultdict(lambda: dict(net=None, dpls=list()))
 
         # Simulation parameters
-        self.widget_tstop = FloatText(value=170, description='tstop (ms):',
-                                      disabled=False)
-        self.widget_dt = FloatText(value=0.025, description='dt (ms):',
-                                   disabled=False)
+        self.widget_tstop = BoundedFloatText(
+            value=170, description='tstop (ms):', min=0, max=1e6,
+            disabled=False)
+        self.widget_dt = BoundedFloatText(
+            value=0.025, description='dt (ms):', min=0, max=10, disabled=False)
         self.widget_ntrials = IntText(value=1, description='Trials:',
                                       disabled=False)
         self.widget_simulation_name = Text(value='default',
@@ -669,9 +669,10 @@ def _get_connectivity_widgets(conn_data):
     style = {}
     sliders = list()
     for receptor_name in conn_data.keys():
-        w_text_input = FloatText(value=conn_data[receptor_name]['weight'],
-                                 disabled=False, continuous_update=False,
-                                 step=0.01, description="weight", style=style)
+        w_text_input = BoundedFloatText(
+            value=conn_data[receptor_name]['weight'], disabled=False,
+            continuous_update=False, min=0, max=1e6, step=0.01,
+            description="weight", style=style)
 
         conn_widget = VBox([
             HTML(value=f"""<p>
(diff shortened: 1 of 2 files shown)


hnn_core/gui/gui.py
```diff
@@ -674,15 +674,15 @@ def _get_connectivity_widgets(conn_data):
...
-                                  min=-5, max=1, step=0.2,
-                                  description=" ",
-                                  disabled=False,
-                                  continuous_update=False,
-                                  orientation='horizontal',
-                                  readout=False,
-                                  readout_format='.2e',
-                                  style=style)
+        w_slider = FloatSlider(value=conn_data[receptor_name]['weight'],
+                               min=-1, max=1, step=0.01,
```
jasmainak, line 678: not sure if this is the right fix. This will prevent user from making finer adjustments.

Also I guess weights can't be negative.

Wondering if the slider and the stepper should have different step sizes ... large logarithmic step size for the slider and small step size for the stepper arrows for finer adjustment?

any opinion @rythorpe @ntolley ?

rythorpe: ^This sounds like a fair suggestion to me. This will be consistent across local network and drive connectivity right?

chenghuzi: It's not negative values that break the thing. Sometimes we have zero values. With `FloatLogSlider` they are all strictly >0. That actually will cause problems in some settings.

jasmainak: as we talked in the meeting today, let's just get rid of the sliders.

jasmainak:
something strange with the way the text box works ... pressing the uparrow does this:

image

the step size is too large. Could we just disable the arrows?

chenghuzi:
> something strange with the way the text box works ... pressing the uparrow does this:
> 
> image
> 
> the step size is too large. Could we just disable the arrows?

how about changing this to 0.01?

hnn_core/gui/gui.py
```diff
@@ -671,25 +671,12 @@ def _get_connectivity_widgets(conn_data):
...
-                                  description=" ",
-                                  disabled=False,
-                                  continuous_update=False,
-                                  orientation='horizontal',
-                                  readout=False,
-                                  readout_format='.2e',
-                                  style=style)
-
-        link((w_slider, 'value'), (w_text_input, 'value'))
+                                 step=0.01, description="weight", style=style)
```
jasmainak, line 674: shouldn't it be bounded to disable negative values?

chenghuzi: No, if we don't have a good upper bound for these variables. Because with `BoundedFloatText` we'll have a default `max` value to 100 even if we only specify the `min` to 0.

jasmainak: just use a large value for `max` like 1e6 or something

chenghuzi: it's now updated. And I also updated some other `FloatText`s to `BoundedFloatText`s. Please help me to check if these changes are also valid or we need to move some of them back to `FloatText`.

hnn_core/gui/gui.py
```diff
@@ -206,10 +205,11 @@ def __init__(self, theme_color="#8A2BE2",
...
         # Simulation parameters
-        self.widget_tstop = FloatText(value=170, description='tstop (ms):',
-                                      disabled=False)
-        self.widget_dt = FloatText(value=0.025, description='dt (ms):',
-                                   disabled=False)
+        self.widget_tstop = BoundedFloatText(
+            value=170, description='tstop (ms):', min=0, max=1e6,
+            disabled=False)
+        self.widget_dt = BoundedFloatText(
+            value=0.025, description='dt (ms):', min=0, max=10, disabled=False)
```
rythorpe, line 212: What are the step sizes here?

hnn_core/gui/gui.py
```diff
@@ -778,27 +763,27 @@ def _get_rhythmic_widget(name, tstop_widget, layout, style, location,
...
-                        **kwargs)
+    burst_rate = BoundedFloatText(
+        value=default_data['burst_rate'], description='Burst rate (Hz)',
+        min=0, max=1e6, **kwargs)
+    burst_std = BoundedFloatText(
+        value=default_data['burst_std'], description='Burst std dev (Hz)',
+        min=0, max=1e6, **kwargs)
+    repeats = BoundedFloatText(
+        value=default_data['repeats'], description='Repeats', min=0, max=1e6,
+        **kwargs)
```
rythorpe, line 786: Does this need to be of type `int`?

hnn_core/gui/gui.py
```diff
@@ -911,12 +894,12 @@ def _get_evoked_widget(name, layout, style, location, data=None,
...
-                   **kwargs)
-    sigma = FloatText(value=default_data['sigma'],
-                      description='Std dev time:',
-                      **kwargs)
+    mu = BoundedFloatText(
+        value=default_data['mu'], description='Mean time:', min=0, max=1e6,
+        **kwargs)
+    sigma = BoundedFloatText(
+        value=default_data['sigma'], description='Std dev time:', min=0,
+        max=1e6, **kwargs)
```
rythorpe, line 902: What's the step size here?

hnn_core/gui/gui.py
chenghuzi: I don't have a good sense for it. Maybe we can set it to 0.025 too?

hnn_core/gui/gui.py
rythorpe: I reviewed the bounds you set in the `BoundedFloatText` objects, and everything looks good to me aside from my few comments below.

hnn_core/gui/gui.py
chenghuzi: I think so.

hnn_core/gui/gui.py
chenghuzi: the default value is 1 I think

chenghuzi: Let's set them to 0.01 for now as they're time related. In the future we can adjust further if needed.

hnn_core/gui/gui.py
chenghuzi: Let's set the `dt` step to 0.01 and  `tstop` step size to 1 as it's more coarse

hnn_core/gui/gui.py
chenghuzi: For some step sizes, we don't have good estimates yet. So my strategy is to set the step size to 1 for things like `tsop` that involve ranges, and to 0.01 for other time-sensitive variables. This way, it will be easy to adjust them in the future if necessary. Let's stick with what we have now, as our main purpose in this PR is to change the input type.

hnn_core/gui/gui.py
```diff
@@ -1201,7 +1185,7 @@ def _init_network_from_widgets(params, dt, tstop, single_simulation_data,
                 single_simulation_data['net'].connectivity[conn_idx][
                     'nc_dict']['A_weight'] = vbox.children[1].value
                 single_simulation_data['net'].connectivity[conn_idx][
-                    'probability'] = vbox.children[3].value
+                    'probability'] = vbox.children[2].value
```
ntolley, line 1188: Not clear to me what change in the code here requires changing the index here?

ntolley: Is it because the text box is taking the place of the slider?

ntolley:
@chenghuzi just read through and tested everything. Looks great!

PR #621, opened 2023-03-19
[MRG] DOC: document retrieve params after optimization

carolinafernandezp:


jasmainak:
Looks great @carolinafernandezp ! Do you mind sharing a screenshot of the output when running this example? It is not run on CircleCI since it is a long-running example

Also, you can edit your description to [close the linked issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) when this PR is merged.

Finally -- if you need to update your PR, you can simply push to the same branch and the PR will get updated. No need to make a new PR. Even if you end having a rebase mess, it is always possible to create a fresh branch with the same name and force push so that the old PR gets updated.

rythorpe:
Was the original issue about creating a plot function for visualizing parameter changes, or just printing to stdout?

carolinafernandezp:
> Looks great @carolinafernandezp ! Do you mind sharing a screenshot of the output when running this example? It is not run on CircleCI since it is a long-running example
> 
> Also, you can edit your description to [close the linked issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) when this PR is merged.
> 
> Finally -- if you need to update your PR, you can simply push to the same branch and the PR will get updated. No need to make a new PR. Even if you end having a rebase mess, it is always possible to create a fresh branch with the same name and force push so that the old PR gets updated.

Here is the screenshot:
image

carolinafernandezp:
> Was the original issue about creating a plot function for visualizing parameter changes, or just printing to stdout?

Here is a link to the Discussion (#578), we spoke about inspecting the net.connectivity attribute so I assumed stdout.

rythorpe:
Got it - fair enough. You currently show how to access a few post-optimization parameters, but my only suggestion here would be to highlight a comparison from pre-opt to post-opt parameters.

carolinafernandezp:
> 

How does this look?

image

rythorpe:
This is better, but I'm concerned that this might still be confusing for some users who aren't familiar with which parameters where optimized and which were held constant (e.g., the connection probability for `evprox1 -> L2_pyramidal` is held constant). WDYT @jasmainak @ntolley? Should there be a more explicit comparison that shows only the parameters that have been optimized over, or is the current comparison that advertises a more fundamental and simple use of the API more important?

rythorpe:
> I think we ultimately need a proper plotting function : #423 . I'd be okay with this much documentation for now ... but the tests seems to be failing

Okay, let's just move forward with this as it is. The tests are failing due to `test_mpi_failure`, which seems to be popping up rather frequently now, albeit at random. I'm guessing it has to due with how the CI resource allocation is handling the `OMPI_MCA_btl` environmental variable which has nothing to do with this PR.

rythorpe (commented):
@carolinafernandezp go ahead and update your code here so it matches the screenshot you shared earlier.

carolinafernandezp:
> @carolinafernandezp go ahead and update your code here so it matches the screenshot you shared earlier.

Done!

examples/howto/optimize_evoked.py
```diff
@@ -93,3 +91,33 @@
...
+# Let's compare to the initial dynamic properties and weight.
+dynamics_initial = net.external_drives['evdist1']['dynamics']
+
+conn_indices = pick_connection(net=net, src_gids='evprox1', target_gids='L2_pyramidal',
+    loc='proximal', receptor='ampa')
+conn_idx = conn_indices[0]
+weight_initial = net.connectivity[conn_idx]
+
+print("Initial dynamic properties: ", dynamics_initial)
+print("\Initial weight: ", weight_initial)
```
rythorpe, line 123:
```suggestion
print("\nInitial weight: ", weight_initial)
```

examples/howto/optimize_evoked.py
```diff
@@ -93,3 +91,33 @@
 best_dpl.plot(ax=axes[0], layer='agg', show=False, color='tab:green')
 axes[0].legend(['experimental', 'initial', 'optimized'])
 net_opt.cell_response.plot_spikes_hist(ax=axes[1])
+
+###############################################################################
+# Finally, let's explore which parameters were changed to cause the 
+# improved dipole fit. 
+# As an example, we will look at the new dynamics of the first proximal drive, 
+# as well as the synaptic weight of its layer 2/3, pyramidal AMPA receptors.
```
rythorpe, line 99:
```suggestion
# as well as the synaptic weight of its layer 2/3 pyramidal AMPA receptors.
```
Nitpick: is this comma supposed to be here?

rythorpe:
@carolinafernandezp I was checking the built docs and found to minor potential typos. LMK what you think, and as soon as they're resolved let's set the PR title to "[MRG]..." so we can rebase and merge!

carolinafernandezp:
> @carolinafernandezp I was checking the built docs and found to minor potential typos. LMK what you think, and as soon as they're resolved let's set the PR title to "[MRG]..." so we can rebase and merge!

Thank you for finding those, I made the changes!

rythorpe:
Looks good! Thanks @carolinafernandezp!

PR #635, opened 2023-04-10
[MRG] Add tonic inputs

chenghuzi:
As titled.

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -1151,6 +1151,8 @@ def add_drive_tab(params, drives_out, drive_widgets, drive_boxes, tstop,
     drive_specs = _extract_drive_specs_from_hnn_params(
         params, cellname_list, legacy_mode=net._legacy_mode)
     bias_specs = _extract_bias_specs_from_hnn_params(params, cellname_list)
+    bias_is_present = len(bias_specs['tonic']) > 0
+    logger.info(f'bias_specs:{bias_specs}')
     # clear before adding drives
     drives_out.clear_output()
     while len(drive_widgets) > 0:
@@ -1160,8 +1162,12 @@ def add_drive_tab(params, drives_out, drive_widgets, drive_boxes, tstop,
     drive_names = sorted(drive_specs.keys())
     for idx, drive_name in enumerate(drive_names):  # order matters
         specs = drive_specs[drive_name]
-        should_render = idx == (len(drive_names) -
-                                1) and len(bias_specs['tonic']) > 0
+        # should_render = idx == (len(drive_names) -
+        #                         1) and len(bias_specs['tonic']) > 0
+        if bias_is_present:
+            should_render = False
+        else:
+            should_render = idx == len(drive_names) - 1
         add_drive_widget(
             specs['type'].capitalize(),
             drive_boxes,
@@ -1180,14 +1186,15 @@ def add_drive_tab(params, drives_out, drive_widgets, drive_boxes, tstop,
             event_seed=specs['event_seed'],
         )
 
-    logger.info(f'Adding bias_specs from params: {bias_specs}')
-    should_render = True
-    add_drive_widget(
-        'Tonic', drive_boxes, drive_widgets, drives_out, tstop, None,
-        layout=layout, prespecified_drive_name=drive_name,
-        prespecified_drive_data=bias_specs['tonic'],
-        render=should_render, expand_last_drive=False,
-    )
+    if bias_is_present:
+        logger.info(f'Adding bias_specs from params: {bias_specs}')
+        should_render = True
+        add_drive_widget(
+            'Tonic', drive_boxes, drive_widgets, drives_out, tstop, None,
+            layout=layout, prespecified_drive_name=drive_name,
+            prespecified_drive_data=bias_specs['tonic'],
+            render=should_render, expand_last_drive=False,
+        )
 
 
 def load_drive_and_connectivity(params, log_out, drives_out,
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -1183,17 +1183,10 @@ def add_drive_tab(params, drives_out, drive_widgets, drive_boxes, tstop,
     logger.info(f'Adding bias_specs from params: {bias_specs}')
     should_render = True
     add_drive_widget(
-        'Tonic',
-        drive_boxes,
-        drive_widgets,
-        drives_out,
-        tstop,
-        None,
-        layout=layout,
-        prespecified_drive_name=drive_name,
+        'Tonic', drive_boxes, drive_widgets, drives_out, tstop, None,
+        layout=layout, prespecified_drive_name=drive_name,
         prespecified_drive_data=bias_specs['tonic'],
-        render=should_render,
-        expand_last_drive=False,
+        render=should_render, expand_last_drive=False,
     )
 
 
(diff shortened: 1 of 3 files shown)


ntolley:
@chenghuzi I think a tonic drive is getting added that should be skipped:
image

Do you think you could follow the logic in #619 to filter out invalid tonic drives? I think the culprit function is `_extract_drive_specs_from_hnn_params()`

chenghuzi:
> _extract_drive_specs_from_hnn_params

sure, but I guess it's `_extract_bias_specs_from_hnn_params`? Either way I'll check and let you know asap

chenghuzi:
> @chenghuzi I think a tonic drive is getting added that should be skipped: image
> 
> Do you think you could follow the logic in #619 to filter out invalid tonic drives? I think the culprit function is `_extract_drive_specs_from_hnn_params()`

It should be fixed now. I didn’t open a new PR as it’s a minor modification, and adding just a filter condition in ﻿`_extract_bias_specs_from_hnn_params ` seems to work.

hnn_core/gui/gui.py
```diff
@@ -1000,8 +1051,18 @@ def add_drive_widget(drive_type, drive_boxes, drive_widgets, drives_out,
...
+            drive, drive_box = _get_tonic_widget(
+                name,
+                layout,
+                style,
+                data=prespecified_drive_data
+            )
+
         if drive_type in [
-                'Evoked', 'Poisson', 'Rhythmic', 'Bursty', 'Gaussian'
+                'Evoked', 'Poisson', 'Rhythmic', 'Bursty', 'Gaussian', 'Tonic'
```
rythorpe, line 1065: Probably not related to this PR, but do we still need references to the `'Gaussian'` drive type anywhere in the GUI?

hnn_core/gui/gui.py
```diff
@@ -1094,8 +1162,12 @@ def add_drive_tab(params, drives_out, drive_widgets, drive_boxes, tstop,
     drive_names = sorted(drive_specs.keys())
     for idx, drive_name in enumerate(drive_names):  # order matters
         specs = drive_specs[drive_name]
-        should_render = idx == (len(drive_names) - 1)
-
+        # should_render = idx == (len(drive_names) -
+        #                         1) and len(bias_specs['tonic']) > 0
```
rythorpe, line 1166: Delete?

hnn_core/gui/gui.py
```diff
@@ -943,6 +946,54 @@ def _get_evoked_widget(name, layout, style, location, data=None,
...
+            default_data.update(data[cell_type])
+        amplitudes[cell_type] = BoundedFloatText(
+            value=copy.deepcopy(default_data['amplitude']),
+            description=cell_type, min=0, max=1e6, step=0.01, **kwargs)
+        start_times[cell_type] = BoundedFloatText(
+            value=copy.deepcopy(default_data['t0']), description=cell_type,
+            min=0, max=1e6, step=0.01, **kwargs)
+        stop_times[cell_type] = BoundedFloatText(
+            value=copy.deepcopy(default_data['tstop']), description=cell_type,
+            min=-1, max=1e6, step=0.01, **kwargs)
```
rythorpe, line 975: Is there a reason why we set independent start/stop times for each cell type?

chenghuzi: you mean the range? or the data values?

hnn_core/gui/gui.py
chenghuzi: yes here

https://github.com/chenghuzi/hnn-core/blob/4cf3dc3b7b33fff2c96280fb78b0bbfe652eabd7/hnn_core/gui/gui.py#L1042

hnn_core/gui/gui.py
rythorpe: Like, I can currently add a single tonic bias that starts and stops at different times depending on the cell type. If someone wants to set different dynamics for different cell types, they should just create separate baises that correspond one-to-one with the different cell types.

hnn_core/tests/test_gui.py
```diff
@@ -66,6 +66,9 @@ def test_gui_upload_params():
     gui._simulate_upload_connectivity(file2_url)
     # now connectivity is refreshed.
     assert gui.connectivity_widgets[0][0].children[1].value == 0.01
+
+    # We use -2 instead of -1 here as now the last one is the tonic inputs.
+    # changed back to -1 as tonic invalid inputs are removed.
```
jasmainak, line 71: ? I have no idea what this means :-)

hnn_core/gui/gui.py
jasmainak: agree about that ... the GUI should reflect what the `hnn-core` API says, which is one cell-type per bias

hnn_core/gui/gui.py
jasmainak: humm ... but I don't get how you can have a Gaussian drive type. It's not an option in the GUI

hnn_core/gui/gui.py
```diff
@@ -1229,66 +1309,82 @@ def _init_network_from_widgets(params, dt, tstop, single_simulation_data,
...
+                    weights_ampa=weights_ampa,
+                    weights_nmda=weights_nmda,
+                    synaptic_delays=synaptic_delays,
+                    event_seed=drive['seedcore'].value)
+
+        elif drive['type'] == 'Tonic':
+            cell_types = list(drive['amplitude'].keys())
+            assert set(cell_types) == set(drive['t0'].keys()) == set(
+                drive['tstop'].keys())
+            for cell_type in cell_types:
```
jasmainak, line 1378: as we discussed above, I think we don't need this external loop ... it will greatly simplify your code and the interface. Just a dropdown for selecting the cell_type is needed

hnn_core/gui/gui.py
rythorpe: Wait, a bias is only supposed to target a single cell type?

hnn_core/tests/test_gui.py
chenghuzi: never mind, these are for myself and we can remove them now

hnn_core/gui/gui.py
chenghuzi: I kind of forgot the details. could you correct me if I am wrong? are you saying that the bias has the same value for all cell types? And thus we don't need to let the user to set different values for different cell types

rythorpe: I think we decided that the start and stop times should be the same for all cell types targetted by the tonic drive, but that this needs to be changed at the API-level within `Network.add_tonic_bias()`.

chenghuzi: For now I think we can simply link those components so that they look all the same. This is the most straightforward solution and visually reinforces to the user they are meant to be the same, rather than implying an error on our end.

chenghuzi: Check the latest commit, we just linked them alltogether

rythorpe: Why not just fix `add_tonic_bias()` so that we don't have to revert all of this later?

rythorpe: It should probably be it's own PR but shouldn't take too long. If I get a chance this week, I'm happy to open a PR.

jasmainak: I agree with Ryan that we should add an option to specify the tonic biases using a dictionary so that the logic of network instantiating is not in the GUI. The GUI should be a thin layer on top of the computational core

rythorpe: Unfortunately I never got around to this.

Yes, the idea was to be able to set which cell types are targeted in `Network.add_tonic_bias()` (similar to the way we set which cells are targeted by setting weights in `Network.add_evoked_drive()`). The best way to do with would probably be to remove the `cell_type` arg and modify the `amplitude` arg to be dict that maps amplitude to a given cell type key.

jasmainak: we should certainly fix the API before adding to the GUI

ntolley:
I think this one can be closed!

PR #636, opened 2023-04-10
[MRG] Add RMSE

chenghuzi:
As titled. 

Demo (updated):
image

### DIFF
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -12,7 +12,9 @@ import numpy as np
 from IPython.display import display
 from ipywidgets import (Box, Button, Dropdown, FloatText, HBox, Label, Layout,
                         Output, Tab, VBox, link)
+from scipy import signal
 
+from hnn_core.dipole import average_dipoles
 from hnn_core.gui._logging import logger
 from hnn_core.viz import plot_dipole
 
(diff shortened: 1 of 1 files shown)


chenghuzi:
@jasmainak maybe we can finish this first and turn that histogram stuff into another PR?

chenghuzi:
We have discussed adding the feature of a layered histogram into the default figure. This will be a separate PR. For now, maybe it would be better for us to merge this one first.

hnn_core/gui/_viz_manager.py
```diff
@@ -87,6 +108,54 @@ def _figname2idx(fname):
     return int(fname.split(" ")[-1])
 
 
+def _calculate_rmse(dpl, dpl_target, tstart, tstop, weights=None):
```
jasmainak, line 111: don't we have a function for this already?

https://github.com/jonescompneurolab/hnn-core/blob/cdd4050f187da4bc84c2869952a8fd54ae1e70dd/hnn_core/dipole.py#L172

hnn_core/gui/_viz_manager.py
```diff
@@ -238,7 +336,40 @@ def _plot_on_axes(b, widgets_simulation, widgets_plot_type,
...
+        # plot the target dipole.
+        target_plot_config = copy.deepcopy(plot_config)
+        # disable scaling for the target dipole.
+        target_plot_config['dipole_scaling'] = 1.
+
+        # plot the target dipole.
+        target_dpl_processed = _update_ax(
+            fig, ax, target_sim, target_sim_name, plot_type,
+            target_plot_config)[0]  # we assume there is only one dipole.
+        try:
```
jasmainak, line 358: I will repeat like a parrot ... avoid `try` / `except` when possible

hnn_core/gui/_viz_manager.py
chenghuzi: oh no... I didn't realize that

hnn_core/gui/_viz_manager.py
```diff
@@ -238,7 +287,41 @@ def _plot_on_axes(b, widgets_simulation, widgets_plot_type,
...
+
+    # If target_simulations is not None and we are plotting a dipole,
+    # we need to plot the target dipole as well.
+    if target_simulations.value in data['simulations'].keys(
+    ) and plot_type == 'current dipole':
+
+        target_sim_name = target_simulations.value
+        target_sim = data['simulations'][target_sim_name]
+        # plot the target dipole.
+        target_plot_config = copy.deepcopy(plot_config)
```
jasmainak, line 301: is the copy necessary?

jasmainak:
I just gave this a shot but I'm confused how exactly to use it:

image

Target data: None means what? I can't figure out what to compare to what. Can you share what you see and/or what you did?

jasmainak:
Also, some of the loaded data starts to appear in the simulations ... that's a bit confusing:

image

I think you should maintain separate lists of simulations and loaded data

hnn_core/gui/_viz_manager.py
chenghuzi: yeah this might be redundant, just habitually copied a dict...

chenghuzi:
> I just gave this a shot but I'm confused how exactly to use it:
> 
> image
> 
> Target data: None means what? I can't figure out what to compare to what. Can you share what you see and/or what you did?

`None` means there's no comparison in the current plot. It's not every time that the user want to compare some simulations against others. Without this null choice, the system will use the first value in options to compare.

chenghuzi:
> Also, some of the loaded data starts to appear in the simulations ... that's a bit confusing:
> 
> image
> 
> I think you should maintain separate lists of simulations and loaded data

I don't think so, we discussed this before and agreed on the current solution, i.e., to put everything under the same structure as essentially they are the same except that there's no `Net` objects in recorded data. But I renamed a bit in the latest commit to make it more clear to the user.

rythorpe:
Is this ready to go or were there more changes to be made after yesterday's mtg?

ntolley (approved):
Just tested this and everything looks great to me! Although I realized that we might consider adding example data to this repository like there is for the original GUI

ntolley:
@chenghuzi happy to merge whenever you give the go ahead

hnn_core/gui/_viz_manager.py
```diff
@@ -217,9 +244,31 @@ def _dynamic_rerender(fig):
 
 
 def _plot_on_axes(b, widgets_simulation, widgets_plot_type,
+                  target_simulations,
                   spectrogram_colormap_selection, dipole_smooth,
                   max_spectral_frequency, dipole_scaling, widgets, data,
                   fig_idx, fig, ax, existing_plots):
+    """Plotting different types of data on the given axes.
+    Now this function is also responsible for comparing multiple simulations,
```
jasmainak, line 252: I like that you started documenting this but could we document this in standard numpy format?

chenghuzi: updated in the latest commit

chenghuzi:
> @chenghuzi happy to merge whenever you give the go ahead

LGTM

hnn_core/gui/_viz_manager.py
```diff
@@ -334,9 +426,19 @@ def _get_ax_control(widgets, data, fig_idx, fig, ax):
...
+
+    def _on_target_comparison_change(new_target_name):
+        return target_comparison_change(new_target_name, simulation_selection,
+                                        data)
+
+    def _on_plot_type_change(new_plot_type):
+        return plot_type_coupled_change(new_plot_type, target_data_selection)
 
     simulation_selection.observe(_on_sim_data_change, 'value')
+    target_data_selection.observe(_on_target_comparison_change, 'value')
```
jasmainak, line 455: do you need this since `target_comparison_change` just contains a `pass` ?

jasmainak:
sorry this was an open tab on my browser ... just realized it has been merged, but feel free to clarify my question :)

hnn_core/gui/_viz_manager.py
chenghuzi: You're right, I didn't notice that. Actually let me create an issue for this and a related topic we discussed last time: to separate the simulation data and recorded loaded data via flags in the data dict. That will be a quickfix

PR #647, opened 2023-05-18
[MRG] Clean up net.cell_response

ntolley:
#644 has brought to light a few bugs with how the CellResponse object is instantiated. Mainly trying to clean up the logic here.

Also removing the cell response indexing that nobody used or knows exists :wink:


hnn_core/cell_response.py
```diff
@@ -75,13 +75,17 @@ class CellResponse(object):
     """
 
     def __init__(self, spike_times=None, spike_gids=None, spike_types=None,
-                 times=None, cell_type_names=None):
+                 times=None, cell_type_names=None, vsec=None, isec=None):
```
jasmainak, line 78: write method and `read_cell_response` to be updated as well

ntolley:
@jasmainak I've removed the `__get_item__()` functionality as I think it added a lot of complexity that was until now, mostly untested and documented.

hnn_core/cell_response.py
```diff
@@ -138,75 +142,9 @@ def __eq__(self, other):
                        for trial in other._spike_times]
         return (times_self == times_other and
                 self._spike_gids == other._spike_gids and
-                self._spike_types == other._spike_types)
-
-    def __getitem__(self, gid_item):
```
jasmainak, line 143: do we need to document this in `whats_new.rst` as an API change?

hnn_core/cell_response.py
jasmainak: you need to document `vsec` and `isec` ?

hnn_core/cell_response.py
jasmainak: Pinging @ntolley ... I see the PR is set to MRG but it's waiting for response to this comment?

ntolley: I'm falling behind :smile: 

The main reason for getting rid of this functionality is that I think it adds a lot of complexity and maintenance difficulties for not much payoff. This was added in the early days of my development, so it's really not documented

hnn_core/cell_response.py
ntolley: this falls under the network saving PR now so no longer needed to update here

ntolley:
@rythorpe @jasmainak @katduecker this is ready to go! The main goal was to remove the overly complicated cell_response indexing functionality

hnn_core/cell_response.py
```diff
@@ -137,77 +146,11 @@ def __eq__(self, other):
         return (times_self == times_other and
                 self._spike_gids == other._spike_gids and
                 self._spike_types == other._spike_types and
+                self._vsec == other._vsec and
+                self._isec == other._isec and
                 self.vsec == other.vsec and
                 self.isec == other.isec)
```
rythorpe, line 152: Do we still need these lines a well?

hnn_core/cell_response.py
```diff
@@ -35,6 +35,11 @@ class CellResponse(object):
     cell_type_names : list
         List of unique cell type names that are explicitly modeled in the
         network
+    vsec : list (n_trials,) of dict, shape
```
rythorpe, line 38:
```suggestion
    vsec : list of dict, shape (n_trials,)
```
There are a few entries above that need a similar wording fix.

jasmainak:
looks good @ntolley 

can you update whats_new.rst?

hnn_core/cell_response.py
ntolley: We could just remove "shape"? and leave the length next to the appropriate descriptor

hnn_core/cell_response.py
```diff
@@ -90,6 +95,10 @@ def __init__(self, spike_times=None, spike_gids=None, spike_types=None,
             spike_types = list()
         if times is None:
             times = list()
+        if vsec is None:
+            vsec = list()
+        if isec is None:
+            isec = list()
```
katduecker, line 101: Maybe I'm missing something but in line 129 following we set self._vsec = list(), are 98 and following therefore redundant?
Also, do we need a vsec and isec input as I don't think we use it?

hnn_core/viz.py
```diff
@@ -465,7 +465,10 @@ def plot_spikes_hist(cell_response, trial_idx=None, ax=None, spike_types=None,
     elif isinstance(color, list):
         color_cycle = cycle(color)
 
-    bins = np.linspace(0, spike_times[-1], 50)
+    if len(cell_response.times) > 0:
+        bins = np.linspace(0, cell_response.times[-1], 50)
+    else:
+        bins = np.linspace(0, spike_times[-1], 50)
```
katduecker, line 471: I have explored this solution before, but I found that it reduces temporal resolution if there are events early on in the trial. Current implementation (top figure) shows 3 bins, at times of APs, this implementation shows 1 bin (bottom figure). I agree that this one makes more sense for the histogram and I think it also make sense to show spiking of the pyramidal neurons and basket cells. But we lose resolution when showing our proximal and distal drives (so they might just be lumped together into one bin and you can't infer the standard deviation of the spikes).

current_bins

new_bins

katduecker (commented):
Just had a quick look and added some comments. I think it all works, I'm just curious about the answers!

hnn_core/cell_response.py
ntolley: A good observation! I think my original motivation was to help with reloading back in data in the future, but you're right this should be removed.

hnn_core/viz.py
ntolley: Personally my preference is to have the bin resolution tied to the length of the simulation, Since we have an option of raster plots it's possible to see everything with the "maximal" temporal resolution.

However the point is well taken that there should be a `bins` argument to this!

PR #649, opened 2023-05-22
[MRG] NetworkPlot class to enable interactive visualizations and animations

ntolley:
Here's my first pass at creating a viz class that can be used to quickly render the full 3D network, as well as membrane potentials of individual sections. This is meant to be largely generic and flexible so that it can be co-opted for other purposes. I think the most interesting applications at the moment would be:

- Quickly generating animations of HNN simulations
- Embedding the 3D plot into the current ipywidgets GUI

The current strategy for the class is to do all of the heavy lifting at the initialization (namely creating the 3D object, and mapping the voltage recordings to the appropriate colors). For a quick demo, the code below can be run on my laptop on a single core in approximately 20 sec:
```py
from hnn_core import jones_2009_model, simulate_dipole
from hnn_core.network_models import add_erp_drives_to_jones_model
from hnn_core.viz import NetworkPlot

net = jones_2009_model()
net.set_cell_positions(inplane_distance=300)
add_erp_drives_to_jones_model(net)
dpl = simulate_dipole(net, dt=0.5, tstop=170, record_vsec='all')

net_plot = NetworkPlot(net)
```
image

If you want to play around with visualizing the voltages, you can try the following code block:
```py
from ipywidgets import interact, IntSlider

def update_plot(t_idx, elev, azim):
    net_plot.update_section_voltages(t_idx)
    net_plot.elev = elev
    net_plot.azim = azim
    return net_plot.fig

time_slider = IntSlider(min=0, max=len(net_plot.times), value=1, continuous_update=False)
elev_slider = IntSlider(min=-100,max=100, value=10, continuous_update=False)
azim_slider = IntSlider(min=-100,max=100, value=-100, continuous_update=False)

interact(update_plot, t_idx=time_slider, elev=elev_slider, azim=azim_slider)
```
image

### DIFF
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -59,7 +59,7 @@ Changelog
   json, by `George Dang`_ and `Rajat Partani`_ in :gh:`757`
 
 - Added :class:`~hnn_core/viz/NetworkPlotter` to visualize and animate network simulations,
-  by `Nick Tolley`_ in :gh:`649`
+  by `Nick Tolley`_ in :gh:`649`.
 
 Bug
 ~~~
--- a/doc/whats_new.rst
+++ b/doc/whats_new.rst
@@ -58,6 +58,9 @@ Changelog
 - Added feature to read/write :class:`~hnn_core.Network` configurations to
   json, by `George Dang`_ and `Rajat Partani`_ in :gh:`757`
 
+- Added :class:`~hnn_core/viz/NetworkPlotter` to visualize and animate network simulations,
+  by `Nick Tolley`_ in :gh:`649`
+
 Bug
 ~~~
 - Fix inconsistent connection mapping from drive gids to cell gids, by
--- a/examples/howto/plot_hnn_animation.py
+++ b/examples/howto/plot_hnn_animation.py
@@ -1,6 +1,6 @@
 """
 ================================
-XX. Modifying local connectivity
+06. Animating HNN simulations
 ================================
 
 This example demonstrates how to animate HNN simulations
@@ -10,28 +10,58 @@ This example demonstrates how to animate HNN simulations
 
 
 ###############################################################################
-from hnn_core import jones_2009_model, simulate_dipole
+# First, we'll import the necessary modules for instantiating a network and
+# running a simulation that we would like to animate.
+import os.path as op
+import hnn_core
+from hnn_core import jones_2009_model, simulate_dipole, read_params
 from hnn_core.network_models import add_erp_drives_to_jones_model
-from hnn_core.viz import NetworkPlot
-
-net = jones_2009_model()
-net.set_cell_positions(inplane_distance=300)
-add_erp_drives_to_jones_model(net)
-dpl = simulate_dipole(net, dt=0.5, tstop=170, record_vsec='all')
-
-net_plot = NetworkPlot(net)
 
 ###############################################################################
-from ipywidgets import interact, IntSlider
+# We begin by instantiating the network. For this example, we will reduce the
+# number of cells in the network to speed up the simulations.
+hnn_core_root = op.dirname(hnn_core.__file__)
+params_fname = op.join(hnn_core_root, 'param', 'default.json')
+params = read_params(params_fname)
+params.update({'N_pyr_x': 3, 'N_pyr_y': 3})
+net = jones_2009_model(params)
 
-def update_plot(t_idx, elev, azim):
-    net_plot.update_section_voltages(t_idx)
-    net_plot.elev = elev
-    net_plot.azim = azim
-    return net_plot.fig
+# Note that we move the cells further apart to allow better visualization of
+# the network (default inplane_distance=1.0 µm).
+net.set_cell_positions(inplane_distance=300)
 
-time_slider = IntSlider(min=0, max=len(net_plot.times), value=1, continuous_update=False)
-elev_slider = IntSlider(min=-100, max=100, value=10, continuous_update=False)
-azim_slider = IntSlider(min=-100, max=100, value=-100, continuous_update=False)
+###############################################################################
+# The :class:`hnn_core.viz.NetworkPlotter` class can be used to visualize
+# the 3D structure of the network.
+from hnn_core.viz import NetworkPlotter
 
-interact(update_plot, t_idx=time_slider, elev=elev_slider, azim=azim_slider)
+net_plot = NetworkPlotter(net)
+net_plot.fig
+
+###############################################################################
+# We can also visualize the network from another angle by adjusting the
+# azimuth and elevation parameters.
+net_plot.azim = 45
+net_plot.elev = 40
+net_plot.fig
+
+###############################################################################
+# Next we add event related potential (ERP) producing drives to the network
+# and run the simulation (see
+# :ref:`evoked example <sphx_glr_auto_examples_plot_simulate_evoked.py>`
+# for more details).
+# To visualize the membrane potential of cells in the
+# network, we need use `simulate_dipole(..., record_vsec='all')` which turns
+# on the recording of voltages in all sections of all cells in the network.
+add_erp_drives_to_jones_model(net)
+dpl = simulate_dipole(net, tstop=170, record_vsec='all')
+net_plot = NetworkPlotter(net)  # Reinitialize plotter with simulated network
+
+###############################################################################
+# Finally, we can animate the simulation using the `export_movie()` method. We
+# can adjust the xyz limits of the plot to better visualize the network.
+net_plot.xlim = (400, 1600)
+net_plot.ylim = (400, 1600)
+net_plot.zlim = (-500, 1600)
+net_plot.azim = 225
+net_plot.export_movie('animation_demo.gif', dpi=100, fps=30, interval=100)
--- a/examples/howto/plot_hnn_animation.py
+++ b/examples/howto/plot_hnn_animation.py
@@ -21,7 +21,7 @@ from hnn_core.network_models import add_erp_drives_to_jones_model
 ###############################################################################
 # We begin by instantiating the network. For this example, we will reduce the
 # number of cells in the network to speed up the simulations.
-net = jones_2009_model(add_drives_from_params=True, mesh_shape=(3, 3))
+net = jones_2009_model(mesh_shape=(3, 3))
 
 # Note that we move the cells further apart to allow better visualization of
 # the network (default inplane_distance=1.0 µm).
--- a/examples/howto/plot_hnn_animation.py
+++ b/examples/howto/plot_hnn_animation.py
@@ -13,6 +13,7 @@ This example demonstrates how to animate HNN simulations
 # First, we'll import the necessary modules for instantiating a network and
 # running a simulation that we would like to animate.
 import os.path as op
+
 import hnn_core
 from hnn_core import jones_2009_model, simulate_dipole, read_params
 from hnn_core.network_models import add_erp_drives_to_jones_model
--- a/examples/howto/plot_hnn_animation.py
+++ b/examples/howto/plot_hnn_animation.py
@@ -21,11 +21,7 @@ from hnn_core.network_models import add_erp_drives_to_jones_model
 ###############################################################################
 # We begin by instantiating the network. For this example, we will reduce the
 # number of cells in the network to speed up the simulations.
-hnn_core_root = op.dirname(hnn_core.__file__)
-params_fname = op.join(hnn_core_root, 'param', 'default.json')
-params = read_params(params_fname)
-params.update({'N_pyr_x': 3, 'N_pyr_y': 3})
-net = jones_2009_model(params)
+net = jones_2009_model(add_drives_from_params=True, mesh_shape=(3, 3))
 
 # Note that we move the cells further apart to allow better visualization of
 # the network (default inplane_distance=1.0 µm).
--- a/examples/howto/plot_hnn_animation.py
+++ b/examples/howto/plot_hnn_animation.py
@@ -1,0 +1,99 @@
+"""
+================================
+XX. Modifying local connectivity
+================================
+
+This example demonstrates how to animate HNN simulations
+"""
+
+# Author: Nick Tolley <nicholas_tolley@brown.edu>
+
+# sphinx_gallery_thumbnail_number = 2
+
+
+###############################################################################
+def plot_network(net, ax, t_idx, colormap):
+    """
+    colormap : str
+        The name of a matplotlib colormap. Default: 'viridis'
+    """
+
+    if ax is None:
+        ax = plt.axes(projection='3d')
+
+    xlim = (-200, 3100)
+    ylim = (-200, 3100)
+    # ylim = (-3000, 3100)
+
+    zlim = (-300, 2200)
+    #viridis = cm.get_cmap('viridis', 8)
+
+    for cell_type in net.cell_types:
+        gid_range = net.gid_ranges[cell_type]
+        for gid_idx, gid in enumerate(gid_range):
+            print(gid, end=' ')
+
+            cell = net.cell_types[cell_type]
+            # vsec = {sec_name: ((np.array(net.cell_response.vsec[0][gid][
+            #         sec_name]) - vmin) / (vmax - vmin)) for
+            #         sec_name in cell.sections.keys()}
+            # section_colors = {sec_name: viridis(vsec[sec_name][t_idx]) for
+            #                   sec_name in cell.sections.keys()}
+
+            section_colors = 'C0'
+
+            pos = net.pos_dict[cell_type][gid_idx]
+            pos = (float(pos[0]), float(pos[2]), float(pos[1]))
+            # plot_cell_morphology(
+            #     cell, ax=ax, show=False, pos=pos,
+            #     xlim=xlim, ylim=ylim, zlim=zlim, color=section_colors)
+            cell.plot_morphology(ax=ax, show=False, color=section_colors,
+                                 pos=pos, xlim=xlim, ylim=ylim, zlim=zlim)
+    # ax.view_init(10, -100)
+    ax.view_init(10, -500)
+
+    ax.set_xlim(xlim)
+    ax.set_ylim(ylim)
+    ax.set_zlim(zlim)
+    # ax.axis('on')
+
+    return ax
+
+def get_colors(net, t_idx, colormap):
+    color_list = list()
+    for cell_type in net.cell_types:
+        gid_range = net.gid_ranges[cell_type]
+        for gid_idx, gid in enumerate(gid_range):
+
+            cell = net.cell_types[cell_type]
+            vmin, vmax = -100, 50
+
+            for sec_name in cell.sections.keys():
+                vsec = (np.array(net.cell_response.vsec[0][gid][sec_name]) - vmin) / (vmax - vmin)
+                color_list.append(colormap(vsec[t_idx]))
+    return color_list
+
+                                
+
+def update_colors(ax, net, t_idx, colormap):
+    color_list = get_colors(net, t_idx, colormap)
+    lines = ax.get_lines()
+    for line, color in zip(lines, color_list):
+        line.set_color(color)
+    ax.view_init(10, -500)
+    
+
+net = jones_2009_model()
+net.set_cell_positions(inplane_distance=300)
+add_erp_drives_to_jones_model(net)
+dpl = simulate_dipole(net, dt=0.5, tstop=100, record_vsec='all')
+
+
+fig = plt.figure()
+ax = fig.add_subplot(projection='3d')
+plot_network(net, ax=ax, t_idx=None, colormap=None)
+
+colormap = colormaps['viridis']
+update_colors(ax, net, t_idx=100, colormap=colormap)
+ax.view_init(20, 100)
+fig
--- a/examples/howto/plot_hnn_animation.py
+++ b/examples/howto/plot_hnn_animation.py
@@ -8,92 +8,30 @@ This example demonstrates how to animate HNN simulations
 
 # Author: Nick Tolley <nicholas_tolley@brown.edu>
 
-# sphinx_gallery_thumbnail_number = 2
-
 
 ###############################################################################
-def plot_network(net, ax, t_idx, colormap):
-    """
-    colormap : str
-        The name of a matplotlib colormap. Default: 'viridis'
-    """
-
-    if ax is None:
-        ax = plt.axes(projection='3d')
-
-    xlim = (-200, 3100)
-    ylim = (-200, 3100)
-    # ylim = (-3000, 3100)
-
-    zlim = (-300, 2200)
-    #viridis = cm.get_cmap('viridis', 8)
-
-    for cell_type in net.cell_types:
-        gid_range = net.gid_ranges[cell_type]
-        for gid_idx, gid in enumerate(gid_range):
-            print(gid, end=' ')
-
-            cell = net.cell_types[cell_type]
-            # vsec = {sec_name: ((np.array(net.cell_response.vsec[0][gid][
-            #         sec_name]) - vmin) / (vmax - vmin)) for
-            #         sec_name in cell.sections.keys()}
-            # section_colors = {sec_name: viridis(vsec[sec_name][t_idx]) for
-            #                   sec_name in cell.sections.keys()}
-
-            section_colors = 'C0'
-
-            pos = net.pos_dict[cell_type][gid_idx]
-            pos = (float(pos[0]), float(pos[2]), float(pos[1]))
-            # plot_cell_morphology(
-            #     cell, ax=ax, show=False, pos=pos,
-            #     xlim=xlim, ylim=ylim, zlim=zlim, color=section_colors)
-            cell.plot_morphology(ax=ax, show=False, color=section_colors,
-                                 pos=pos, xlim=xlim, ylim=ylim, zlim=zlim)
-    # ax.view_init(10, -100)
-    ax.view_init(10, -500)
-
-    ax.set_xlim(xlim)
-    ax.set_ylim(ylim)
-    ax.set_zlim(zlim)
-    # ax.axis('on')
-
-    return ax
-
-def get_colors(net, t_idx, colormap):
-    color_list = list()
-    for cell_type in net.cell_types:
-        gid_range = net.gid_ranges[cell_type]
-        for gid_idx, gid in enumerate(gid_range):
-
-            cell = net.cell_types[cell_type]
-            vmin, vmax = -100, 50
-
-            for sec_name in cell.sections.keys():
-                vsec = (np.array(net.cell_response.vsec[0][gid][sec_name]) - vmin) / (vmax - vmin)
-                color_list.append(colormap(vsec[t_idx]))
-    return color_list
-
-                                
-
-def update_colors(ax, net, t_idx, colormap):
-    color_list = get_colors(net, t_idx, colormap)
-    lines = ax.get_lines()
-    for line, color in zip(lines, color_list):
-        line.set_color(color)
-    ax.view_init(10, -500)
-    
+from hnn_core import jones_2009_model, simulate_dipole
+from hnn_core.network_models import add_erp_drives_to_jones_model
+from hnn_core.viz import NetworkPlot
 
 net = jones_2009_model()
 net.set_cell_positions(inplane_distance=300)
 add_erp_drives_to_jones_model(net)
-dpl = simulate_dipole(net, dt=0.5, tstop=100, record_vsec='all')
+dpl = simulate_dipole(net, dt=0.5, tstop=170, record_vsec='all')
 
+net_plot = NetworkPlot(net)
 
-fig = plt.figure()
-ax = fig.add_subplot(projection='3d')
-plot_network(net, ax=ax, t_idx=None, colormap=None)
+###############################################################################
+from ipywidgets import interact, IntSlider
 
-colormap = colormaps['viridis']
-update_colors(ax, net, t_idx=100, colormap=colormap)
-ax.view_init(20, 100)
-fig
+def update_plot(t_idx, elev, azim):
+    net_plot.update_section_voltages(t_idx)
+    net_plot.elev = elev
+    net_plot.azim = azim
+    return net_plot.fig
+
+time_slider = IntSlider(min=0, max=len(net_plot.times), value=1, continuous_update=False)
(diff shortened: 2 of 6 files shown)


ntolley:
@jasmainak @rythorpe perhaps it'd be a good idea to make this a separate file? If we can find a solution to updating the section colors very rapidly, we can even consider adding a "play" button inside the GUI (or on the HNN website!) like the example here: https://matplotlib.org/stable/gallery/animation/unchained.html

@chenghuzi you might have some good insight on how to make the current code more "reactive"

examples/howto/plot_hnn_animation.py
```diff
@@ -0,0 +1,37 @@
+"""
+================================
+XX. Modifying local connectivity
```
jasmainak, line 3: to be fixed :)

hnn_core/viz.py
```diff
@@ -1185,3 +1198,234 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
...
+    def bg_color(self, bg_color):
+        self._bg_color = bg_color
+        self.ax.set_facecolor(self._bg_color)
+
+    @property
+    def voltage_colormap(self):
+        return self._voltage_colormap
+
+    @voltage_colormap.setter
+    def voltage_colormap(self, voltage_colormap):
```
jasmainak, line 1428: an `export` method? I think matplotlib allows exporting animations ... I want to see HNN animations in presentations of people I don't know ;-)

jasmainak:
Agnostic about saving to a different file ... see my comment about the azimuth/elevation control. Maybe it makes the class thinner?

hnn_core/viz.py
```diff
@@ -1185,3 +1198,234 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
     plt_show(show)
 
     return ax.get_figure()
+
+
+class NetworkPlot:
```
jasmainak, line 1203:
```suggestion
class NetworkDataPlotter:
```

?

ntolley: `NetworkPlotter`? It might useful too for just visualizing the 3D morphology without any data too

ntolley:
Dropping this here for easy testing once we get closer to merging. I think rendering movies is going to be a longish wait for larger networks, but not absurdly so (maybe like 5-10 minutes? if you downsample?)
```py
import hnn_core
import os.path as op
from hnn_core import jones_2009_model, simulate_dipole, read_params
from hnn_core.network_models import add_erp_drives_to_jones_model
from hnn_core.viz import NetworkPlot

params.update({'N_pyr_x': 3, 'N_pyr_y': 3})
net = jones_2009_model(params)
net.set_cell_positions(inplane_distance=300)
add_erp_drives_to_jones_model(net)
dpl = simulate_dipole(net, dt=0.5, tstop=170, record_vsec='all')

net_plot = NetworkPlot(net)
net_plot.export_movie('demo.gif', dpi=200)
```
demo

hnn_core/viz.py
```diff
@@ -1185,3 +1198,247 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
...
+                                     ylim=self.ylim, zlim=self.zlim)
+
+    def _update_axes(self):
+        self.ax.set_xlim(self._xlim)
+        self.ax.set_ylim(self._ylim)
+        self.ax.set_zlim(self._zlim)
+
+        self.ax.view_init(self._elev, self._azim)
+
+    def export_movie(self, fname, dpi=300):
```
ntolley, line 1325: @jasmainak here's the export function!

ntolley: There are some solutions here that might speed up the exporting by circumventing the matplotlib animation module:
https://stackoverflow.com/questions/4092927/generating-movie-from-python-without-saving-individual-frames-to-files

hnn_core/viz.py
jasmainak: Yeah if data is None

hnn_core/viz.py
jasmainak: Is it possible to add a progressbar using a callback? I don't think the speed is a showstopper here. Maybe a param called "decim" could be added to take every nth frame. Make it work, make it nice, make it fast ...

chenghuzi:
> @jasmainak @rythorpe perhaps it'd be a good idea to make this a separate file? If we can find a solution to updating the section colors very rapidly, we can even consider adding a "play" button inside the GUI (or on the HNN website!) like the example here: https://matplotlib.org/stable/gallery/animation/unchained.html
> 
> @chenghuzi you might have some good insight on how to make the current code more "reactive"

Great work! Although, I'm concerned about the reactivity of our GUI. We are taking snapshots from the buffer and using ﻿`imshow` to render network images for now. So, it might be difficult to have real-time interaction in the new GUI with these new settings. But this will not affect the current function so there is nothing else to worry about.

hnn_core/viz.py
ntolley: Definitely would be a nice addition but I think I'd need to be much more matplotlib savvy to know how to implement that :smile:

hnn_core/viz.py
```diff
@@ -1222,3 +1235,294 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
...
+        # Trial and time indices
+        _validate_type(trial_idx, int, 'trial_idx')
+        _validate_type(time_idx, int, 'time_idx')
+
+        self._trial_idx = trial_idx
+        self._time_idx = time_idx
+
+        # Get voltage data and corresponding colors
+        self.vsec_array = self._get_voltages()
+        self.color_array = self.colormap(self.vsec_array)
```
rythorpe, line 1319: It'd be nice if we could populate `self.color_array` with zeros or ones at the onset so that the network can be plotted even if the simulation is run with `record_vsec=False`.

ntolley: Super good suggestion! Also makes for a good smoke test to plot a simulated/unsimulated network

rythorpe: Overall, this branch is such a cool feature! Super excited to get it merged.

examples/howto/plot_hnn_animation.py
```diff
@@ -0,0 +1,37 @@
...
+This example demonstrates how to animate HNN simulations
+"""
+
+# Author: Nick Tolley <nicholas_tolley@brown.edu>
+
+
+###############################################################################
+from hnn_core import jones_2009_model, simulate_dipole
+from hnn_core.network_models import add_erp_drives_to_jones_model
+from hnn_core.viz import NetworkPlot
```
rythorpe, line 15:
```suggestion
from hnn_core.viz import NetworkPlotter
```

rythorpe: I'm guessing the name of this class got changed somewhere along the way and you forgot to update it here?

examples/howto/plot_hnn_animation.py
```diff
@@ -0,0 +1,37 @@
...
+###############################################################################
+from ipywidgets import interact, IntSlider
+
+def update_plot(t_idx, elev, azim):
+    net_plot.update_section_voltages(t_idx)
+    net_plot.elev = elev
+    net_plot.azim = azim
+    return net_plot.fig
+
+time_slider = IntSlider(min=0, max=len(net_plot.times), value=1, continuous_update=False)
```
rythorpe, line 33: For some reason the interactive slider didn't work for me. It might be a version issue and totally unrelated to your code here. Have you encountered a similar issue at any point @ntolley?

ntolley:
@rythorpe @jasmainak this PR is ready for review! Two weird errors have come up however. First it seems like sphinx-gallery doesn't support python versions older than 3.8
image

Does this mean our minimum python version needs to be updated?

Also the windows unit tests are failing with an error in the GUI code...
image

This is the offending line in case you guys have any ideas on the possible cause:
https://github.com/ntolley/hnn-core/blob/e3aa27c91ba4ed7478d2214e53fea618b793f45b/hnn_core/tests/test_viz.py#L23C5-L23C5

ntolley:
Yup just confirmed, sphinx now requires python>=3.8

https://github.com/sphinx-gallery/sphinx-gallery/blob/e2903c9e43572a0a29850f7ab153ff69098f5c81/setup.py#L58C5-L58C5

.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ jobs:
               command: |
                   export PATH=~/miniconda/bin:$PATH
                   conda update --yes --quiet conda
-                  conda create -n testenv --yes pip python=3.7
+                  conda create -n testenv --yes pip python=3.8
```
rythorpe, line 30: Do you think we need to update `setup.py` and our unit tests with `python>=3.8` as well to ensure all builds are getting tested properly?

rythorpe:
> @rythorpe @jasmainak this PR is ready for review! Two weird errors have come up however. First it seems like sphinx-gallery doesn't support python versions older than 3.8 image
> 
> Does this mean our minimum python version needs to be updated?
> 
> Also the windows unit tests are failing with an error in the GUI code... image
> 
> This is the offending line in case you guys have any ideas on the possible cause: https://github.com/ntolley/hnn-core/blob/e3aa27c91ba4ed7478d2214e53fea618b793f45b/hnn_core/tests/test_viz.py#L23C5-L23C5

Ugh, I'm guessing the most recent version of matplotlib has diverged for windows vs. linux. Something to do with the `agg` backend....

ntolley:
Ok updating the python version used for CircleCI worked, however it seems something in the MNE code is breaking the somato example now:
image

.circleci/config.yml
jasmainak: that's true but you won't realize if other dependencies start requiring 3.8 now that you have updated it here. I don't think we have a CI that tests for python < 3.8? Better to be consistent and update readme/setup etc.

examples/howto/plot_hnn_animation.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+
+# Author: Nick Tolley <nicholas_tolley@brown.edu>
+
+
+###############################################################################
+# First, we'll import the necessary modules for instantiating a network and
+# running a simulation that we would like to animate.
+import os.path as op
+import hnn_core
```
jasmainak, line 16:
```suggestion
import os.path as op

import hnn_core
```

examples/howto/plot_hnn_animation.py
```diff
@@ -0,0 +1,67 @@
...
+from hnn_core import jones_2009_model, simulate_dipole, read_params
+from hnn_core.network_models import add_erp_drives_to_jones_model
+
+###############################################################################
+# We begin by instantiating the network. For this example, we will reduce the
+# number of cells in the network to speed up the simulations.
+hnn_core_root = op.dirname(hnn_core.__file__)
+params_fname = op.join(hnn_core_root, 'param', 'default.json')
+params = read_params(params_fname)
+params.update({'N_pyr_x': 3, 'N_pyr_y': 3})
```
jasmainak, line 26: this won't work with the `hdf5` format. We'll need an API to do this.

examples/howto/plot_hnn_animation.py
```diff
@@ -0,0 +1,67 @@
...
+# the network (default inplane_distance=1.0 µm).
+net.set_cell_positions(inplane_distance=300)
+
+###############################################################################
+# The :class:`hnn_core.viz.NetworkPlotter` class can be used to visualize
+# the 3D structure of the network.
+from hnn_core.viz import NetworkPlotter
+
+net_plot = NetworkPlotter(net)
+net_plot.fig
```
jasmainak, line 39: I don't see a figure in `ipython`. Maybe you need a `plt.show` or `plt.draw` ? Maybe `net_plot.show()` ?

hnn_core/viz.py
```diff
@@ -1222,3 +1235,318 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
     plt_show(show)
 
     return ax.get_figure()
+
+
+class NetworkPlotter:
+    """Helper class to visualize full
+       morphology of HNN model.
```
jasmainak, line 1242: needs to be single line according to pep 257

hnn_core/viz.py
```diff
@@ -1222,3 +1235,318 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
...
+        a new figure is created.
+    vmin : int | float
+        Lower limit of colormap for plotting voltage
+        Default: -100 mV
+    vmax : int | float
+        Upper limit of colormap for plotting voltage
+        Default: 50 mV
+    bg_color : str
+        Background color of ax. Default: 'black'
+    voltage_colormap : str
```
jasmainak, line 1259: we're missing a `colorbar` parameter that would default to `True`

hnn_core/viz.py
```diff
@@ -1222,3 +1235,318 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
...
+        self.ax.set_xlim(self._xlim)
+        self.ax.set_ylim(self._ylim)
+        self.ax.set_zlim(self._zlim)
+
+        self.ax.view_init(self._elev, self._azim)
+
+    def export_movie(self, fname, fps=30, dpi=300, decim=10,
+                     interval=30, frame_start=0, frame_stop=None,
+                     writer='pillow'):
+        """Export movie of network activity
```
jasmainak, line 1395: needs a `Parameters` heading and an empty line

examples/howto/plot_hnn_animation.py
```diff
@@ -0,0 +1,67 @@
...
+
+###############################################################################
+# Next we add event related potential (ERP) producing drives to the network
+# and run the simulation (see
+# :ref:`evoked example <sphx_glr_auto_examples_plot_simulate_evoked.py>`
+# for more details).
+# To visualize the membrane potential of cells in the
+# network, we need use `simulate_dipole(..., record_vsec='all')` which turns
+# on the recording of voltages in all sections of all cells in the network.
+add_erp_drives_to_jones_model(net)
```
jasmainak, line 56: we can leave this for now, but in the future I'd be in favor of deprecating this function for a boilerplate file that contains the drives in an `hdf5` file and we read it directly from that ... functions should not be black boxes ...

hnn_core/tests/test_viz.py
```diff
@@ -208,4 +217,108 @@ def test_dipole_visualization():
...
+    with pytest.raises(TypeError, match='azim must be'):
+        _ = NetworkPlotter(net, azim='blah')
+    with pytest.raises(TypeError, match='vmin must be'):
+        _ = NetworkPlotter(net, vmin='blah')
+    with pytest.raises(TypeError, match='vmax must be'):
+        _ = NetworkPlotter(net, vmax='blah')
+    with pytest.raises(TypeError, match='trial_idx must be'):
+        _ = NetworkPlotter(net, trial_idx=1.0)
+    with pytest.raises(TypeError, match='time_idx must be'):
+        _ = NetworkPlotter(net, time_idx=1.0)
```
jasmainak, line 238: I guess github co-pilot wasn't clever enough to compress this in a loop? ;-)

hnn_core/tests/test_viz.py
```diff
@@ -208,4 +217,108 @@ def test_dipole_visualization():
...
+    with pytest.raises(TypeError, match='azim must be'):
+        net_plot.azim = 'blah'
+    with pytest.raises(TypeError, match='vmin must be'):
+        net_plot.vmin = 'blah'
+    with pytest.raises(TypeError, match='vmax must be'):
+        net_plot.vmax = 'blah'
+    with pytest.raises(TypeError, match='trial_idx must be'):
+        net_plot.trial_idx = 1.0
+    with pytest.raises(TypeError, match='time_idx must be'):
+        net_plot.time_idx = 1.0
```
jasmainak, line 290: why is this repeated?

examples/howto/plot_hnn_animation.py
```diff
@@ -0,0 +1,67 @@
...
+net_plot = NetworkPlotter(net)  # Reinitialize plotter with simulated network
+
+###############################################################################
+# Finally, we can animate the simulation using the `export_movie()` method. We
+# can adjust the xyz limits of the plot to better visualize the network.
+net_plot.xlim = (400, 1600)
+net_plot.ylim = (400, 1600)
+net_plot.zlim = (-500, 1600)
+net_plot.azim = 225
+net_plot.export_movie('animation_demo.gif', dpi=100, fps=30, interval=100)
```
jasmainak, line 67: something is wrong with the default parameters ... it doesn't play well with our default simulation parameters. I tried:

```py
net_plot.export_movie('test.mov')
```

and it was stuck forever.

hnn_core/viz.py
```diff
@@ -1222,3 +1235,318 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
     plt_show(show)
 
     return ax.get_figure()
+
+
+class NetworkPlotter:
```
jasmainak, line 1240: needs to be added to `api.rst`

.circleci/config.yml
rythorpe: I went ahead and updated this in #677. We'll see if it fixes the Windows `test_viz.py` error....

jasmainak: excellent!!

jasmainak:
> it seems something in the MNE code is breaking the somato example now:

looks like `nibabel` is now being used in MNE for reading the surface files ... I reported this upstream:

https://github.com/mne-tools/mne-python/pull/11565/files#r1337642633

hopefully it'll get fixed soon

hnn_core/tests/test_viz.py
ntolley: It is redundant, but I figure there should be separate tests for setting parameters before or after initialization

examples/howto/plot_hnn_animation.py
ntolley: are you ok to leave this for now and replace when we push #651 over the finish line?

examples/howto/plot_hnn_animation.py
ntolley: This shows up in circleCI docs: 
image

honestly this issue of when we do and don't need to use `plt.show()` is really opaque to me, it is only a problem sometimes on some certain systems...

examples/howto/plot_hnn_animation.py
ntolley: how long did you wait? on my laptop it takes 1min 30sec to generate the gif

The default `writer` is `pillow` which doesn't support the `.mov` extension. I was able to export using the `ffmpeg` writer.

Having a programmatic check on the extensions would be nice to have, but it seems like it would be a bit annoying to query this for every valid writer: https://stackoverflow.com/questions/22571044/what-formats-can-matplotlib-animations-be-saved-as 

(I've dug into the attributes and unfortunately there isn't a clear list I can find)

jasmainak:
@rythorpe @gtdang could you guys review and merge this PR if it's good by you? I am finding it difficult to find time and don't want to block progress ... this could be a good excuse for a release.

examples/howto/plot_hnn_animation.py
rythorpe: I think we can now remove the reference to `default.json` and avoiding using/modifying the params dict here.

hnn_core/cell.py
```diff
@@ -749,7 +749,9 @@ def parconnect_from_src(self, gid_presyn, nc_dict, postsyn,
 
         return nc
 
-    def plot_morphology(self, ax=None, color=None, show=True):
+    def plot_morphology(self, ax=None, color=None, pos=(0, 0, 0),
+                        xlim=(-250, 150), ylim=(-100, 100), zlim=(-100, 1200),
+                        show=True):
```
rythorpe, line 754: These new args should be added to the docstring.

hnn_core/viz.py
```diff
@@ -847,7 +847,9 @@ def _linewidth_from_data_units(ax, linewidth):
     return linewidth * (length / value_range)
 
 
-def plot_cell_morphology(cell, ax, color=None, show=True):
+def plot_cell_morphology(
+        cell, ax, color=None, pos=(0, 0, 0), xlim=(-250, 150),
+        ylim=(-100, 100), zlim=(-100, 1200), show=True):
```
rythorpe, line 852: Same here: add to docstring.

rythorpe (commented):
This is almost ready @ntolley! I just tested it out and made a few minor comments.

hnn_core/tests/test_viz.py
rythorpe: This is a nitpick, but could we put these tests in `for` loop to conserve space + enhance readability?

rythorpe:
Just in case this slipped your attention @ntolley, this PR is sooooo close ;-)

hnn_core/viz.py
```diff
@@ -887,18 +892,26 @@ def plot_cell_morphology(cell, ax, color=None, show=True):
...
         linewidth = _linewidth_from_data_units(ax, section.diam)
         end_pts = section.end_pts
         xs, ys, zs = list(), list(), list()
         for pt in end_pts:
-            dx = cell.pos[0] - cell.sections['soma'].end_pts[0][0]
-            dy = cell.pos[1] - cell.sections['soma'].end_pts[0][1]
-            dz = cell.pos[2] - cell.sections['soma'].end_pts[0][2]
+            dx = pos[0] - cell.sections['soma'].end_pts[0][0]
+            dy = pos[1] - cell.sections['soma'].end_pts[0][1]
+            dz = pos[2] - cell.sections['soma'].end_pts[0][2]
```
gtdang, line 914: It doesn't seem like these lines need to be nested in this for loop? Seems like all of those values are static. Should they be using pt instead of pos? Or the iterated section?

hnn_core/viz.py
```diff
@@ -1222,3 +1235,365 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
...
+            # Check if voltage recorded
+            if net._params['record_vsec'] == 'all':
+                self._vsec_recorded = True
+            else:
+                self._vsec_recorded = False
+        else:
+            self._vsec_recorded = False
+            self.times = None
+
+        _validate_type(vmin, (int, float), 'vmin')
```
gtdang, line 1299: I would suggest moving all the type validation to the top of the init for both logic and readability. I'm assuming that if any of the validations fail an error is being raised and you won't continue with the init?

gtdang (changes requested):

hnn_core/tests/test_viz.py
```diff
@@ -211,4 +222,117 @@ def test_dipole_visualization():
     with pytest.raises(ValueError, match="'beta_dist' must be"):
         net.cell_response.plot_spikes_hist(color={'beta_prox': 'r'})
 
+    # test NetworkPlotter class
```
gtdang, line 225: The tests in this module are too long. Generally, each unit test should focus on one specific aspect of the behavior of a function to better isolate location of bugs. That refactor should probably come in a separate PR, but for this one I think the NetworkPlotter should be spun out into its own test at the very least. Ideally it would be broken even further into the main operations you are testing. Here are the separate operations I see from a quick glance: 

1. init  keyword validation 
2. setter validation 
3. simulation validation 
4. movie export 

Creating a pytest fixture for the network will help with redundancies. Also 2nd creating a loop for the pytest raises. You can use the f-strings and dictionary unpacking. Something like:
```python
args = ['xlim', 'ylim', 'zlim']
    for arg in args:
        d_kw = {arg: 'blah'}
        with pytest.raises(TypeError, match=f'{arg} must be'):
            _ = NetworkPlotter(net, **d_kw)
```

hnn_core/tests/test_viz.py
```diff
@@ -211,4 +222,117 @@ def test_dipole_visualization():
...
+    assert net_plot.trial_idx == 1
+    assert net_plot.time_idx == 5
+
+    assert net_plot.bgcolor == 'white'
+    assert net_plot.fig.get_facecolor() == (1.0, 1.0, 1.0, 1.0)
+
+    assert net_plot.voltage_colormap == 'jet'
+
+    # Test animation export and voltage plotting
+    net_plot.export_movie('demo.gif', dpi=200, decim=100, writer='pillow')
```
gtdang, line 336: I guess this will test if this operation crashes. A more comprehensive test would be to make sure this file actually exists. By following up with an os.path.exists() or os.path.isfile(). Bonus points for writing to temp space.

gtdang (changes requested):

hnn_core/tests/test_viz.py
ntolley: Thanks for the suggestions @gtdang! Just updated the tests accordingly, let me know what you think

rythorpe:
@ntolley do you know if `plot_cell_morphology` sets the length of each cell section before plotting it? Or are the section endpoints already updated with the appropriate length? Apologies if we already hashed this out, I just can't remember and was wondering about this the other day....

ntolley:
@rythorpe they are updated automatically! #433 fixed this early on, but then Rajat's project made a new and improved version where he implemented the Neuron functions for calculating end points directly in our python code (necessary for saving this infromation in a JSON/dict format)

hnn_core/tests/test_viz.py
ntolley: Borrowed your logic from the network config PR FOR THIS ONE

hnn_core/tests/test_viz.py
```diff
@@ -212,5 +226,123 @@ def test_dipole_visualization():
...
+    with pytest.raises(TypeError, match='vmin must be'):
+        _ = NetworkPlotter(net, vmin='blah')
+    with pytest.raises(TypeError, match='vmax must be'):
+        _ = NetworkPlotter(net, vmax='blah')
+    with pytest.raises(TypeError, match='trial_idx must be'):
+        _ = NetworkPlotter(net, trial_idx=1.0)
+    with pytest.raises(TypeError, match='time_idx must be'):
+        _ = NetworkPlotter(net, time_idx=1.0)
+    with pytest.raises(TypeError, match='colorbar must be'):
+        _ = NetworkPlotter(net, colorbar='blah')
```
gtdang, line 255: If you wanted a loop for these you could do it like this:

```python
not_num = 'string'
not_int = 1.0
test_kws = {'xlim': not_num, 'ylim': not_num,
            'trial_idx': not_int, 'time_idx': not_int,
            }
for keyword, value in test_kws.items():
    kwargs = {keyword: value}
    with pytest.raises(TypeError, match=f'{keyword} must be'):
        _ = NetworkPlotter(net, **kwargs)
```

But if up to you if you find it less interpretable!

gtdang (commented):
Looking good! Like the test update and the init. Left one comment on the tests.

hnn_core/tests/test_viz.py
```diff
@@ -212,5 +226,108 @@ def test_dipole_visualization():
...
+
+def test_network_plotter_init(setup_net):
+    """Test init keywords of NetworkPlotter class."""
+    net = setup_net
+    # test NetworkPlotter class
+    args = ['xlim', 'ylim', 'zlim', 'elev', 'azim', 'vmin', 'vmax',
+            'trial_idx', 'time_idx', 'colorbar']
+    for arg in args:
+        with pytest.raises(TypeError, match=f'{arg} must be'):
+            net_plot = NetworkPlotter(net, **{arg: 'blah'})
```
gtdang, line 240: Ah that's simpler! Nice!

ntolley:
@jasmainak this is ready for a final round of reviews when you have the chance!

I think there are some tasks that might be delegated to followup PR's (e.g. a better jupyter notebook tutorial?), but in terms of basic functionality and readability I think this is in a good place

hnn_core/viz.py
```diff
@@ -1228,3 +1249,381 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
...
+
+        frames = np.arange(frame_start, frame_stop, decim)
+        ani = animation.FuncAnimation(
+            self.fig, self._set_time_idx, frames, interval=interval)
+
+        writer = animation.writers[writer](fps=fps)
+        ani.save(fname, writer=writer, dpi=dpi)
+        return ani
+
+    # Axis limits
```
jasmainak, line 1475: use the docstring for comments :)

hnn_core/viz.py
```diff
@@ -1228,3 +1249,381 @@ def plot_laminar_csd(times, data, contact_labels, ax=None, colorbar=True,
...
+    def zlim(self):
+        return self._zlim
+
+    @zlim.setter
+    def zlim(self, zlim):
+        _validate_type(zlim, tuple, 'zlim')
+        self._zlim = zlim
+        self.ax.set_zlim(self._zlim)
+
+    # Eleevation and azimuth of 3D viewpoint
```
jasmainak, line 1506: our codespell checker isn't catching this?

jasmainak:
@ntolley this is tremendous!!! 🥳 when can I have the privilege of merging it? I tried it and it works really nice. @rythorpe did you want to look? It's good to go on my end.

One feature request: perhaps we can have a small text in the corner showing the timestamp ... I was a bit confused whether the plot was still updating (or is it instantaneous?) when I moved it around.

jasmainak:
@gtdang do you still have remaining objections? Merging seems blocked because of your comments

jasmainak (approved):

jasmainak:
Don't forget to update `whats_new.rst` !

gtdang (approved):
Looks good!

doc/whats_new.rst
```diff
@@ -58,6 +58,9 @@ Changelog
 - Added feature to read/write :class:`~hnn_core.Network` configurations to
   json, by `George Dang`_ and `Rajat Partani`_ in :gh:`757`
 
+- Added :class:`~hnn_core/viz/NetworkPlotter` to visualize and animate network simulations,
+  by `Nick Tolley`_ in :gh:`649`
```
rythorpe, line 62:
```suggestion
  by `Nick Tolley`_ in :gh:`649`.
```
I know we haven't been good about this so far, but we should try to be more consisent with punctuation.

rythorpe:
Aside from @jasmainak's spelling error and my incredibly nitpicky punctuation suggestion above, this looks great @ntolley !

hnn_core/viz.py
ntolley, line 1506:
```suggestion
    # Elevation and azimuth of 3D viewpoint
```

rythorpe:
I've enabled auto-merge! Thanks @ntolley for this incredible feature - I'm very excited to put it to good use :smile:

PR #651, opened 2023-05-24
[ENH] Added read/write method for network object

raj1701:
GSoC 2023 Project - Develop IO routines for HNN-core outputs

The idea of the project is to add read/write capabilities for all HNN objects such as Network, Cell, Section, Dipole etc. in the HDF5 format. During the summer, I wrote IO functions for all HNN objects, the associated tests as well as refactored some existing code along the journey.

The following PR contains read/write functions and the associated tests for the Network object. The network class encapsulates objects such as Cell, Section, Cell response and Extracellular arrays. These objects are stored along the network itself. Each and every aspect of the functions is tested and documented. The IO functions are capable of handling all network models in HNN core as well any modified network instances. Capability to saving and loading up simulated networks is also present. 

Some other PRs related to the project are -
- #661 - This PR contains python version of NEURON functions such as `define_shape()` and `distance()`. This is done to reduce the dependency on NEURON for saving a Network. NEURON objects are only needed while simulating a Network now.
- #648 - This PR contains read/write functions and associated tests for the Dipole object.
- #644 - This PR contains read/write functions and associated tests for the Cell response object.
- #654  - This PR reverted the IO functionalities for the Cell response object as they are included in the IO functions for the Network object in #651.

Plans for the project after GSoC - Add load/save buttons in HNN GUI for the Network and Dipole object.


hnn_core/tests/test_network.py
```diff
@@ -17,6 +17,19 @@
...
+def test_network_io(tmpdir):
+    net_jones = jones_2009_model()
+    # Writing network
+    net_jones.write(tmpdir.join('net_jones.hdf5'))
+    # Reading network
+    net_jones_read = read_network(tmpdir.join('net_jones.hdf5'))
+    # Check cell_types
+    # assert net_jones.cell_types == net_jones_read.cell_types
+    print(net_jones.cell_types)
+    print(net_jones_read.cell_types)
```
jasmainak, line 29: For test-driven development, you can use `pytest` with `--pdb` to directly drop into the interpreter when a test fails. Then you can use `pdb` commands like `p net_jones.cell_types` etc. Just letting you know the different options on debugging efficiently

hnn_core/network.py
```diff
@@ -412,6 +414,11 @@ def __repr__(self):
                  len(self.pos_dict['L5_basket'])))
         return '<%s | %s>' % (class_name, s)
 
+    def __eq__(self, other):
+        if not isinstance(other, Network):
+            return NotImplemented
+        # Check for all attributes (Discuss)
```
jasmainak, line 420: I think the right way to implement this is to first implement `__eq__` for `Section` and `Cell` object and call them from here. Python builtins already have it implemented so presumably `net.external_drives` etc should be straightforward to check for equality

jasmainak:
For partial function, I think you should expand it out. Iterate over the segments and get the `gbar` values. Like here:

https://github.com/jonescompneurolab/hnn-core/blob/07374a358b3f93abc4093e4d4b5c45bf217a5059/hnn_core/tests/test_network.py#L57

and store them. It's less compact but I think the only way to make it work

raj1701:
I tried rebasing but it didn't work. I followed the following steps
- git fetch origin master
- git checkout <feature_branch>
- git push origin <feature_branch>

I encountered no errors but I couldn't see any changes. Where did I go wrong?

jasmainak:
@raj1701 I think you just need to do:

```sh
$ git rebase master
```

after checking out the `feature_branch` if you already fetched master

raj1701:
Hey @jasmainak @ntolley, I found a bug while saving `cell_response` after simulation

Bug Description
The bug occurs because of the structure of `vsec` and `isec`. `vsec` contains a `list of dict`. The keys of each dict are of type `int`. For hdf5 keys must be of type `str`

Solution
Convert the dict keys to the required types before writing and after reading.

- I have implemented this solution in network object where I have used `_vsec` and `_isec` in `cell_response`. A similar solution will be required for `read/write` in `cell_response.py` (A separate PR mostly).
-  Since `isec` and `vsec` always remained empty when I was working explicitly on the `cell_response` object, the bug never occured. After running a simulation, I got vsec values and thus the bug.
- LMK your thoughts on the solution and whether I should open a separate PR for `cell_response` object

jasmainak:
Yes, you should open a new PR and rebase this branch when that PR is merged

hnn_core/network.py
```diff
@@ -1335,6 +1390,235 @@ def plot_cells(self, ax=None, show=True):
...
+                             conn_seed=drive_data['conn_seed'])
+
+    net.external_drives[drive_data['name']]['events'] = drive_data['events']
+
+
+def read_network(fname):
+    net_data = read_hdf5(fname)
+    params = dict()
+    params['N_pyr_x'] = 10
+    params['N_pyr_y'] = 10
```
jasmainak, line 1600: could this be inferred from `net.gid_ranges` ?

hnn_core/network.py
```diff
@@ -1335,6 +1390,235 @@ def plot_cells(self, ax=None, show=True):
...
+        # Write connectivity
+        # Write rec arrays
+        # Write threshold
+        # Write delay
+
+        # Saving file
+        write_hdf5(fname, net_data, overwrite=True)
+
+
+def _get_section_as_dict(section):
```
jasmainak, line 1430: I would make this a method of `Cell` called `as_dict` ... similar to pandas API

hnn_core/network.py
```diff
@@ -1335,6 +1390,235 @@ def plot_cells(self, ax=None, show=True):
...
+    section_data['end_pts'] = section.end_pts
+    # Need to solve the partial function problem
+    # in mechs
+    # section_data['mechs'] = section.mechs
+    # print(section.mechs)
+    section_data['syns'] = section.syns
+    return section_data
+
+
+def _get_cell_as_dict(cell):
```
jasmainak, line 1445: same, make it a method

hnn_core/cell.py
```diff
@@ -328,6 +339,29 @@ def __repr__(self):
...
+            return False
+
+        if not (self.sections.keys() == other.sections.keys()):
+            return False
+
+        for key in self.sections.keys():
+            if not (self.sections[key] == other.sections[key]):
+                return False
+
+        return True
```
jasmainak, line 363: you should add a test in `test_cell.py` for this ... simple way to test is:

```py
cell2 = cell.copy()
assert cell2 == cell
```

hnn_core/network.py
```diff
@@ -412,6 +415,43 @@ def __repr__(self):
                  len(self.pos_dict['L5_basket'])))
         return '<%s | %s>' % (class_name, s)
 
+    def __eq__(self, other):
```
jasmainak, line 418: add a test for this method

raj1701: In `net.copy()` method, the simulation data related to `external drives` and `rec_arrays` eg- `events` is deleted.
Is a similar behavior intended for read\write functionality?

hnn_core/network.py
```diff
@@ -1335,6 +1407,280 @@ def plot_cells(self, ax=None, show=True):
         """
         return plot_cells(net=self, ax=ax, show=show)
 
+    def write(self, fname):
```
jasmainak, line 1410: docstring missing, also add `overwrite` parameter

raj1701:
image
I think docstrings are a better way to document as h5tree throws some random errors

jasmainak:
Let's do docstrings for now

hnn_core/network.py
```diff
@@ -1340,6 +1415,310 @@ def plot_cells(self, ax=None, show=True):
         """
         return plot_cells(net=self, ax=ax, show=show)
 
+    def write(self, fname, overwrite=True, save_unsimulated=False):
```
jasmainak, line 1418: This is a fair amount of code, I propose moving this to a new file ... `io.py` maybe? You'd basically house all the `write_xxx` functions and `read_xxx` functions there. Then here you'd just call the `write_network` function

hnn_core/network.py
```diff
@@ -1340,6 +1415,310 @@ def plot_cells(self, ax=None, show=True):
...
+        # Tasks
+        # 1) Remove params dict (Partial done)
+        # 2) Add overwrite warnings (Done)
+        # 3) Add boolean to read/write un-simulated network (Done)
+        # 4) Add boolean to read/write simulated network (Done)
+        # 5) Think about documentation
+        # 6) Add tests
+        # 7) Store the type of object saved and issue warning (Done)
```
jasmainak, line 1426: suggest moving this to the PR description ... github has a nice checkbox feature you can use

hnn_core/network.py
jasmainak: There would then be a new test file `test_io.py`

hnn_core/cell.py
```diff
@@ -443,6 +529,9 @@ def build(self, sec_name_apical=None):
             with this section. The section should belong to the apical dendrite
             of a pyramidal neuron.
         """
+        # Confirm correctness here
+        from .network_builder import load_custom_mechanisms
+        load_custom_mechanisms()
```
jasmainak, line 534: does it need to be in the other place now that you have the function call here?

hnn_core/network.py
raj1701: Sure will do this

hnn_core/network.py
raj1701: Ooh its a pretty cool feature. I just saw it.

hnn_core/cell.py
raj1701: I'll try removing it from the other place

hnn_core/cell.py
```diff
@@ -363,9 +440,18 @@ def _set_biophysics(self, sections):
...
                     if hasattr(val, '__call__'):
                         sec.push()
+                        seg_xs, seg_vals = list(), list()
                         for seg in sec:
-                            setattr(seg, attr, val(h.distance(seg.x)))
+                            seg_xs.append(seg.x)
+                            seg_vals.append(val(h.distance(seg.x)))
+                            setattr(seg, attr, seg_vals[-1])
                         h.pop_section()
+                        p_mech[attr] = [seg_xs, seg_vals]
```
rythorpe, line 449: I'm assuming the purpose of this is to overwrite the callable stored in `p_mech`? If so, this will require calling `cell.build()` any time you need to write write a cells attached to a `Network` instance, which will in turn imply that we're going to instantiate NEURON objects from within `Network`. I'm just thinking out loud here, but will that cause any unintended issues with the NEURON network? In the past, we've maintained the convention that NEURON objects are only instantiated from `NetworkBuilder` in order to silo how, when, and where such objects are created.

raj1701: Yes, calling `cell.build()` outside the `NetworkBuilder` does throw an error later when Network is simulated. For resolving it, I first made a copy of the cell and called `build()` on the copy instead of the original.

hnn_core/io.py
```diff
@@ -0,0 +1,340 @@
...
+        Key : name of cell type
+        Value : dict version of Cell object
+    gid_ranges : dict of dict
+        key : cell name or drive name
+        Value : dict
+    pos_dict : dict
+        key : cell type name
+        value : All co-ordintes of the cell types
+
+    params description
```
jasmainak, line 46: get rid of params

hnn_core/tests/test_io.py
```diff
@@ -0,0 +1,119 @@
...
+        read_network(tmpdir.join('not_net.hdf5'))
+
+    # Checking wrong object type error
+    dummy_data['object_type'] = "net"
+    write_hdf5(tmpdir.join('not_net.hdf5'), dummy_data, overwrite=True)
+    with pytest.raises(ValueError,
+                       match="The object should be of type Network."):
+        read_network(tmpdir.join('not_net.hdf5'))
+
+    # Add test to check weights are equal in connections and drives (todo)
```
jasmainak, line 119: add test to check each attribute is documented

hnn_core/tests/test_io.py
```diff
@@ -0,0 +1,119 @@
...
+@pytest.mark.parametrize("network_model",
+                         ["law_2021", "calcium", "jones_2009"])
+def test_network_io(tmpdir, network_model):
+    # Instantiating network along with drives
+    if network_model == "jones_2009":
+        net = jones_2009_model(add_drives_from_params=True)
+    elif network_model == "law_2021":
+        net = law_2021_model(add_drives_from_params=True)
+    elif network_model == "calcium":
+        net = calcium_model(add_drives_from_params=True)
```
jasmainak, line 24:
```suggestion
net = network_model(add_drives_from_params=True)
```

hnn_core/network.py
```diff
@@ -414,6 +415,61 @@ def __repr__(self):
...
+        # Check cell_response
+        self.cell_response == other.cell_response
+
+        # Check external drives
+        if not (self.external_drives == other.external_drives):
+            return False
+
+        # Check external biases
+        if not (self.external_biases == other.external_biases):
+            return False
```
jasmainak, line 443:
```suggestion
       for attr in ['cell_response', 'external_drives']:
            if getattr(self, attr) != getattr(other, attr):
                return False
```

hnn_core/network.py
```diff
@@ -414,6 +415,61 @@ def __repr__(self):
...
+        if not (self.rec_arrays == other.rec_arrays):
+            return False
+
+        # Check threshold
+        if not (self.threshold == other.threshold):
+            return False
+
+        # Check delay
+        if not (self.delay == other.delay):
+            return False
```
jasmainak, line 469: include these as well

hnn_core/network.py
```diff
@@ -414,6 +415,61 @@ def __repr__(self):
...
+
+        # Check external biases
+        if not (self.external_biases == other.external_biases):
+            return False
+
+        # Check connectivity
+        if not (len(self.connectivity) == len(other.connectivity)):
+            return False
+        for conn in self.connectivity:
```
jasmainak, line 448:
```suggestion
        # order of connections doesn't matter
        for conn in self.connectivity:
```

hnn_core/io.py
```diff
@@ -0,0 +1,340 @@
...
+    # Write threshold
+    net_data['threshold'] = net.threshold
+    # Write delay
+    net_data['delay'] = net.delay
+
+    # Saving file
+    write_hdf5(fname, net_data, overwrite=overwrite)
+
+
+def read_network(fname, read_raw=False):
```
jasmainak, line 126: docstring ...

hnn_core/io.py
```diff
@@ -0,0 +1,340 @@
+"""Network io"""
+
+import os
+from h5io import write_hdf5, read_hdf5
+
+from .cell import Cell, Section
+from .cell_response import CellResponse
+
+
+def write_network(net, fname, overwrite=True, save_unsimulated=False):
```
jasmainak, line 10: use copy_doc and/or fill_doc. See MNE

hnn_core/io.py
```diff
@@ -0,0 +1,340 @@
+"""Network io"""
```
jasmainak, line 1: Author name missing

hnn_core/tests/test_io.py
```diff
@@ -0,0 +1,119 @@
...
+import hnn_core
+from hnn_core import read_network
+from hnn_core import jones_2009_model, law_2021_model, calcium_model
+
+hnn_core_root = op.dirname(hnn_core.__file__)
+params_fname = op.join(hnn_core_root, 'param', 'default.json')
+
+
+@pytest.mark.parametrize("network_model",
+                         ["law_2021", "calcium", "jones_2009"])
```
jasmainak, line 16:
```suggestion
                         [law_2021, calcium, jones_2009])
```

hnn_core/cell.py
rythorpe: Do we know if the NEURON objects are getting properly destroyed and de-referenced after the cell copy is destroyed by the garbage collector? The only reason I'm bringing this up is because great care needs to be taken when handling NEURON network objects which are generally global and have in the past caused caused segmentation faults that are hard to diagnose and/or make tests for.

Is there a way around this that allows us to maintain our convention of only building NEURON objects from within `NetworkBuilder`? CC'ing @jasmainak and @ntolley for their input.

jasmainak: yeah I agree we need to be careful here. See: https://github.com/jonescompneurolab/hnn-core/blob/c81e5b87fbcc13bd4c6e5a403e838ca63ea16760/hnn_core/network_builder.py#L658 

The trade-off is between doing our computation ... which can be error-prone if we don't get the positions of the segments exactly right. Neuron does some mumbo-jumbo here:

https://github.com/jonescompneurolab/hnn-core/blob/c81e5b87fbcc13bd4c6e5a403e838ca63ea16760/hnn_core/cell.py#L431-L434

vs

potentially slowing up or causing segmentation faults in all our simulations that have run the `save` method

rythorpe: The fact that the segment coordinates prior to NEURON instantiation aren't their final values (and thus need to be re-set using length modifiers) is annoying. Maybe we should rip the bandaid off and fix this, which I think would also nullify the need to build the NEURON cells in order to write the network.

rythorpe: I'm proposing the latter, which I think would allow us to get rid of the length parameter. While this may introduce small numerical differences compared to what we currently have, it's a much more sustainable solution.

jasmainak: what would be concrete steps for @raj1701 ?

1. Raise warning if length is defined in `param` file? It should say that it has no effect. Same with network attribute : they should not be modifiable
2. Update the coordinates here to match the default length parameters: https://github.com/jonescompneurolab/hnn-core/blob/c81e5b87fbcc13bd4c6e5a403e838ca63ea16760/hnn_core/cells_default.py#L69-L77
3. Remove `h.define_shape` line

Anything else I missed? Does this sound doable @raj1701 ? Any inputs @ntolley ?

rythorpe: I think there are a quite a few params in the param files that are silently ignored already, right? Since most of this stuff is already specified in [`params_default.py`](https://github.com/jonescompneurolab/hnn-core/blob/c81e5b87fbcc13bd4c6e5a403e838ca63ea16760/hnn_core/params_default.py#L286), is the param file warning necessary?

We should also remove the lines where the NEURON length and radius attributes are set (see [here](https://github.com/jonescompneurolab/hnn-core/blob/c81e5b87fbcc13bd4c6e5a403e838ca63ea16760/hnn_core/cell.py#L409C1-L412C43)).

hnn_core/cell.py
```diff
@@ -363,9 +440,18 @@ def _set_biophysics(self, sections):
                 for attr, val in p_mech.items():
                     if hasattr(val, '__call__'):
                         sec.push()
+                        seg_xs, seg_vals = list(), list()
                         for seg in sec:
-                            setattr(seg, attr, val(h.distance(seg.x)))
+                            seg_xs.append(seg.x)
+                            seg_vals.append(val(h.distance(seg.x)))
```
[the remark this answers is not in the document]
raj1701: No its not.
I will rebase it and bring it to a working condition today. Will need to remove some things.

raj1701:
Screenshot from 2023-08-30 16 54 19
I am getting this error while rebasing. Any ideas how to resolve this?

hnn_core/tests/test_io.py
```diff
@@ -0,0 +1,134 @@
+import os.path as op
+from numpy.testing import assert_allclose
```
jasmainak, line 2:
```suggestion
import os.path as op

from numpy.testing import assert_allclose
```

separate system builtin imports from other libraries

hnn_core/io.py
```diff
@@ -0,0 +1,394 @@
...
+    params = dict()
+    params['N_pyr_x'] = net_data['N_pyr_x']
+    params['N_pyr_y'] = net_data['N_pyr_y']
+    params['celsius'] = net_data['celsius']
+    params['threshold'] = net_data['threshold']
+
+    # Instantiating network
+    # Cannot do this globally due to circular import
+    from .network import Network
+    net = Network(params)
```
jasmainak, line 181: put the import at the top of the function

jasmainak: I would actually do the nested import in the `write` method

hnn_core/docs.py
```diff
@@ -0,0 +1,255 @@
...
+    has keys for parameters of the mechanism, e.g., 'e', 'tau1',
+    'tau2'.
+topology : list of list
+    The topology of cell sections. Each element is a list of
+    4 items in the format
+    [parent_sec, parent_loc, child_sec, child_loc] where
+    parent_sec and parent_loc are float between 0 and 1
+    specifying the location in the section to connect and
+    parent_sec and child_sec are names of the connecting
+    sections.
```
jasmainak, line 87: is this still true?

hnn_core/extracellular.py
```diff
@@ -354,6 +354,19 @@ def __repr__(self):
     def __len__(self):
         return len(self._data)  # length == number of trials
 
+    def __eq__(self, other):
```
ntolley, line 357: Is this tested anywhere? Couldn't find any down below

ntolley: Nevermind! I see where it shows up in the net __eq__ comparison

hnn_core/cell.py
```diff
@@ -220,6 +224,52 @@ def __init__(self, L, diam, Ra, cm, end_pts=None):
     def __repr__(self):
         return f'L={self.L}, diam={self.diam}, cm={self.cm}, Ra={self.Ra}'
 
+    def __eq__(self, other):
```
ntolley, line 227: Are there two `__eq__` functions for this class? I think one is being overridden

hnn_core/cell.py
```diff
@@ -220,6 +224,52 @@ def __init__(self, L, diam, Ra, cm, end_pts=None):
...
+        if not isinstance(other, Section):
```
ntolley, line 228: I don't believe this is tested explicitly

hnn_core/extracellular.py
```diff
@@ -354,6 +354,19 @@ def __repr__(self):
...
+        if not isinstance(other, ExtracellularArray):
+            return NotImplemented
```
ntolley, line 359: This could be tested very quickly

hnn_core/cell.py
```diff
@@ -355,6 +405,57 @@ def __repr__(self):
         class_name = self.__class__.__name__
         return f'<{class_name} | gid={self._gid}>'
 
+    def __eq__(self, other):
+        if not isinstance(other, Cell):
+            return NotImplemented
```
ntolley, line 410: needs a test

hnn_core/cell.py
```diff
@@ -220,6 +224,52 @@ def __init__(self, L, diam, Ra, cm, end_pts=None):
...
+            self_mech = self.mechs[mech_name]
+            other_mech = other.mechs[mech_name]
+            for attr in self_mech.keys():
+                if self_mech[attr] != other_mech[attr]:
+                    return False
+
+        # Check end_pts
+        for self_end_pt, other_end_pt in zip(self.end_pts, other.end_pts):
+            if self_end_pt != other_end_pt:
+                if np.testing.assert_almost_equal(self_end_pt,
```
ntolley, line 242:
```suggestion
                if not np.testing.assert_almost_equal(self_end_pt,
```
?

hnn_core/tests/test_io.py
```diff
@@ -0,0 +1,134 @@
...
+        assert_allclose(dpl1.times, dpl2.times, rtol=0.00051, atol=0)
+        for dpl_key in dpl1.data.keys():
+            assert_allclose(dpl1.data[dpl_key],
+                            dpl2.data[dpl_key], rtol=0.000051, atol=0)
+    # Writing simulated network and reading it
+    net.write(tmp_path / 'net_sim.hdf5')
+    net_sim = read_network(tmp_path / 'net_sim.hdf5')
+    assert net == net_sim
+
+    # For cell response vsec isec bug (Not checked by __eq__)
```
ntolley, line 70: I'm not sure I remember, what is the bug?

hnn_core/network.py
```diff
@@ -414,6 +416,34 @@ def __repr__(self):
...
+            match_conns = pick_connection(other,
+                                          src_gids=src_gids,
+                                          target_gids=target_gids,
+                                          loc=conn['loc'],
+                                          receptor=conn['receptor'])
+            if len(match_conns) == 0:
+                return False
+
+        # Check all other attributes
+        for attr in ['cell_types', 'gid_ranges', 'pos_dict', 'cell_response',
```
ntolley, line 439: Comment for everyone, @jasmainak @rythorpe @raj1701 

Any way we can add a safe guard to make sure the equality operation checks any new attributes added to the network? Ideally something that breaks the code when you add a public attribute not reflected in this list.

rythorpe: We could potentially use `self.__dict__` which should return a dict of all attribute names and their respective values for the given object instance.

rythorpe: Or the [`dir()`](https://docs.python.org/3/library/functions.html#dir) built-in....

hnn_core/docs.py
raj1701: No, its not. Will make this change

hnn_core/io.py
```diff
@@ -0,0 +1,394 @@
...
+
+    external drive description
+    --------------------------
+    %(external_drive_description)s
+
+    external bias description
+    -------------------------
+    %(external_bias_description)s
+
+    connection description
```
jasmainak, line 57: I don't see any of this getting rendered here: https://output.circle-artifacts.com/output/job/0f0c9fc8-ab90-49f1-90eb-558f1ecf063b/artifacts/0/html/generated/hnn_core.Network.html#hnn_core.Network.write

jasmainak: you might want to try having a separate element for each argument ... it will also maximize reusability

raj1701: I made this change. But no matter how much I try it is not rendering headings correctly. For example - I want `cell type description` as a heading same as `Parameters` but it is not rendering. I read some numpy docs that only special keywords such as `Parameters`, `Returns`, `Yields` etc are rendered as headings - https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard.

hnn_core/network.py
raj1701: Sure will make this change.

hnn_core/tests/test_io.py
raj1701: Its not a bug exactly. From what I remember `vsec` and `isec` were not getting compared in the `__eq__` method and putting this logic there was breaking some flow. I will look into it again and figure out a way to place in the `__eq__` method as that is the correct place for such comparison.

raj1701: Moved this to __eq__ in cell response

hnn_core/cell.py
raj1701: So I found out `np.testing.assert_almost_equal` returns `None` when the condition is true. `None` is treated as `false` in `if` condition

hnn_core/cell.py
raj1701: There is one each for Section and Cell classes

hnn_core/io.py
raj1701: We are using `copy doc` decorator for `write` method, hence importing `write_network` inside `net.write` is not working. Doing it inside the class leads to circular import.

hnn_core/cell.py
```diff
@@ -355,6 +405,57 @@ def __repr__(self):
...
+        if not (self.sections.keys() == other.sections.keys()):
+            return False
+
+        for key in self.sections.keys():
+            if self.sections[key] != other.sections[key]:
+                return False
+
+        return True
+
+    def to_dict(self):
```
jasmainak, line 431: public method without docstring

hnn_core/network.py
```diff
@@ -1340,6 +1386,10 @@ def plot_cells(self, ax=None, show=True):
         """
         return plot_cells(net=self, ax=ax, show=show)
 
+    @copy_doc(write_network)
+    def write(self, fname, overwrite=True, save_unsimulated=False):
```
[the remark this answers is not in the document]
jasmainak: also should `overwrite` be `False` or `True` by default?

hnn_core/cell.py
```diff
@@ -220,6 +224,49 @@ def __init__(self, L, diam, Ra, cm, end_pts=None):
...
+           self.diam != other.diam or
+           self.Ra != other.Ra or
+           self.cm != other.cm or
+           self.nseg != other.nseg or
+           self.syns != other.syns):
+            return False
+
+        return True
+
+    def to_dict(self):
```
jasmainak, line 256: public method is missing docstring

doc/network_file_desc.rst
```diff
@@ -0,0 +1,196 @@
...
+.. _network_file_content:
+
+====================
+Network File Content
+====================
+
+File Content Description
+------------------------
+
+    object_type : str
```
jasmainak, line 12: you have an extra tab level which is making the rendering appear as if it's quoted:

https://output.circle-artifacts.com/output/job/2e226fdd-4b37-4b45-84b2-23faa4488796/artifacts/0/html/network_file_desc.html

doc/network_file_desc.rst
```diff
@@ -0,0 +1,196 @@
+:orphan:
+
+.. _network_file_content:
+
+====================
+Network File Content
+====================
```
jasmainak, line 7: can you add some more information what the file structure is like? what is the file format etc? we need to provide some more information for users if they open the file using other tools and also for documentation

doc/conf.py
```diff
@@ -117,6 +117,7 @@
         ("Examples", "auto_examples/index"),
         ("GUI", "gui/index"),
         ("API", "api"),
+        ("Network File", "network_file_desc"),
```
jasmainak, line 120: don't add it here. You can add a line in the Readme linking to this file. It should say something to the effect that hnn-core uses hdf5 for IO and that the description of the file structure can be found here and the API of the functions for the IO can be found here

hnn_core/externals/mne.py
```diff
@@ -765,3 +768,254 @@ def _check_tfr_param(freqs, sfreq, method, zero_mean, n_cycles,
...
+        doc = source.__doc__
+        if func.__doc__ is not None:
+            doc += func.__doc__
+        func.__doc__ = doc
+        return func
+
+    return wrapper
+
+
+def copy_function_doc_to_method_doc(source):
```
jasmainak, line 870: is this actually used somewhere? If not let's remove. We should keep the minimum required. The more code we have, the more we have to maintain

hnn_core/io.py
```diff
@@ -0,0 +1,338 @@
...
+
+def str_to_node(node_string):
+    node_tuple = node_string.split(',')
+    node_tuple[1] = int(node_tuple[1])
+    node = (node_tuple[0], node_tuple[1])
+    return node
+
+
+@fill_doc
+def write_network(net, fname, overwrite=True, save_unsimulated=False):
```
jasmainak, line 23: you need to change `save_unsimulated`

hnn_core/io.py
```diff
@@ -0,0 +1,338 @@
...
+
+import os
+from h5io import write_hdf5, read_hdf5
+
+from .cell import Cell, Section
+from .cell_response import CellResponse
+from .externals.mne import fill_doc
+
+
+def str_to_node(node_string):
```
jasmainak, line 15: private function?

hnn_core/extracellular.py
```diff
@@ -524,6 +537,17 @@ def plot_csd(self, colorbar=True, ax=None, show=True):
 
         return fig
 
+    def to_dict(self):
```
jasmainak, line 540: missing docstring

hnn_core/docs.py
```diff
@@ -0,0 +1,551 @@
...
+pos : tuple
+    The (x, y, z) coordinates.
+"""
+
+docdict[
+    "sections"
+] = """
+sections : dict of Section
+    Dictionary with keys as section name.
+"""
```
jasmainak, line 114: let's remove entries which are not used anywhere and keep the bare minimum. This PR has become too big and I worry about maintaining the code.

hnn_core/io.py
raj1701: What was the name we decided? `save_output`?

raj1701: Then accordingly `read_raw` will become `read_output`?

hnn_core/externals/mne.py
raj1701: Sure

hnn_core/network.py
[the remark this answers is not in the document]
raj1701: Should we give parameters `write_cell_response` and `write_extracellular_array` separately. `write_output` might be better because some simulation outputs are linked to `external_drives` also.

jasmainak:
While reviewing Nick's PR I realized that we haven't updated any of the examples to use `hdf5` files. Do you think that would be possible? In other words, we need to practice what we preach

hnn_core/io.py
jasmainak: yep, exactly! `save_output` and `read_output`

hnn_core/network.py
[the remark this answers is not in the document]
jasmainak: `external_drives` are definitely not "outputs" ... I see now why it was called "unsimulated". But we should do the following:

when writing: `save_output=False` will save network and external drives but not extracellular array and cell response.

when reading, two different parameters: `read_drives=True` will also read the external drives and `read_output=True` will also read the extracellular array and cell response

raj1701:
> While reviewing Nick's PR I realized that we haven't updated any of the examples to use `hdf5` files. Do you think that would be possible? In other words, we need to practice what we preach

Surely can be done. We can have a discussion on this tomorrow

hnn_core/network.py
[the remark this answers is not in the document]
raj1701: If `read_drives=False` we won't be attaching drives at all?

hnn_core/cell.py
raj1701: `assert (net == "net") == False.`
Is this test fine?

doc/network_file_desc.rst
```diff
@@ -0,0 +1,200 @@
...
+
+File Content Description
+------------------------
+hdf5 is the file format used for storing the Network object. The network is stored in a layered format. The first layer consists of the network attributes.
```
ntolley, line 11:
```suggestion
hdf5 is the file format used for storing the Network object. The network is stored in a multi-level format. The first level consists of the network attributes.
```

Using "layer" as the terminology might get confusing, as there is the biologically "layers" of the model

doc/network_file_desc.rst
```diff
@@ -0,0 +1,200 @@
...
+The attributes of the network are then broken down until the network can be representated as a collection of key value pairs. For example - cell_types is a network
+attribute therefore in the first layer. The description of each cell type is in layer 2. Each cell has various sections. The description of a section is in layer 3.
```
ntolley, line 13:
```suggestion
attribute therefore in the first level. The description of each cell type is in level 2. Each cell has various sections. The description of a section is in level 3.
```

doc/network_file_desc.rst
```diff
@@ -0,0 +1,200 @@
...
+Below are the attributes of the network.
+
+object_type : str
+    Type of object (Network) saved.
+N_pyr_x : int
+    Nr of cells (x).
+N_pyr_y : int
+    Nr of cells (y).
```
ntolley, line 21:
```suggestion
N_pyr_x : int
    Number of cells (x).
N_pyr_y : int
    Number of cells (y).
```

hnn_core/tests/test_io.py
```diff
@@ -0,0 +1,130 @@
...
+
+    # Adding bias
+    net.add_tonic_bias(cell_type='L2_pyramidal', amplitude=1.0)
+
+    # Test __eq__ method
+    net_copy = net.copy()
+    assert net_copy == net
+
+    # Test other not NotImplemented for Network Class
+    assert (net == "net") is False
```
ntolley, line 40:
```suggestion
    with pytest.raises(NotImplemented):
        net == "net"
```
This is a more robust test for this error

hnn_core/tests/test_extracellular.py
```diff
@@ -33,6 +33,9 @@ def test_extracellular_api():
     assert len(net.rec_arrays) == 2
     assert len(net.rec_arrays['arr1'].positions) == 2
 
+    # Test other not NotImplemented for ExtracellularArray Class
+    assert (net.rec_arrays['arr1'] == "extArr") is False
```
ntolley, line 37:
```suggestion
    with pytest.raises(NotImplemented):
        net.rec_arrays['arr1'] == "extArr"
```

hnn_core/cell.py
```diff
@@ -220,6 +224,55 @@ def __init__(self, L, diam, Ra, cm, end_pts=None):
     def __repr__(self):
         return f'L={self.L}, diam={self.diam}, cm={self.cm}, Ra={self.Ra}'
 
+    def __eq__(self, other):
+        if not isinstance(other, Section):
+            return NotImplemented
```
ntolley, line 229: Oh I think this is why the pytest.raises test didn't work. You're returning `NotImplemented`, but you actually need to do `raise NotImplemented` (check out how the other errors get raised)

jasmainak:
@raj1701 can you make the CIs happy? you have `flake8` errors

gtdang:
@ntolley @jasmainak 
I've changed the base of this PR to new branch on hnn-core so that we can more easily make edits before merging into the master. Shall we go ahead and merge this into the new branch so I can begin work on updating this PR?

ntolley:
@gtdang can you open a PR with the new branch? I've figured out the problem you found in `test_io.py` and it'd be best to discuss on the new branch directly.

gtdang:
> @gtdang can you open a PR with the new branch? I've figured out the problem you found in `test_io.py` and it'd be best to discuss on the new branch directly.

Ok! the new PR is here https://github.com/jonescompneurolab/hnn-core/pull/704

PR #661, opened 2023-07-03
[MRG] Find end pts of section and a numpy version of h.distance()

raj1701:


hnn_core/find_coords.py
```diff
@@ -0,0 +1,16 @@
...
+
+from hnn_core import pyramidal
+
+
+def find_coords(func_name, cell_name):
+    cell = func_name(cell_name=cell_name)
+    # print(cell.sections['soma'].end_pts)
+    for section_name, section in cell.sections.items():
+        print(section_name)
+        print(section.end_pts)
```
raj1701, line 11: The `end_pts` printed here are different from those defined in `cells_default` for cell `L2Pyr`. 
Screenshot from 2023-07-03 01 42 33

Do I simply need to replace these `end_pts` in `cells_default` or should I build the cell and find them using `cell._nrn_sections`

jasmainak: You need to build the cell and get the coords from that

raj1701: When any new cell is created using `Cell` class, a private function `__update_end_pts()` is called at the end of `__init__`. This function creates neuron sections by calling `_create_sections()` then updates the `end_pts` of the `section`. The neuron objects are then deleted using `del self._nrn_sections[name]`.

I have a few questions regarding this flow -
- Neuron objects are created and destroyed in this flow which according to the discussion in last meeting should be avoided outside of simulation.
- If this functionality is already there and the `end_pts` are automatically updated, is manually updating the `end_pts` required ?
- Should this function `_update_end_pts()` be removed?

raj1701: Also if `neuron objects` can be cleared using `del`, can't we follow a similar scheme while saving the network? After calling `build` on the copy of a cell, can we just iteratively delete the `_nrn_sections` and `_nrn_synapses` of the built cell??

jasmainak: The problem is that Neuron objects are wrappers around C objects ... when `del` is called, it makes a request to the garbage collector to clear the object from memory. It's a request, not an order. If a pointer is not cleared from memory because of how it is coded in C, it will lead to a segmentation fault and/or slow down the following simulation. The latter will be very hard to debug ... we do have a scheme to clear neuron objects in `NetworkBuilder` but it was contributed by someone really in the weeds who understands Neuron. 

But you have a good point about `_update_end_pts`  ... I think @ntolley implemented that. Not sure how well it's been tested if the `del` in that function does not lead to a performance regression. Any thoughts Nick?

raj1701:
As discussed in the meeting, I stored a base network and dipole and checked equality of them after removing `_update_end_pts()` and `h.define_shape()` .  Here are some observations -
- It seems that both `_update_end_pts()` and `h.define_shape()` are required in order to obtain the final `end_pts`. Removal of either one of them lead to a mismatch
- The reason is `_update_end_pts()` is dependent on `h.define_shape()`. Also the `end_pts` are only updated once when the cell is initialised. This function is only used by `modify_section()` which itself is not used anywhere.

raj1701:
I think if all `end_pts` are updated in `cells_default`, building any neuron objects before simulation may be avoided. I will try this now.

hnn_core/cell.py
```diff
@@ -696,6 +775,132 @@ def _update_end_pts(self):
 
         self._nrn_sections = dict()
 
+    def create_cell_tree(self):
+        """Create cell tree from topology
+
+        Returns
+        -------
+        Tree representation of the cell.
```
jasmainak, line 783: clearly explain the data structure

hnn_core/cell.py
```diff
@@ -696,6 +775,132 @@ def _update_end_pts(self):
...
+        y = self.sections[sec_name].end_pts[end_num][1]
+        z = self.sections[sec_name].end_pts[end_num][2]
+        self.sections[sec_name].end_pts[end_num][0] = x + dpt[0]
+        self.sections[sec_name].end_pts[end_num][1] = y + dpt[1]
+        self.sections[sec_name].end_pts[end_num][2] = z + dpt[2]
+        if (sec_name, end_num) in self.cell_tree.keys():
+            for (child_name, conn) in self.cell_tree[(sec_name, end_num)]:
+                self._update_section_end_pts_L(child_name, conn, dpt)
+
+    def _tree_traversal(self, sec_name, flag_start):
```
jasmainak, line 842: method name appropriate?

hnn_core/cell.py
```diff
@@ -341,6 +360,30 @@ def gid(self, gid):
         else:
             raise RuntimeError('Global ID for this cell already assigned!')
 
+    def distance_section(self, target_sec_name, base_sec_name, base_sec_end):
+        if (base_sec_name, base_sec_end) not in self.cell_tree.keys():
+            return 100000  # Ask about this
+        if (((target_sec_name, 0) in (
+            self.cell_tree[(base_sec_name, base_sec_end)])) or (
+                ((target_sec_name, 1) in (
+                    self.cell_tree[(base_sec_name, base_sec_end)])))):
```
jasmainak, line 369: simplify ...

jasmainak:
@raj1701 let me know when I should look !

hnn_core/cell.py
```diff
@@ -202,6 +203,14 @@ def __init__(self, L, diam, Ra, cm, end_pts=None):
         self.mechs = dict()
         self.syns = list()
 
+        # For distance functionality
+        self.nseg = 1
```
jasmainak, line 207: you need to document it

hnn_core/cell.py
```diff
@@ -322,7 +336,15 @@ def __init__(self, name, pos, sections, synapses, topology, sect_loc,
         if gid is not None:
             self.gid = gid  # use setter method to check input argument gid
 
-        self._update_end_pts()
+        # This will store the tree representation of the cell4
+        self.cell_tree = dict()
+        # Create the tree
```
jasmainak, line 341:
```suggestion
```

comment is pretty redundant with the line below

hnn_core/cell.py
```diff
@@ -286,6 +295,11 @@ class Cell:
     sect_loc : dict of list
         Can have keys 'proximal' or 'distal' each containing
         names of section locations that are proximal or distal.
+    cell_tree : dict of list
+        Stores the tree representation of a cell. Built from topology.
+        Root is the 0 end of 'soma'. Nodes are a tuple (sec_name, sec_end_pt)
+        where sec_name is the name of the section and sec_end_pt is the o end
```
jasmainak, line 301:
```suggestion
        where sec_name is the name of the section and sec_end_pt is the 0 end
```

hnn_core/cell.py
```diff
@@ -286,6 +295,11 @@ class Cell:
     sect_loc : dict of list
         Can have keys 'proximal' or 'distal' each containing
         names of section locations that are proximal or distal.
+    cell_tree : dict of list
+        Stores the tree representation of a cell. Built from topology.
+        Root is the 0 end of 'soma'. Nodes are a tuple (sec_name, sec_end_pt)
```
jasmainak, line 300:
```suggestion
        Root is the 0 end of 'soma'. Nodes are a tuple (sec_name, node_pos)
```

node_pos = node position

more correct / easier to understand?

hnn_core/cell.py
```diff
@@ -696,6 +787,163 @@ def _update_end_pts(self):
...
+        of the cell.
+        """
+        # Find the end pts of the section
+        flag_end = int(not flag_start)
+        pts = self.sections[sec_name].end_pts
+        x0, y0, z0 = pts[flag_start][0], pts[flag_start][1], pts[flag_start][2]
+        x1, y1, z1 = pts[flag_end][0], pts[flag_end][1], pts[flag_end][2]
+
+        # Find the factor by which length is changed
+        old_len = math.sqrt((x1 - x0) ** 2 + (y1 - y0) ** 2 + (z1 - z0) ** 2)
```
jasmainak, line 893: just use `np`

hnn_core/cell.py
```diff
@@ -696,6 +787,163 @@ def _update_end_pts(self):
...
+
+        cell_tree[soma_0_end] = list()
+        # Connect end 1 to end 0
+        cell_tree[soma_0_end].append(soma_1_end)
+
+        end_pts_used.add(soma_0_end)
+        end_pts_used.add(soma_1_end)
+
+        # Add all connections
+        for connection in self.topology:
```
jasmainak, line 815: if we got rid of `self.topology` would we need this function? how does the `self.cell_tree` look for a pyramidal cell? or should we still keep `self.topology`? They seem to be redundant ...

hnn_core/cell.py
```diff
@@ -202,6 +203,14 @@ def __init__(self, L, diam, Ra, cm, end_pts=None):
         self.mechs = dict()
         self.syns = list()
 
+        # For distance functionality
```
ntolley, line 206: For this function as well as any relevant ones below, can you add a comment with a URL to the relevant C++ code in NEURON? Think it will be useful to have the origin of this documented

hnn_core/cell.py
```diff
@@ -341,6 +363,39 @@ def gid(self, gid):
         else:
             raise RuntimeError('Global ID for this cell already assigned!')
 
+    def distance_section(self, target_sec_name, base_sec_name, base_sec_end):
```
jasmainak, line 366:
```suggestion
    def distance_section(self, target_sec_name, base_node):
```

to be consistent with the tree representation? seems to simplify lot of stuff below if you stick to well-defined data structure

hnn_core/cell.py
```diff
@@ -341,6 +363,39 @@ def gid(self, gid):
...
+        if (base_sec_name, base_sec_end) not in self.cell_tree.keys():
```
jasmainak, line 367:
```suggestion
        if base_node not in self.cell_tree:
```

hnn_core/cell.py
```diff
@@ -341,6 +363,39 @@ def gid(self, gid):
...
+            return 1000000  # Ask about this
+
+        # Children of the current section
+        curr_sec_children = self.cell_tree[(base_sec_name, base_sec_end)]
```
jasmainak, line 371:
```suggestion
        curr_sec_children = self.cell_tree[base_node]
```

hnn_core/cell.py
```diff
@@ -286,6 +295,11 @@ class Cell:
     sect_loc : dict of list
         Can have keys 'proximal' or 'distal' each containing
         names of section locations that are proximal or distal.
+    cell_tree : dict of list
```
jasmainak, line 298: what are the keys and values of the dictionary?

hnn_core/cell.py
```diff
@@ -341,6 +363,39 @@ def gid(self, gid):
...
+        # As distances are measured till the centre of the target section
+        for end_pt in end_pts:
+            if (target_sec_name, end_pt) in curr_sec_children:
+                return self.sections[target_sec_name].L / 2
+
+        dist = 1000000  # Return large value
+
+        # Recursion to find distance
+        for sec_name, sec_end_pt in self.cell_tree[(base_sec_name,
+                                                    base_sec_end)]:
```
jasmainak, line 387:
```suggestion
        for node in self.cell_tree[base_node]:
```

hnn_core/cell.py
```diff
@@ -341,6 +363,39 @@ def gid(self, gid):
...
+            if (sec_name == base_sec_name):
+                dist_1 = (self.distance_section(target_sec_name, sec_name,
+                          sec_end_pt) + self.sections[sec_name].L)
+            else:
+                dist_1 = self.distance_section(target_sec_name,
+                                               sec_name, sec_end_pt)
```
jasmainak, line 393:
```suggestion
                dist_1 = self.distance_section(target_sec_name, node)
```

I don't fully follow the logic of these if/else statements

hnn_core/cell.py
```diff
@@ -696,6 +787,163 @@ def _update_end_pts(self):
...
+        dx = x_new - x1
+        dy = y_new - y1
+        dz = z_new - z1
+        dpt = [dx, dy, dz]
+
+        # Update all coordinates in the subtree
+        self._update_section_end_pts_L(sec_name, flag_end, dpt)
+
+        # Check for change in section lengths in the subtree
+        if (sec_name, flag_start) in self.cell_tree.keys():
```
jasmainak, line 910:
```suggestion
        test_node = (sec_name, flag_start)
        if test_node in self.cell_tree:
```

hnn_core/cell.py
```diff
@@ -696,6 +787,163 @@ def _update_end_pts(self):
...
+            for (child_name, coord) in self.cell_tree[(sec_name, flag_start)]:
```
jasmainak, line 911:
```suggestion
            for child_node in self.cell_tree[test_node]:
```

jasmainak (changes requested):
I spent some time reading the code ... everything isn't fully clear to me yet but it's getting there! Let's iterate a bit more so we can clean it up nicely. I'm passing the ball back to you @raj1701 . Let me know as soon as you do your pass (in a day or two?) and I'll read it one more time

rythorpe:
This looks like a recreation of the NEURON functionality, where the cell sections are first defined with endpoints, attached to each other into a "tree", and then morphed according a specified length parameter for each section. This is exactly what we're trying to avoid, which can be accomplished by specifying the final endpoints to begin with.

jasmainak:
@rythorpe this was done so as to avoid calling Neuron functions when saving the network ... right now, it's all implemented in pure python, so the ionic concentrations are known without using Neuron. I remember you were concerned about dangling neuron objects. This takes care of that issue. Furthermore, we better understand how the neuron distance function works :)

rythorpe:
Yes, I thought the whole goals was just to update the hardcoded endpoints because it'd be the most lightweight solution. I don't understand why we needed to re-implement everything in Python.

jasmainak:
hardcoding wouldn't have achieved what you wanted because of:

https://github.com/jonescompneurolab/hnn-core/blob/c81e5b87fbcc13bd4c6e5a403e838ca63ea16760/hnn_core/cell.py#L682-L697

see it already calls neuron functions. You'd have deprecate this ...

hnn_core/tests/test_cell.py
```diff
@@ -103,6 +104,54 @@ def test_cell():
...
+    end_pts_new = list()
+    for sec_name in cell1.sections.keys():
+        section = cell1.sections[sec_name]
+        end_pts_original.append(section.end_pts)
+        section._L = section._L * 2
+        cell1.sections[sec_name] = section
+    cell1.update_end_pts()
+    for sec_name in cell1.sections.keys():
+        section = cell1.sections[sec_name]
+        section._L = section._L / 2
```
ntolley, line 120: Why are we dividing by two after calling update endpoints? wouldn't this lead to a mismatch?

ntolley: Definitely think this is the right format for the test, and the explicit checks below are good too, but not sure I totally follow all the logic

hnn_core/cell.py
raj1701: Sure will add them

hnn_core/cell.py
raj1701: Should I rename all instances of `sec_end_pt` as `node_pos`?

hnn_core/cell.py
```diff
@@ -696,6 +787,163 @@ def _update_end_pts(self):
...
+            End pt of the section closer to the soma
+
+        Note
+        ----
+        Using sec_name as 'soma' and flag_start as 0 checks for changes
+        in any section length of the cell as (soma, 0) is the root node
+        of the cell.
+        """
+        # Find the end pts of the section
+        flag_end = int(not flag_start)
```
ntolley, line 887: Think of making the indexing more explicit, even with if/else statements, rather than boolean logic

hnn_core/cell.py
raj1701: The algorithm searches for the destination nodes (end 0 or 1 of the target section) in all available paths. If the target section is in the subtree of the current node dist_1 (now named it `dist_temp`) will contain distance between the current node to the centre of the target section

hnn_core/tests/test_cell.py
raj1701: We divide all section lengths by 2 to get back the original end points because earlier we had doubled them and updated end_pts in line 117.

hnn_core/cell.py
raj1701: Added this

hnn_core/cell.py
```diff
@@ -848,50 +849,55 @@ def create_cell_tree(self):
...
+        # Python version of Neuron define_shape function
+        # https://nrn.readthedocs.io/en/latest/python/modelspec/programmatic/topology/geometry.html?highlight=pt3dadd#pt3dadd  # noqa
         # Find the end pts of the section
-        flag_end = int(not flag_start)
-        pts = self.sections[sec_name].end_pts
-        x0, y0, z0 = pts[flag_start][0], pts[flag_start][1], pts[flag_start][2]
-        x1, y1, z1 = pts[flag_end][0], pts[flag_end][1], pts[flag_end][2]
+        node_opp_end = 1
+        if node[1] == 1:
+            node_opp_end = 0
```
ntolley, line 890: Admittedly uglier but thank you for humoring me :1st_place_medal:

hnn_core/cells_default.py
```diff
@@ -178,19 +180,22 @@ def _cell_L5Pyr(override_params, pos=(0., 0., 0), gid=0.):
...
+        ('soma', 1): [('apical_trunk', 0)],
+        ('apical_trunk', 0): [('apical_trunk', 1)],
+        ('apical_trunk', 1): [('apical_1', 0), ('apical_oblique', 0)],
+        ('apical_1', 0): [('apical_1', 1)],
+        ('apical_1', 1): [('apical_2', 0)],
+        ('apical_2', 0): [('apical_2', 1)],
+        ('apical_2', 1): [('apical_tuft', 0)],
+        ('apical_tuft', 0): [('apical_tuft', 1)],
+        ('apical_oblique', 0): [('apical_oblique', 1)],
+        ('basal_1', 0): [('basal_1', 1)],
```
jasmainak, line 194: for sake of clarity, do you mind separating the connections where the section name is same with those that connect different sections? Something like:

```py
    cell_tree = {
        ('apical_1', 0): [('apical_1', 1)],
        ('basal_1', 0): [('basal_1', 1)],
        ('apical_2', 0): [('apical_2', 1)],
        ('apical_tuft', 0): [('apical_tuft', 1)],
        ('apical_oblique', 0): [('apical_oblique', 1)],
        # different sections connected
        ('soma', 0): [('soma', 1), ('basal_1', 0)],
        ('soma', 1): [('apical_trunk', 0)],
        ('apical_trunk', 0): [('apical_trunk', 1)],
        ('apical_trunk', 1): [('apical_1', 0), ('apical_oblique', 0)],
        ('apical_1', 1): [('apical_2', 0)],
        ('apical_2', 1): [('apical_tuft', 0)]
}
```

hnn_core/cell.py
```diff
@@ -361,14 +419,48 @@ def _set_biophysics(self, sections):
...
+                                               (i * adjacent_seg_dist))
+
+                        for seg_x in seg_centres:
+                            # sec_end_dist is distance between 0 end of soma to
+                            # the 0 or 1 end of section (whichever is closer)
+                            sec_end_dist = section_distance - (section.L / 2)
+                            seg_xs.append(seg_x)
+                            seg_vals.append(val(sec_end_dist +
+                                                (seg_x * section.L)))
+                        p_mech[attr] = [seg_xs, seg_vals]
```
jasmainak, line 459: can this logic be moved here: https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/cells_default.py ?

So we don't use the `partial` in the first place and directly store the values when creating the cell.

hnn_core/cell.py
```diff
@@ -361,14 +419,48 @@ def _set_biophysics(self, sections):
...
+                        first_seg_centre = (0.5 - (((section.nseg - 1) / 2) *
+                                            adjacent_seg_dist))
+                        seg_centres = list()
+
+                        # Finding centres of all segments in the section
+                        # If number of segments is 5 then seg_centres will
+                        # be 0.1, 0.3, 0.5, 0.7 and 0.9.
+                        for i in range(0, section.nseg):
+                            seg_centres.append(first_seg_centre +
+                                               (i * adjacent_seg_dist))
```
jasmainak, line 450: any chance we can use `np.linspace` ? That seems exactly for this purpose

hnn_core/cells_default.py
```diff
@@ -95,19 +95,21 @@ def _cell_L2Pyr(override_params, pos=(0., 0., 0), gid=0.):
...
-        ['apical_1', 1, 'apical_tuft', 0],
-        # apical_oblique comes off distal end of apical_trunk
-        ['apical_trunk', 1, 'apical_oblique', 0],
-        # Proximal (basal)
-        ['soma', 0, 'basal_1', 0],
-        ['basal_1', 1, 'basal_2', 0],
-        ['basal_1', 1, 'basal_3', 0]
-    ]
+    # Node description - (section_name, end_point)
+    cell_tree = {
```
jasmainak, line 99: same comment as https://github.com/jonescompneurolab/hnn-core/pull/661/files#r1300563151

hnn_core/tests/test_cell.py
```diff
@@ -103,6 +106,59 @@ def test_cell():
...
+        cell1.sections[sec_name] = section
+
+    end_pts_new = list()
+    cell1.update_end_pts()
+    for sec_name in cell1.sections.keys():
+        section = cell1.sections[sec_name]
+        cell1.sections[sec_name] = section
+        end_pts_new.append(section.end_pts)
+    # print(end_pts_original)
+    # print(end_pts_new)
```
jasmainak, line 142:
```suggestion
```

remove any commented lines to make ready for merge !

hnn_core/tests/test_cell.py
```diff
@@ -103,6 +106,59 @@ def test_cell():
...
+    cell1.plot_morphology(show=False)
+    # Checking equaliy till 5 decimal places
```
jasmainak, line 144:
```suggestion
    # Checking equality till 5 decimal places
```

hnn_core/cell.py
```diff
@@ -696,6 +793,115 @@ def _update_end_pts(self):
...
+    def define_shape(self, node):
+        """Redefines end_pts according to section lengths.
+        Detects change in section lengths of the sections in the
+        subtree of the input node.
+
+        Parameters
+        ----------
+        node : tuple of size 2
+        The first element is the section name
+        The second element is the node end used (0 or 1)
```
jasmainak, line 823: needs to be indented

hnn_core/cell.py
```diff
@@ -696,6 +793,115 @@ def _update_end_pts(self):
...
+        # Update all coordinates in the subtree
+        self._update_section_end_pts_L((node[0], node_opp_end), dpt)
+
+        # Check for change in section lengths in the subtree
+        if node in self.cell_tree:
+            for child_node in self.cell_tree[node]:
+                self.define_shape(child_node)
+
+    def update_end_pts(self):
+        """Udpate all end pts according to the length of the sections.
```
jasmainak, line 870:
```suggestion
        """Update all end pts according to the length of the sections.
```

hnn_core/cell.py
```diff
@@ -361,14 +419,48 @@ def _set_biophysics(self, sections):
...
-                        sec.push()
-                        for seg in sec:
-                            setattr(seg, attr, val(h.distance(seg.x)))
-                        h.pop_section()
+                    if isinstance(val, list):
+                        seg_xs, seg_vals = val[0], val[1]
+                        for seg, seg_x, seg_val in zip(sec, seg_xs, seg_vals):
+                            # Checking equaliy till 5 decimal places
+                            np.testing.assert_almost_equal(seg.x, seg_x, 5)
+                            setattr(seg, attr, seg_val)
```
jasmainak, line 427: let's be brave and delete `_set_biophysics`

rythorpe: Don't we still need this function for inserting nrn mechanisms and such?

jasmainak: I think `_set_section_mechs` is doing that now. This method only exists for checking equality with the old stuff

hnn_core/cells_default.py
raj1701: Sure no problem

hnn_core/cell.py
raj1701: Very slight precision errors are there. But it is passing the check in `_set_biophysics()`

hnn_core/cell.py
raj1701: I think it is still required as it is inserting mechanisms in the NEURON section objects.
`_set_section_mechs()` is just expanding the partial functions for now. If we include the `_set_biophysics()` logic here then NEURON objects might still be created before simulating the network (in case of network write)

ntolley:
image
I'm a bit stumped as to why this is only failing on windows...

hnn_core/cell.py
```diff
@@ -361,14 +419,48 @@ def _set_biophysics(self, sections):
...
                     else:
                         setattr(sec, attr, val)
 
+    def _set_section_mechs(self):
```
jasmainak, line 431:
```suggestion
    def _compute_section_mechs(self):
```

hnn_core/cell.py
```diff
@@ -361,14 +419,48 @@ def _set_biophysics(self, sections):
...
+                        for seg_x in seg_centres:
+                            # sec_end_dist is distance between 0 end of soma to
+                            # the 0 or 1 end of section (whichever is closer)
+                            sec_end_dist = section_distance - (section.L / 2)
+                            seg_xs.append(seg_x)
+                            seg_vals.append(val(sec_end_dist +
+                                                (seg_x * section.L)))
+                        p_mech[attr] = [seg_xs, seg_vals]
+                    else:
+                        p_mech[attr] = val
```
jasmainak, line 461:
```suggestion
```

jasmainak:
@raj1701 please rebase this PR so that the CIs pass

hnn_core/cell.py
```diff
@@ -361,14 +419,46 @@ def _set_biophysics(self, sections):
...
+                        # If number of segments is 5 then seg_centres will
+                        # be 0.1, 0.3, 0.5, 0.7 and 0.9.
+                        adjacent_seg_dist = 1 / section.nseg
+                        first_seg_centre = (0.5 - (((section.nseg - 1) / 2) *
+                                            adjacent_seg_dist))
+                        last_seg_centre = (0.5 + (((section.nseg - 1) / 2) *
+                                           adjacent_seg_dist))
+                        seg_centres = list(np.linspace(first_seg_centre,
+                                                       last_seg_centre,
+                                                       num=section.nseg))
```
jasmainak, line 450:
```suggestion
            seg_centers = np.linspace(0, 1, section.nseg * 2 + 1)[1::2]
```

does this work?

hnn_core/cell.py
```diff
@@ -202,6 +204,14 @@ def __init__(self, L, diam, Ra, cm, end_pts=None):
         self.mechs = dict()
         self.syns = list()
 
+        # For distance functionality
+        self.nseg = 1
+        if self.L > 100.:  # 100 um
+            self.nseg = int(self.L / 50.)
+            # make dend.nseg odd for all sections
+            if not self.nseg % 2:
+                self.nseg += 1
```
jasmainak, line 213: you need to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) your code:

https://github.com/raj1701/hnn-core/blob/neuron_coordinates/hnn_core/cell.py#L504-L508

make a private function `_get_nseg` that you use in both parts of the code.

hnn_core/cell.py
```diff
@@ -443,7 +536,9 @@ def build(self, sec_name_apical=None):
             with this section. The section should belong to the apical dendrite
             of a pyramidal neuron.
         """
-        self._create_sections(self.sections, self.topology)
+        from .network_builder import load_custom_mechanisms
+        load_custom_mechanisms()
```
jasmainak, line 540: is this for testing purposes? Do you need it?

hnn_core/cell.py
```diff
@@ -341,6 +364,41 @@ def gid(self, gid):
         else:
             raise RuntimeError('Global ID for this cell already assigned!')
 
+    def distance_section(self, target_sec_name, curr_node):
```
jasmainak, line 367: make private since not documented?

hnn_core/cell.py
```diff
@@ -322,7 +339,13 @@ def __init__(self, name, pos, sections, synapses, topology, sect_loc,
         if gid is not None:
             self.gid = gid  # use setter method to check input argument gid
 
-        self._update_end_pts()
+        # Store the tree representation of the cell
+        self.cell_tree = cell_tree
+
+        # self._update_end_pts()  # Old implementation
+        self.update_end_pts()  # New implementation
```
jasmainak, line 346: do you still need the method `_update_end_pts` ? We should consistently update everywhere and remove the old method if not needed

hnn_core/cell.py
```diff
@@ -696,6 +791,115 @@ def _update_end_pts(self):
...
+        if node not in self.cell_tree:
+            return
+
+        # If current node is an internal node
+        for child_node in self.cell_tree[node]:
+            self._update_section_end_pts_L(child_node, dpt)
+
+    def define_shape(self, node):
+        """Redefines end_pts according to section lengths.
+        Detects change in section lengths of the sections in the
```
jasmainak, line 814:
```suggestion
        """Redefines end_pts according to section lengths.

        Detects change in section lengths of the sections in the
```

hnn_core/cell.py
```diff
@@ -696,6 +791,115 @@ def _update_end_pts(self):
...
+        self._update_section_end_pts_L((node[0], node_opp_end), dpt)
+
+        # Check for change in section lengths in the subtree
+        if node in self.cell_tree:
+            for child_node in self.cell_tree[node]:
+                self.define_shape(child_node)
+
+    def update_end_pts(self):
+        """Update all end pts according to the length of the sections.
+        Can be used whenever length of any section is updated
```
jasmainak, line 869:
```suggestion
        """Update all end pts according to the length of the sections.
 
        Can be used whenever length of any section is updated
```

hnn_core/cell.py
```diff
@@ -256,6 +259,14 @@ class Cell:
         Each cell in a network is uniquely identified by it's "global ID": GID.
         The GID is an integer from 0 to n_cells, or None if the cell is not
         yet attached to a network. Once the GID is set, it cannot be changed.
+    cell_tree : dict of list
+        Stores the tree representation of a cell.
+        Root is the 0 end of 'soma'. Nodes are a tuple (sec_name, node_pos)
+        where sec_name is the name of the section and node_pos is the 0 end
+        or 1 end. The data structure is the adjacency list represetation of a
+        tree. The keys of the dict are the parent nodes. The value is the
+        list of nodes (chldren nodes) connected to the parent node.
```
ntolley, line 268:
```suggestion
        list of nodes (children nodes) connected to the parent node.
```

hnn_core/cell.py
```diff
@@ -286,6 +297,13 @@ class Cell:
     sect_loc : dict of list
         Can have keys 'proximal' or 'distal' each containing
         names of section locations that are proximal or distal.
+    cell_tree : dict of list
+        Stores the tree representation of a cell.
+        Root is the 0 end of 'soma'. Nodes are a tuple (sec_name, node_pos)
+        where sec_name is the name of the section and node_pos is the 0 end
+        or 1 end. The data structure is the adjacency list represetation of a
+        tree. The keys of the dict are the parent nodes. The value is the
+        list of nodes (chldren nodes) connected to the parent node.
```
ntolley, line 306:
```suggestion
        list of nodes (children nodes) connected to the parent node.
```

hnn_core/cell.py
```diff
@@ -681,7 +776,7 @@ def plot_morphology(self, ax=None, color=None, show=True):
 
     def _update_end_pts(self):
```
ntolley, line 777: To be clear this function should be deleted right?

hnn_core/cell.py
```diff
@@ -341,6 +364,41 @@ def gid(self, gid):
...
+        dist = 1000000  # Return large value
+
+        # Recursion to find distance
+        for node in self.cell_tree[curr_node]:
+            if (node[0] == curr_node[0]):
+                dist_temp = (self.distance_section(target_sec_name, node) +
+                             self.sections[node[0]].L)
+            else:
+                dist_temp = self.distance_section(target_sec_name, node)
+            dist = min(dist, dist_temp)
```
ntolley, line 398:
```suggestion
            dist = np.nanmin(dist, dist_temp)
```

hnn_core/cell.py
```diff
@@ -341,6 +364,41 @@ def gid(self, gid):
...
             raise RuntimeError('Global ID for this cell already assigned!')
 
+    def distance_section(self, target_sec_name, curr_node):
+        # Python version of the Neuron distance function
+        # https://nrn.readthedocs.io/en/latest/python/modelspec/programmatic/topology/geometry.html#distance  # noqa
+        if self.cell_tree is None:
+            raise TypeError("distance_section() "
+                            "cannot work with cell_tree as None.")
+        if curr_node not in self.cell_tree:
+            return 1000000
```
ntolley, line 374:
```suggestion
            return np.nan
```

jasmainak:
@raj1701 let us know when we can look again!

hnn_core/cell.py
raj1701: Yes it does

hnn_core/cell.py
raj1701: It throws a warning when all nan values are passed to the function. The results are correct. This warning is fine right?

raj1701: For now I have added a check to resolve this issue.

jasmainak (approved):
Looks good to me! @rythorpe or @ntolley do you want to merge?

rythorpe:
I won't be able to look this over until tomorrow afternoon. If we don't feel like waiting until then, feel free to merge without me @ntolley.

jasmainak:
I went ahead and merged to keep the momentum! @raj1701 let's bring your other PR to merge state as soon as possible :)

PR #701, opened 2023-12-25
Hnn som

wagdy88:
Creating a function to rename cells


rythorpe:
Great, let's set this one to "[WIP]" as well and keep it on the back burner until #702 is complete.

PR #722, opened 2024-02-28
Gui hdf5 io

### LINKED ISSUE #708: GUI loading params from hdf5
Add support for loading connectivity and drive parameters from a hdf5 file. 

**Input buttons**
  1. Load connection 
        1. hdf5 only
        2. Only loads connection information from hdf5 file
  2. Load drives
        1. hdf5 only
        3. Only loads drive information from hdf5 file 

Should incorporate planned deprecation of param files. 

**To Do:**
- [ ] Load local connectivity button support for hdf5 
- [ ] Load external drives button support for hdf5
- [ ] Param file deprecation warning

gtdang:
Add support for loading connectivity and drive parameters from an hdf5 saved network object.

closes #708


gtdang:
@ntolley 
Hit a blocker on this PR. The GUI operates on a self.param attribute that is a dict of a parameter json. 

I've gotten to the point of being able to read in the hdf5 file as a Network object and now I need to get it into the same dict format of self.param so I can update the GUI's parameter state. 

Is there a way to create a param dict/json from a Network object? If not I'll need to come up with a way to map all of a network's attributes to the correct key-value pairs.

ntolley:
@gtdang unfortunately I think the latter option is necessary. The reason we developed the hdf5 format is precisely because we didn't have a way to output HNN-core network models into a `.param` file. The ultimate goal is to break ties with the `param` format entirely, and instead operate on the `dict` used to store the network.

ntolley:
Can you point me to a line where this crops up? It may be more expedient for me to figure out that mapping since I have a good sense of what attributes in the Network object map to `.param` values.

gtdang:
@ntolley 
Sorry it took some time to respond. There isn't a single place where parameter mapping necessarily appears. To get a better sense of how the GUI utilizes the param format I've mapped the flow of the params attribute in the image below. 
hnn-gui-params-flow

gtdang:
> @ntolley Sorry it took some time to respond. There isn't a single place where parameter mapping necessarily appears. To get a better sense of how the GUI utilizes the param format I've mapped the flow of the params attribute in the image below.
> 
> The GUI functions are often initializing a Network object with the params attribute and then uses the Network object in another function to display the encoded Connections and Drives. I suppose if we wanted to do away with the params attribute we could just load the HDF5 Network to the GUI and pass that around. The function that would need to be reworked is the `add_drive_tab` and `extract_drive_specs_from_hnn_params` functions, which require the param dictionary format.
> 
> hnn-gui-params-flow

One solution to support both hdf5 and the param file is: 
* Add a new class attribute to flag if the params are in a Network or param format (`self.is_param`) 
* Add a new class attribute to save a Network from hdf5 (`self.network`)

1. In the diagram above, any function that "Initializes Network with self.params" will we rewritten to handle either param dict or a Network object. Probably will add `self.is_param` flag as an argument and logic to either initialize a Network from params dict or use `self.network`. 
2. Add a new function `extract_drive_specs_from_hnn_network`

jasmainak:
@gtdang the more complexity you add, the more maintenance work and the harder for a future contributor ... I think you guys need to take a judgement call whether to support hdf5 or param or both and whether for read/write or both. My plan was to phase out params by outputting hdf5 always (never params/json) but allowing to read json for backwards compatibility ... for a while.

gtdang:
> @gtdang the more complexity you add, the more maintenance work and the harder for a future contributor ... I think you guys need to take a judgement call whether to support hdf5 or param or both and whether for read/write or both. My plan was to phase out params by outputting hdf5 always (never params/json) but allowing to read json for backwards compatibility ... for a while.

Totally agree with keeping complexity at a minimum. I'm curious about the shift to hdf5. Is the plan for the Network to only be instantiated by an hdf5 read-in? Seems like a chicken and egg situation where you need a Network to write to hdf5 yet need to init a Network with an hdf5. 

Is the idea that users are not likely create & edit their own hdf5 Network and are always starting pre-constructed Network? Do we have a recommendations for hdf5 viewers/editors for those that want to edit params outside of the the API?

jasmainak:
We don't want to encourage users to edit params outside the API ... they are on their own if they do so. 

In the past, we had the params being passed around willy-nilly in different functions and it was hard to keep track what was happening. Modifying the Network only through the API allows us to validate the parameters while also ensuring that they are consistent ... for example start/end time of the drives cannot be outside the simulation start/end time.

Hdf5 was chosen as it's a simple hierarchical format that allows you to translate python objects into a binary format while also being future-proof (unlike pickling). Users should always start with a template network that we provide and make changes through the API. They can then save the network and load it back but they cannot make arbitrary changes to the network object.

hnn_core/tests/test_params.py
```diff
@@ -71,3 +74,71 @@ def test_base_params():
...
+    outpath = Path(tmp_path, 'default.hdf5')
+    convert_to_hdf5(params_base_fname, outpath)
+    net_hdf5 = read_network(outpath)
+    assert net_hdf5 == net_params
+
+
+def test_convert_to_hdf5_bad_type():
+    """Tests type validation in convert_to_hdf5 function"""
+    good_path = hnn_core_root
+    path_str = good_path.__str__()
```
hollandjg, line 121: You could use the `str` function rather than the dunder method here – it should do the same thing and seems a bit more standard than calling the dunder method directly (though I can't find a good reference for that – just some posts on StackExchange). 
```suggestion
    path_str = str(good_path)
```

hollandjg (commented):
This looks good! I've got a few comments – a couple of potential simplifications.

dev_scripts/convert_params.py
```diff
@@ -0,0 +1,93 @@
+"""Script converting legacy param and json parameter files to hdf5.
```
jasmainak, line 1: Is this PR still relevant. I see: https://github.com/jonescompneurolab/hnn-core/tree/master/dev_scripts already in `main` ? Or does it need a rebase?

gtdang: It's currently held up by #733. Will be rebased after that one is merged.

PR #749, opened 2024-04-16
[WIP] update convert_to_hdf5 to use Jones 2009 connectivity by default

gtdang:
Previous hdf5 conversion from param and json format did not include connectivity information because the conversion was using the base Network object to initialize the networks. This change makes the default to use the Jones 2009 model so that network connectivity is encoded in the hdf5. This was decided because the tutorial only uses this model. Other saved model types are also available as well as an option to not use any model.

### DIFF
--- a/hnn_core/params.py
+++ b/hnn_core/params.py
@@ -671,10 +671,11 @@ def convert_to_hdf5(params_fname,
         Path to file
     out_fname: str
         Path to output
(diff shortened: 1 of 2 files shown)


gtdang:
Closing this because we are moving to a new direction with saving outputs because hdf5 files in the current format are too  large.

PR #752, opened 2024-04-23
[WIP] Make all plots with time end at stop

### LINKED ISSUE #544: Make all plots with time end at `tstop`

katduecker:
Continues the PR #683 
Closes #544

### DIFF
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -468,16 +468,11 @@ class Dipole(object):
                                             self.sfreq)
         return self
 
-    def plot(self, tmin=None, tmax=None, layer='agg', decim=None, ax=None,
-             color='k', show=True):
+    def plot(self, layer='agg', decim=None, ax=None, color='k', show=True):
         """Simple layer-specific plot function.
 
         Parameters
         ----------
-        tmin : float or None
-            Start time of plot (in ms). If None, plot entire simulation.
-        tmax : float or None
-            End time of plot (in ms). If None, plot entire simulation.
         layer : str
             The layer to plot. Can be one of 'agg', 'L2', and 'L5'
         decimate : int
@@ -494,8 +489,8 @@ class Dipole(object):
         fig : instance of plt.fig
             The matplotlib figure handle.
         """
-        return plot_dipole(self, tmin=tmin, tmax=tmax, ax=ax, layer=layer,
-                           decim=decim, color=color, show=show)
+        return plot_dipole(self, ax=ax, layer=layer, decim=decim, color=color,
+                           show=show)
 
     def plot_psd(self, fmin=0, fmax=None, tmin=None, tmax=None, layer='agg',
                  color=None, label=None, ax=None, show=True):
--- a/hnn_core/dipole.py
+++ b/hnn_core/dipole.py
@@ -14,7 +14,6 @@ from .externals.mne import _check_option
 
 from .viz import plot_dipole, plot_psd, plot_tfr_morlet
 
-
 def simulate_dipole(net, tstop, dt=0.025, n_trials=None, record_vsec=False,
                     record_isec=False, postproc=False):
     """Simulate a dipole given the experiment parameters.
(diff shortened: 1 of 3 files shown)


katduecker:
Hey, I have now taken on this pull request. I'm thinking about the options for the deprecation cycle of defining tmin and tmax for plot_dipole et al.
1. Should it still be possible to set the xlims using tmax and tmin, and if tmax and tmin aren't defined then rasterplots and dipole are plotted from 0 to stop?
2. OR should the data always be plotted from 0 to tstop, and the deprecation warning tells the user to set xlim after plotting? (i.e. setting tmin and tmax doesn't do anything)

I'm happy to make all of these clear in the tutorials when I'm done!

jasmainak:
I think we should not change behavior while deprecating ... 

tmin and tmax should be set to None by default. If it is not None, then raise deprecation warning telling the user that they should set directly with plt.xlim() and plt.ylim() to avoid this warning

if it is None, then use the times attribute: extracellular.times, cell_response.times, dipole.times etc.

do the `hdf5` files now save `times` attribute @ntolley ? This was a [concern raised by @rythorpe ](https://github.com/jonescompneurolab/hnn-core/pull/683#pullrequestreview-1735875760)

Thanks for taking over the PR @katduecker !

katduecker:
Hi all,

I've been trying to push the changes but it keeps failing the CircleCI. I wonder if that is because of a deprecation?

The build_docs test gets stuck when running plot_firing_pattern.py. This is the associated error message.

Screenshot 2024-05-08 at 15 35 13

When I run plot_firing_pattern.py locally, the spike raster is plotted, but I receive this deprecation warning.
Screenshot 2024-05-08 at 15 35 57

It seems that cell_response.times (loaded in from the .txt file) is empty, but net.cell_response.times is not. Any ideas on what's going on and why it works locally, but not when doing the unit test?

Thanks!

jasmainak:
@katduecker I am able to reproduce locally. Are you sure you have the editable `hnn-core` version that you are testing? You can check by:

```py
import hnn_core
hnn_core.__version__
hnn_core.__path__
```

it should point to your github folder not to the anaconda folder.

katduecker:
Thanks Mainak. I reinstalled the developer version, cherry picked the PR and added my changes. Now it can't load custom mechanisms.. this in files that I haven't even edited, so I'm not sure what's causing this

Screenshot 2024-05-08 at 15 35 13

jasmainak:
Did you look at the [contributing guide](https://github.com/jonescompneurolab/hnn-core/blob/master/doc/contributing.rst) ?

You need to run the command:

```py 
python setup.py build_mod
```

You don't need two local installations :) Just switch branches on git ... once you have a developer install, you can switch between different versions simply by switching branches. 

In case you want to maintain a copy of hnn-core installation for reproducibility reasons, I would recommend creating a separate conda environment for your project and installing hnn-core there.

ntolley:
@katduecker I think one reason the tests are failing is because the GUI code depends on some of these plotting functions:
image

Try running `pytest test_gui.py` to see if you can reproduce the error locally

hnn_core/dipole.py
```diff
@@ -106,25 +105,20 @@ def simulate_dipole(net, tstop, dt=0.025, n_trials=None, record_vsec=False,
     return dpls
 
 
-def _read_dipole_txt(fname, extension='.txt'):
+def _read_dipole_txt(fname):
```
ntolley, line 108: @katduecker here is the problem! Since the time you had copied `dipole.py` to your desktop the master branch has changed. Therefore these modifications are undoing the latest commits to the master branch.

I would reccomend copying this function exactly as it's written on the current master branch and replacing it in your code. That way the only changes that show up in the diff are related to the `tstop` PR.

If done correctly you will no longer see `_read_dipole_txt()` show up in the files changed.

hnn_core/dipole.py
```diff
@@ -179,6 +173,10 @@ def read_dipole(fname):
         The instance of Dipole class
     """
 
+    # For supporting tests in test_gui.py
+    if isinstance(fname, StringIO):
+        return _read_dipole_txt(fname)
```
ntolley, line 178: you will want to copy the `read_dipole()` function from the master branch and replace it in your code as well

ntolley:
@katduecker I think I figured out the issue. If you're not planning on working on this over the weekend then we can sort things our in person on Monday.

The cleanest solution would be to `git reset --hard` this branch to the last commit, and then copy in the changes directly. The approach we took (copying in the entire files you had backed up) failed because the master branch was updated since the time you had backed up those files.

If you don't want to mess with the `git reset` business the alternative is to make a new commit where you copy `read_dipole` and `read_dipole_txt` directly from the master branch and paste it into your code. This will remove these functions from the diff, since these were the ones that were updated in the master branch most recently.

katduecker:
@ntolley thanks so much, that makes sense! I'll get everything ready for Monday up to the point where I'd commit the changes so that we could talk this over before I do to avoid further problems?

katduecker:
Due to issues when resetting to the commits from this PR, I have opened a new PR [#769](https://github.com/jonescompneurolab/hnn-core/pull/769)

PR #753, opened 2024-04-25
[MRG] Save simulation to csv file

kmilo9999:
Added GUI features to export simulation to a csv file
solves #718 
image

[default.txt](https://github.com/jonescompneurolab/hnn-core/files/15121821/default.txt)
[default.zip](https://github.com/jonescompneurolab/hnn-core/files/15169333/default.zip)

A few things still missing from this PR:

- [x] How to select path to folder where the file will be saved?
- [x] Select the right tab to place this "Widget"


kmilo9999:
@kmilo9999 @ntolley This PR is ready to review

hnn_core/gui/_viz_manager.py
```diff
@@ -894,6 +896,61 @@ def _layout_template_change(self, template_type):
...
+
+    def _create_zip(self, csv_data_list, simulation_name):
+        # Zip all files and keep it in memory
+        with io.BytesIO() as zip_buffer:
+            with zipfile.ZipFile(zip_buffer, 'w', zipfile.ZIP_DEFLATED) as zf:
+                for index, csv_data in enumerate(csv_data_list):
+                    zf.writestr(f'{simulation_name}_{index+1}.csv', csv_data)
+            zip_buffer.seek(0)
+            return zip_buffer.read()
+
```
gtdang, line 953: I believe these functions are better placed in gui.py as they are not related or dependent on the visualizations.

hnn_core/tests/test_gui.py
```diff
@@ -563,3 +563,41 @@ def add_child_decorated(self, to_add):
...
+    gui = HNNGUI()
+    _ = gui.compose()
+    gui.params['N_pyr_x'] = 3
+    gui.params['N_pyr_y'] = 3
+
+    # Run a simulation with 3 trials
+    gui.widget_dt.value = 0.85
+    gui.widget_ntrials.value = 3
+
+    # set synch inputs to first driver in simulation
```
gtdang, line 579: Seems like a legacy comment from copying the another test.

hnn_core/tests/test_gui.py
```diff
@@ -563,3 +563,41 @@ def add_child_decorated(self, to_add):
...
+def test_gui_download_simulation():
+    """Test the GUI download simulation pipeline."""
+    gui = HNNGUI()
+    _ = gui.compose()
+    gui.params['N_pyr_x'] = 3
+    gui.params['N_pyr_y'] = 3
+
+    # Run a simulation with 3 trials
+    gui.widget_dt.value = 0.85
+    gui.widget_ntrials.value = 3
```
gtdang, line 577: Let's do 2 here so it runs faster.

gtdang (changes requested):
Looks good Camilo! Have some minor changes. Also remember to update the whats_new.rst file. :)

hnn_core/gui/_viz_manager.py
kmilo9999: Done!

hnn_core/tests/test_gui.py
kmilo9999: Done

hnn_core/tests/test_gui.py
kmilo9999: Done

hnn_core/dipole.py
```diff
@@ -106,20 +106,26 @@ def simulate_dipole(net, tstop, dt=0.025, n_trials=None, record_vsec=False,
...
     """Read dipole values from a txt file and create a Dipole instance.
 
     Parameters
     ----------
     fname : str
-        Full path to the input file (.txt)
-
+        Full path to the input file (.txt or .csv)
+    fname: io.StringIO
+        Content of file in memory as a StringIO
```
gtdang, line 117: I think it's better convention to combine these in the docstring. 

```suggestion
    fname: str or io.StringIO
        Full path to the input file (.txt or .csv) or content of file in memory as a StringIO
```

hnn_core/gui/gui.py
```diff
@@ -229,11 +234,35 @@ def __init__(self, theme_color="#8A2BE2",
...
+        </a>
+        '''
+        # Create widget wrapper
+        self.save_simuation_button = (
+            HTML(self.html_download_button.
+                 format(payload=payload,
+                        filename={""},
+                        is_disabled="disabled",
+                        btn_height=self.layout['run_btn'].height,
+                        color_theme=self.layout['theme_color'])))
```
gtdang, line 260: This is a bit busy for the init. Can we put this into a hidden method to make the init cleaner? I don't think we need to save the html as an attribute. It can just be defined in the scope of the method. 
```suggestion
        self.save_simulation_button = self._init_html_download_button()
```

hnn_core/tests/test_gui.py
```diff
@@ -563,3 +564,42 @@ def add_child_decorated(self, to_add):
     # Check if the widgets are relinked, the selected index should be synced
     gui.tab_group_1.selected_index = 0
     assert gui.tab_group_2.selected_index == 0
+
+
+def test_gui_download_simulation():
```
gtdang, line 569: Could you add a test for csv upload in `test_gui_upload_data`?  That current test downloads test files from GitHub for the GUI testing. I don't think that's best practice as there is then a dependency on an external api. That will be change in another PR. 

For the scope of this PR you can add a test csv to `test/assets` for you test. A small/short file is preferred.

hnn_core/dipole.py
kmilo9999: Done

hnn_core/gui/gui.py
kmilo9999: I moved the init logic to a private function. I keep the html as an attribute due it needs to be updated later in the workflow

hnn_core/tests/test_gui.py
kmilo9999: Done. I created a test that uploads a csv file.

doc/whats_new.rst
```diff
@@ -55,6 +55,9 @@ Changelog
 - Added gui widget to enable/disable synchronous input in simulations, 
   by `Camilo Diaz`_ in :gh:`750`
 
+- Added gui widget to save simulations in csv format,
```
gtdang, line 58:
```suggestion
- Added gui widgets to save simulation as csv and updated the file upload to support csv data,
```

gtdang (approved):
Looks great! Thanks Camilo!

hnn_core/gui/gui.py
```diff
@@ -1378,6 +1441,73 @@ def run_button_clicked(widget_simulation_name, log_out, drive_widgets,
...
+
+
+def serialize_simulation(simulations_data, simulation_name):
+    """
+        Serializes the simulation data for a given simulation name
+        into either a single CSV file
+        or a ZIP file containing multiple CSVs, depending on the number
+        of trials in the simulation.
+        Parameters:
+        - simulation_name (str): The name of the simulation to serialize.
```
ntolley, line 1460: Can you fix the docstring to conform to the format used in the repository? You can look at any of the public functions to see examples

kmilo9999: Done!

hnn_core/gui/gui.py
```diff
@@ -1378,6 +1441,73 @@ def run_button_clicked(widget_simulation_name, log_out, drive_widgets,
...
+        logger.info(f"Saving {sim_name}.txt")
+        return serialize_simulation(sim_data, sim_name)
+
+
+def serialize_simulation(simulations_data, simulation_name):
+    """
+        Serializes the simulation data for a given simulation name
+        into either a single CSV file
+        or a ZIP file containing multiple CSVs, depending on the number
+        of trials in the simulation.
```
gtdang: The docstring format should look like this with a single line, black line, and then more detailed description. 

```suggestion
    """Serializes simulation data to CSV.
    
        Creates a a single CSV file or a ZIP file containing multiple CSVs, 
        depending on the number of trials in the simulation.
```

kmilo9999: done

hnn_core/tests/test_gui.py
```diff
@@ -563,3 +565,68 @@ def add_child_decorated(self, to_add):
...
+        # ensure we have three slashes after 'file:'
+        file_url = 'file:///' + absolute_path.replace('\\', '/')
+    else:  # UNIX-like systems
+        file_url = 'file://' + absolute_path
+    _ = gui._simulate_upload_data(file_url)
+
+    assert len(gui.data['simulation_data']) == 1
+    assert 'test_default' in gui.data['simulation_data'].keys()
+    assert gui.data['simulation_data']['test_default']['net'] is None
+    assert type(gui.data['simulation_data']['test_default']['dpls']) is list
```
ntolley, line 631: Could you add a few more checks to make sure all the columns are loaded correctly and have the same length? Namely checking that `times` column (leftmost), `'agg'`, `'L2'`, and `'L5'` have the same once loaded into this data structure.

Perhaps hard code the value for the length they should have like `len(gui.data[...]) == VALUE`

hnn_core/tests/test_gui.py
```diff
@@ -563,3 +565,68 @@ def add_child_decorated(self, to_add):
...
+def test_gui_upload_csv_simulation():
+    """Test if gui handles uploaded csv data"""
+    gui = HNNGUI()
+    _ = gui.compose()
+
+    assert len(gui.viz_manager.data['figs']) == 0
+    assert len(gui.data['simulation_data']) == 0
+
+    # Relative path to the file
+    file_relative_path = "./hnn_core/tests/assets/test_default.csv"
```
ntolley, line 618: possible to add an error raising test when the extension is not `.csv` or `.txt`?

hnn_core/tests/test_gui.py
```diff
@@ -563,3 +565,68 @@ def add_child_decorated(self, to_add):
     # Check if the widgets are relinked, the selected index should be synced
     gui.tab_group_1.selected_index = 0
     assert gui.tab_group_2.selected_index == 0
+
+
+def test_gui_download_simulation():
+    """Test the GUI download simulation pipeline."""
+    gui = HNNGUI()
+    _ = gui.compose()
+    gui.params['N_pyr_x'] = 3
```
ntolley, line 574: Is there any way to set this with the `Network(..., mesh_param=(3, 3)` API? We've worked really hard to remove code that depends on modifying the `.params` object so that we can remove this attribute entirely.

hnn_core/tests/assets/test_default.csv
```diff
@@ -0,0 +1,202 @@
+# times,agg,L2,L5
+0.000000, 0.004901, -0.000004, 0.004905
```
ntolley, line 2: I suppose this is ok since the file is so small, we generally try to avoid having data files in the commit history, see:
https://github.com/jonescompneurolab/hnn-core/blob/0346dbf5d69a190491d83431e6d4209abc2b61ae/hnn_core/tests/test_parallel_backends.py#L199-L208

@jasmainak @gtdang the file size is just 8kb so I don't think there's much concern. Do you want to leave the test as is or move this to a data branch?

gtdang: My preference would to have small tests files included with the dev install/clone. This allows for offline development and also removes the dependency of an external api.

hnn_core/tests/test_gui.py
gtdang: The GUI currently saves the params as an attribute and initializes a new Network object when it needs one to run a function (such as running simulations).  This is so that any connectivity or drive updates though the widgets is used in a new Network object. There isn't a Network object saved to an attribute which that can be accessed with the API call. Does that make sense? 

We could potentially change this with the new json implementation though.

hnn_core/tests/test_gui.py
kmilo9999: I added new asserts verifying the size of the data arrays

hnn_core/tests/test_gui.py
kmilo9999: The logic for uploading csv files is implemented entirely in the GUI layer, and the FileUpload widget already filters files by type. This test programmatically bypasses that filter, but I cant see a scenario where someone would attempt to upload files other than txt or csv.

ntolley (approved):
Great work at @kmilo9999!!

hnn_core/tests/test_gui.py
ntolley: Would this be easy to tack on to that PR? I suppose it might be a whole separate PR to disentangle the params from the GUI code.

PR #754, opened 2024-04-28
[MRG] ENH: `plot_spikes_raster` to always include all cell types

### LINKED ISSUE #658: cell response raster plots should show all neurons in the network model

samadpls:
Closes #658

- Updated `plot_spikes_raster` logic to always include all cell types.

### DIFF
--- a/hnn_core/viz.py
+++ b/hnn_core/viz.py
@@ -534,17 +534,12 @@ def plot_spikes_raster(cell_response, trial_idx=None, ax=None, show=True):
     _validate_type(trial_idx, list, 'trial_idx', 'int, list of int')
 
     # Extract desired trials
-    if len(cell_response._spike_times[0]) > 0:
-        spike_times = np.concatenate(
-            np.array(cell_response._spike_times, dtype=object)[trial_idx])
-        spike_types = np.concatenate(
-            np.array(cell_response._spike_types, dtype=object)[trial_idx])
-        spike_gids = np.concatenate(
-            np.array(cell_response._spike_gids, dtype=object)[trial_idx])
-    else:
-        spike_times = np.array([])
-        spike_types = np.array([])
-        spike_gids = np.array([])
+    spike_times = np.concatenate(
+        np.array(cell_response._spike_times, dtype=object)[trial_idx])
+    spike_types = np.concatenate(
+        np.array(cell_response._spike_types, dtype=object)[trial_idx])
+    spike_gids = np.concatenate(
+        np.array(cell_response._spike_gids, dtype=object)[trial_idx])
 
     cell_types = ['L2_basket', 'L2_pyramidal', 'L5_basket', 'L5_pyramidal']
     cell_type_colors = {'L5_pyramidal': 'r', 'L5_basket': 'b',
@@ -553,7 +548,6 @@ def plot_spikes_raster(cell_response, trial_idx=None, ax=None, show=True):
     if ax is None:
         _, ax = plt.subplots(1, 1, constrained_layout=True)
 
-    ypos = 0
     events = []
     for cell_type in cell_types:
         cell_type_gids = np.unique(spike_gids[spike_types == cell_type])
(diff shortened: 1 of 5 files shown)


jasmainak:
Can you share before/after for a situation where the cell wasn't spiking before?

You can try plotting the spike raster in the [alpha example](https://jonescompneurolab.github.io/hnn-core/stable/auto_examples/workflows/plot_simulate_alpha.html#sphx-glr-auto-examples-workflows-plot-simulate-alpha-py) which is subthreshold activity

samadpls:
Apologies for the delay. Here's the comparison,

| Before | After |
|--------|-------|
| image | Figure_1  |

hnn_core/viz.py
```diff
@@ -514,17 +514,13 @@ def plot_spikes_raster(cell_response, trial_idx=None, ax=None, show=True):
...
     # Extract desired trials
-    if len(cell_response._spike_times[0]) > 0:
-        spike_times = np.concatenate(
-            np.array(cell_response._spike_times, dtype=object)[trial_idx])
-        spike_types = np.concatenate(
-            np.array(cell_response._spike_types, dtype=object)[trial_idx])
-        spike_gids = np.concatenate(
-            np.array(cell_response._spike_gids, dtype=object)[trial_idx])
-    else:
-        spike_times = np.array([])
```
ntolley, line 525: Any reason for not casting this as an `np.array` up here? It seems inefficient to be calling `np.array()` multiple times in a loop

samadpls: Yes, I tried that approach; however, initializing with np.array was not able to produce the plot and list. I've been getting empty results.

jasmainak:
> Here's the comparison,

something is off here ... you shouldn't be getting spiking in this example, no?

samadpls:
You're correct that in the simulation of alpha and beta rhythms in subthreshold conditions, [cells should not be spiking](https://jonescompneurolab.github.io/hnn-tutorials/alpha_and_beta/alpha_and_beta#toc_one).
here is the comparison, 
| Before | After |
|--------|-------|
| image |  Figure_1|

jasmainak:
Better! But does the y-axis go from 0 to 270? That is to say ... how do we make it obvious to the user that there is no spiking or that only some subsets of neurons are spiking?

jasmainak:
perhaps @ntolley or @rythorpe can share examples of raster plots they have seen in publications / other software ...

samadpls:
> Better! But does the y-axis go from 0 to 270? That is to say ... how do we make it obvious to the user that there is no spiking or that only some subsets of neurons are spiking?

Perhaps we can set the y-axis limit to this particular range [0,270].

ntolley:
> Edit: Perhaps we can set the y-axis limit to the range from -total_neurons to 0

This is really all we need to do!

jasmainak:
Let's do that. Can you share how the other example plots look after doing so?

samadpls:
> Let's do that. Can you share how the other example plots look after doing so?

Sure, I just ran the [Gamma Rhythms](https://jonescompneurolab.github.io/hnn-core/stable/auto_examples/workflows/plot_simulate_gamma.html#sphx-glr-auto-examples-workflows-plot-simulate-gamma-py). However, I received a graph with a lot of empty space at the bottom, which I believe was due to setting the Y-limit for all conditions. Therefore, I just moved the position and got the same plot as in the example. Here are the before and after changes.
 | Before | After |
|--------|-------|
| output| Figure_1|

ntolley:
Ah I see! Sorry I think the actual solution is to have the y-axis correspond to the gid of each cell. The tricky part is if we specify only a subset of cells to plot, then the gids used for plotting will need to be updated so that they stop at zero.

Ultimately the idea is that if you compare plots from two different simulations, you want the y-axis to mean the same thing. Right now the y-axis changes depending on the number of cells that spike.

ntolley:
@samadpls when you're back from exams can you try deleting the text in `gui/index.rst`? Everything except "The brand-new lightweight web-based hnn GUI that works with hnn-core."

I think the GUI docs are still technically being built despite removing it from the navbar through `conf.py`

samadpls:
Hey @ntolley , I made the changes and added `gui/index.rst` to `exclude_patterns` in `conf.py` to exclude it from the build. The GUI section is no longer being built.

samadpls:
Hey @ntolley, this is all set for merging. Feel free to do it whenever it's convenient for you!

ntolley (approved):

PR #759, opened 2024-05-06
Debug codecov issues

gtdang:
Debugging GitHub actions for uploading codecov report. Unittest actions have been failing since switching to using the [codecov-action](https://github.com/codecov/codecov-action) v4 in #758 .

Codecov uploads have sporadically been failing due to rate-limiting of 'tokenless' uploads.  It is not actually tokenless, see explanation below.

> If Codecov detects that reports are being uploaded from a fork of an open source repository, Codecov will automatically switch to tokenless uploading...Token-less uploads still use a token to upload to Codecov - just that it's Codecov's Github token instead of the project's own. As such, uploading to Codecov via tokenless (whether it's via forks, or by an older version of the Codecov action) ends up consuming from the shared request pool of a lot of Codecov users, instead of your own. **Therefore, there may be cases when uploading to Codecov via tokenless fails because Codecov is being rate limited.** [Source](https://docs.codecov.com/docs/codecov-uploader#supporting-token-less-uploads-for-forks-of-open-source-repos-using-codecov)

This issue has been documented by others:
https://github.com/codecov/feedback/issues/358
https://github.com/codecov/feedback/issues/301
https://github.com/codecov/feedback/issues/112

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Test with pytest
         shell: bash -el {0}
         run: |
-          python -m pytest .  --cov=hnn_core hnn_core/tests/ --cov-report=xml
+          python -m pytest ./hnn_core/tests/test_io.py --cov=hnn_core --cov-report=xml
       - name:  Upload coverage to Codecov
         uses: codecov/codecov-action@v4
         with:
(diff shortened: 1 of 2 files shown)


gtdang:
One option is changing the setting `fail_ci_if_error = false`. This will not fail the test if the coverage isn't uploaded. There just won't get a coverage report.

gtdang:
Bash implementation doesn't fail. I don't think it generated a new report though. 

Bash uploader is being deprecated, though there has been no indication on when they are fully deprecating it.
https://docs.codecov.com/docs/about-the-codecov-bash-uploader

gtdang:
Closing because this was just used to debug workflows.

PR #761, opened 2024-05-07
Codecov upload revert

### LINKED ISSUE #760: Coverage not uploading to Codecov
Unittest actions have been failing since switching to using the [codecov-action](https://github.com/codecov/codecov-action) v4 in https://github.com/jonescompneurolab/hnn-core/pull/758 .

Codecov uploads have sporadically been failing due to rate-limiting of 'tokenless' uploads. It is not actually tokenless, see explanation below.

> If Codecov detects that reports are being uploaded from a fork of an open source repository, Codecov will automatically switch to tokenless uploading...Token-less uploads still use a token to upload to Codecov - just that it's Codecov's Github token instead of the project's own. As such, uploading to Codecov via tokenless (whether it's via forks, or by an older version of the Codecov action) ends up consuming from the shared request pool of a lot of Codecov users, instead of your own. **Therefore, there may be cases when uploading to Codecov via tokenless fails because Codecov is being rate limited**. [Source](https://docs.codecov.com/docs/codecov-uploader#supporting-token-less-uploads-for-forks-of-open-source-repos-using-codecov)

This issue has been documented by others:
https://github.com/codecov/feedback/issues/358
https://github.com/codecov/feedback/issues/301
https://github.com/codecov/feedback/issues/112

I've looked into this problem with #759. I think we will revert back to the bash uploader for now. I'm not sure if it's working properly but at least it doesn't fail the workflow. The bash uploader will be deprecated eventually but they will hopefully update their upload action by then.

gtdang:
Reverting back to the bash uploader until the codecov upload action is corrected. See details in #760 and #759.

closes #760

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -48,10 +48,6 @@ jobs:
         run: |
           python -m pytest .  --cov=hnn_core hnn_core/tests/ --cov-report=xml
       - name:  Upload coverage to Codecov
-        uses: codecov/codecov-action@v4
-        with:
-            directory: ./
-            fail_ci_if_error: true
-            files: ./coverage.xml
-            flags: unittests
-            token: ${{ secrets.CODECOV_TOKEN }}
+        shell: bash -el {0}
+        run: |
+          bash <(curl -s https://codecov.io/bash) -f ./coverage.xml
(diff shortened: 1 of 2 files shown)


ntolley (approved):

ntolley:
Thanks @gtdang ! 

@jasmainak can you see #281 for updating the codecov secrets?

PR #769, opened 2024-05-15
[MRG] Make all plots with time end at tstop 

katduecker:
This pull request supersedes PR[#752](https://github.com/jonescompneurolab/hnn-core/pull/752)

The x/time-axis of all plots now ends at tstop. Deprecation cycle for tmin and tmax added to plot_laminar_lfp, plot_dipole. tmin and tmax are still required for plot_psd and plot_tfr_morlet.

### DIFF
--- a/hnn_core/extracellular.py
+++ b/hnn_core/extracellular.py
@@ -457,11 +457,6 @@ class ExtracellularArray:
             Trial number(s) to plot
         contact_no : int | list of int | slice
             Electrode contact number(s) to plot
-        tmin : float | None
-            Start time of plot in milliseconds. If None, plot entire
-            simulation.
-        tmax : float | None
-            End time of plot in milliseconds. If None, plot entire simulation.
         ax : instance of matplotlib figure | None
             The matplotlib axis
         decim : int | list of int | None (default)
--- a/hnn_core/tests/test_extracellular.py
+++ b/hnn_core/tests/test_extracellular.py
@@ -258,3 +258,21 @@ def test_rec_array_calculation():
         assert_allclose(net.rec_arrays['arr1']._data[trial_idx][1],
                         net.rec_arrays['arr2']._data[trial_idx][1],
                         rtol=1e-3, atol=1e-3)
+
+def test_extracellular_viz():
+    """Test if deprecation warning is raised in plot_laminar_lfp."""
+    hnn_core_root = op.dirname(hnn_core.__file__)
+    params_fname = op.join(hnn_core_root, 'param', 'default.json')
+    params = read_params(params_fname)
+    params.update({'t_evprox_1': 7, 't_evdist_1': 17})
+    net = jones_2009_model(params, mesh_shape=(3, 3),
+                           add_drives_from_params=True)
+
+    # one electrode inside, one above the active elements of the network,
+    # and two more to allow calculation of CSD (2nd spatial derivative)
+    electrode_pos = [(1, 2, 1000), (2, 3, 3000), (3, 4, 5000), (4, 5, 7000)]
+    net.add_electrode_array('arr1', electrode_pos)
+    _ = simulate_dipole(net, tstop=5, n_trials=1)
+
+    with pytest.deprecated_call():
+        net.rec_arrays['arr1'].plot_lfp(show=False, tmin=10, tmax=100)
--- a/hnn_core/tests/test_extracellular.py
+++ b/hnn_core/tests/test_extracellular.py
@@ -259,6 +259,7 @@ def test_rec_array_calculation():
                         net.rec_arrays['arr2']._data[trial_idx][1],
                         rtol=1e-3, atol=1e-3)
 
+
 def test_extracellular_viz():
     """Test if deprecation warning is raised in plot_laminar_lfp."""
     hnn_core_root = op.dirname(hnn_core.__file__)
--- a/hnn_core/viz.py
+++ b/hnn_core/viz.py
@@ -180,7 +180,7 @@ def plot_laminar_lfp(times, data, contact_labels, tmin=None, tmax=None,
             col = color
         ax.plot(plot_times, plot_data + trace_offsets[contact_no],
                 label=f'C{contact_no}', color=col)
-        
+
         # To be removed after deprecation cycle
         if tmin is not None or tmax is not None:
             ax.set_xlim(left=tmin, right=tmax)
@@ -291,7 +291,7 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
 
             if layer in dpl_trial.data.keys():
 
-                 # extract scaled data and times
+                # extract scaled data and times
                 data = dpl_trial.data[layer]
                 times = dpl_trial.times
 
@@ -304,7 +304,7 @@ def plot_dipole(dpl, tmin=None, tmax=None, ax=None, layer='agg', decim=None,
                     alpha = 0.5 if average else 1.
                     ax.plot(times, data, color=_lighten_color(color, 0.5),
                             alpha=alpha, lw=1.)
-            
+
             # To be removed after deprecation cycle
             if tmin is not None or tmax is not None:
                 if tmin is not None or tmax is not None:
@@ -491,7 +491,7 @@ def plot_spikes_hist(cell_response, trial_idx=None, ax=None, spike_types=None,
         hist_color = spike_color[spike_label]
         ax.hist(plot_data, bins,
                 label=spike_label, color=hist_color, **kwargs_hist)
-        
+
     if len(cell_response.times) > 0:
         ax.set_xlim(left=0, right=cell_response.times[-1])
     else:
@@ -579,7 +579,6 @@ def plot_spikes_raster(cell_response, trial_idx=None, ax=None, show=True):
         ax.set_xlim(left=0, right=cell_response.times[-1])
     else:
         ax.set_xlim(left=0)
-        
     ax.set_xlim(left=0)
 
     plt_show(show)
--- a/hnn_core/viz.py
+++ b/hnn_core/viz.py
@@ -7,6 +7,7 @@
 import numpy as np
 from itertools import cycle
 import colorsys
+import warnings
 
 from .externals.mne import _validate_type
 
@@ -21,7 +22,7 @@ def _lighten_color(color, amount=0.5):
     return colorsys.hls_to_rgb(c[0], 1 - amount * (1 - c[1]), c[2])
 
 
-def _get_plot_data_trange(times, data, tmin, tmax):
+def _get_plot_data_trange(times, data, tmin=None, tmax=None):
     """Get slices of times and data based on tmin and tmax"""
     if isinstance(times, list):
         times = np.array(times)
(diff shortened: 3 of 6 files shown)


hnn_core/viz.py
```diff
@@ -554,7 +573,11 @@ def plot_spikes_raster(cell_response, trial_idx=None, ax=None, show=True):
     ax.set_facecolor('k')
     ax.set_xlabel('Time (ms)')
     ax.get_yaxis().set_visible(False)
-    ax.set_xlim(left=0)
+
+    if len(cell_response.times) > 0:
```
ntolley, line 577: this is great @katduecker!

does `plot_spikes_histogram()` need the same logic?

ntolley:
Just some flake8 errors and one comment to address but otherwise this is about ready to merge! Feel free to add [MRG] to the title when you're ready for a final code review

hnn_core/extracellular.py
```diff
@@ -457,11 +457,12 @@ def plot_lfp(self, *, trial_no=None, contact_no=None, tmin=None, tmax=None,
...
         contact_no : int | list of int | slice
             Electrode contact number(s) to plot
-        tmin : float | None
-            Start time of plot in milliseconds. If None, plot entire
-            simulation.
-        tmax : float | None
-            End time of plot in milliseconds. If None, plot entire simulation.
+        tmin : float | None [deprecated]
+            Start time of plot in milliseconds.
+            If None, plot entire simulation.
```
jasmainak, line 462: This makes it look like `None` is deprecated. I'd suggest removing the docstring of these arguments altogether since the entire argument will be removed. Also, if we want to move towards deprecating this argument, we don't want to advertise it to the users.

hnn_core/viz.py
```diff
@@ -182,7 +185,13 @@ def plot_laminar_lfp(times, data, contact_labels, tmin=None, tmax=None,
             col = color
         ax.plot(plot_times, plot_data + trace_offsets[contact_no],
                 label=f'C{contact_no}', color=col)
-
+        # To be removed after deprecation cycle
+        if tmin is not None or tmax is not None:
+            ax.set_xlim(left=tmin, right=tmax)
+            warnings.warn('tmin and tmax are deprecated and will be removed in future releases of hnn-core.'
+                          'By default, dipoles and laminar LFPs are now plotted from 0 to tstop.', DeprecationWarning)
```
jasmainak, line 192:
```suggestion
                          'Please use matplotlib directly to set tmin and tmax', DeprecationWarning)
```

nice to tell the user what they should do instead

jasmainak:
@katduecker can you add a tiny test to check that the deprecation warning is raised? Also, do any of our old tests or examples use `tmin` or `tmax` ? We should update them if they do

hnn_core/viz.py
```diff
@@ -168,7 +170,8 @@ def plot_laminar_lfp(times, data, contact_labels, tmin=None, tmax=None,
         trace_offsets = np.arange(n_offsets)[:, np.newaxis] * voltage_offset
 
     for contact_no, trace in enumerate(np.atleast_2d(data)):
-        plot_data, plot_times = _get_plot_data_trange(times, trace, tmin, tmax)
+        plot_data = trace
+        plot_times = times
```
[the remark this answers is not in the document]
katduecker: Not sure if we mean the same thing here. The result will be the same as before: the data will be plotted from tmin to tmax if they are not none. Before, plot_data and plot_times ranged from tmin to tmax, now plot_data and plot_times range from 0 to tstop, but the x-axis will be adjusted to range from tmin to tmax if not None. I have tested this locally and it does not break the scripts. My logic was that this way, we could easily remove the xlim setting after the deprecation cycle (it's commented in the code). 

I also saw that because xlim wasn't set in the function, there was some whitespace on the left and/or right side of the plot when tmin and tmax are set to a value that is not None and sharex=True, as in this example https://jonescompneurolab.github.io/hnn-core/dev/auto_examples/howto/plot_record_extracellular_potentials.html#sphx-glr-auto-examples-howto-plot-record-extracellular-potentials-py

There is no whitespace for this version anymore, see:

tstop_fix


For psd and csd tmin and tmax are still used in _get_data_range following your vote in the first PR. I understand that this will save the user time and compute when calculating CSD and PSD. Hope this is okay. Will address the rest tomorrow with a fresh mind!

hnn_core/viz.py
katduecker: haven't touched plot_spikes_hist(), but it seems that xlim is not set explicitly in this function and tmin and tmax are not required in that function.

katduecker:
> @katduecker can you add a tiny test to check that the deprecation warning is raised? Also, do any of our old tests or examples use `tmin` or `tmax` ? We should update them if they do

@jasmainak thanks for bringing this up! Just trying this out locally it seems that a DeprecationWarning is suppressed in Jupyter Notebook, which I just found out might be the default in many python environments.

Options I found to fix this were to either use warnings.simplefilter('always', DeprecationWarning) at the beginning of the script (probably not advisable?) or use UserWarning instead of DeprecationWarning.
There may be a DeprecationWarning used in other functions, so I add that fix there, too.
What do you think?

jasmainak:
@katduecker you should become familiar with `pytest` framework. See: 

https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#ensuring-code-triggers-a-deprecation-warning

the rest of the page is informative too. You'll see different ways in which to capture/check warnings.

Jupyter notebook is not an ideal environment for coding ... it's good for data exploration but it is not ideal for coding. I'd recommend using VS Code (or other editor) + terminal. The way to run your tests is to do:

```py
$ pytest .
```

then if you want to do post-mortem debugging, you can do:


```py
$ pytest . --pdb
```

etc.

katduecker:
@jasmainak sorry, it seems that I didn't express myself clearly. My question was if I should use UserWarning instead of DeprecationWarning, in case users are working in a python environment that suppresses DeprecationWarnings.

jasmainak:
oh I see, interesting question! I didn't know that. A bit of digging revealed that [FutureWarnings are now recommended over DeprecationWarning](https://github.com/mne-tools/mne-python/issues/11111). One is for developers, the other for users ... we might want to make this change repo-wide. Can be your next PR!

katduecker:
Okay, I'll add the pytest for the deprecation warning for now, and then we can decide where to go from there. Should the deprecation test be added to test_viz.py?

jasmainak:
yep, add it to test_viz.py

hnn_core/viz.py
katduecker: @ntolley this turned out to be more complicated than anticipated. For some reason, I cannot set xlim outside plot_spikes_histogram() (can show you in person). Not sure why that is because it works if I change xlim inside the function, after the histogram loop. It seems that in this case, we might actually want to add tmin and tmax to plot_spikes_histogram() ... I wonder if it would be more user-friendly in this case to just keep tmin and tmax for all plotting functions?

hnn_core/tests/test_extracellular.py
```diff
@@ -258,3 +256,22 @@ def test_rec_array_calculation():
...
+                           add_drives_from_params=True)
+
+    # one electrode inside, one above the active elements of the network,
+    # and two more to allow calculation of CSD (2nd spatial derivative)
+    electrode_pos = [(1, 2, 1000), (2, 3, 3000), (3, 4, 5000), (4, 5, 7000)]
+    net.add_electrode_array('arr1', electrode_pos)
+    _ = simulate_dipole(net, tstop=5, n_trials=1)
+
+    with pytest.deprecated_call():
+        net.rec_arrays['arr1'].plot_lfp(show=False, tmin=10, tmax=100)
```
jasmainak, line 277: is it possible to update an existing test rather than writing a new test altogether?

hnn_core/tests/test_extracellular.py
```diff
@@ -210,15 +210,13 @@ def test_rec_array_calculation():
...
-                   't_evdist_1': 17})
+    params.update({'t_evprox_1': 7, 't_evdist_1': 17})
     net = jones_2009_model(params, mesh_shape=(3, 3),
                            add_drives_from_params=True)
 
     # one electrode inside, one above the active elements of the network,
     # and two more to allow calculation of CSD (2nd spatial derivative)
-    electrode_pos = [(1, 2, 1000), (2, 3, 3000), (3, 4, 5000),
-                     (4, 5, 7000)]
+    electrode_pos = [(1, 2, 1000), (2, 3, 3000), (3, 4, 5000), (4, 5, 7000)]
```
jasmainak, line 219: as a general rule, one should keep lines not relevant to the PR untouched. Formatting changes can be delegated to a PR of its own. The smaller your diff, the less comments, and easier it is for us to merge.

jasmainak:
@katduecker one last comment and then you can update `whats_new.rst` !

hnn_core/tests/test_extracellular.py
katduecker: There is no test for the visualization of the LFP (in test_viz or test_extracellular) which is why I did this here. Could write a test_lfp_visualization() for a next PR?

hnn_core/tests/test_extracellular.py
katduecker: Sorry, I think this happened because I had a flake8 error for this line (empty whitespace)

gtdang: Yes, I've noticed that there are many flake8 formatting errors in the codebase that were probably merged before flake8 was implemented as a gh action check,  this is because the gh action only checks edited files in the current PR, not the rest of the codebase. So these error are only flagged when a contributor edits a file with one of these legacy errors. I can make separate PR that sweeps the codebase for these legacy formatting errors.

katduecker:
@jasmainak @ntolley hey, I'm not sure why the tests were canceled because they run locally? any way tor resolve this?

jasmainak:
you have a rebase issue ... I see 126 commits!

katduecker:
It was all working in commit 59f1620. I then struggled to rebase because it seems that a few lines were added to whats_new.rst just before I did. I was struggling a bit to get it right (that was 3 commits), but the rebase was eventually successful for 3c422b5 and whats_new.rst looks sensible now.

ntolley:
@katduecker the tests failing is a separate issue we're working on so no worries!

Also I think there's some confusion on what the rebase issue is. @jasmainak is referring to the fact that a large number of commits from other PR's are included here. If you look at the "Files Changed" the majority of these changes are not ones you made (e.g. mine and George's commits from another PR that was merged into the `master` branch);

image


I'm not sure which commands were run, but this typically happens when `git rebase` is used incorrectly.

ntolley:
To avoid this in the future it's best to verify that the rebase (and merge conflict resolution) was successful by running `git log` locally, and verifying that 1) the most recent commits are exclusively ones from your PR, and 2) the commits before your PR match exactly the ones from the master branch. If something went wrong, then you can pull the last version from github and try again.

Now that these changes are pushed to github @jasmainak is the best option to reset and cherry pick these commits or something?

jasmainak:
honestly given the rebase issues ... I feel the easiest is to copy the edits into a new branch and make commits on that branch. If you'd like to give credit to the previous contributors, you can use [co-authored commits](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors). Then rename the branch to `tstop_plot` and force push.

ntolley:
Sorry @katduecker totally understand that rebase issues are frustrating to understand/fix

The skills for resolving them are super useful for code development, and you'll get a lot of street cred as a git superuser :smile:

katduecker:
Thanks for clarifying, both! I'll do what Mainak said and copy the edits into a new branch. And thanks for clarifying what the rebase issue was! Do I need to wait until the problem with the tests failing is resolved?

ntolley:
No need to wait on the tests being fixed! It's an issue specific to the github servers that run the tests, but only the ubuntu tests are impacted

Your local tests (and the unit tests for Mac and Windows) are still passing so you're all good from a unit testing stand point

katduecker:
Okay thanks! And just to confirm, I delete my local tstop_plot branch, create a new local branch, implement changes, rename new local tstop branch to tstop_plot and run git push origin tstop_plot -f ?

PR #792, opened 2024-06-12
[MRG] Record calcium concentration

katduecker:
Recording calcium concentration in all sections of the L5 pyramidal neurons

### DIFF
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -348,6 +348,11 @@ class Cell:
         by synapse type (keys can be soma_gabaa, soma_gabab etc.).
         Must be enabled by running simulate_dipole(net, record_isec=True)
         or simulate_dipole(net, record_isoma=True)
+    ca : dict
+        Contains recording of section speicifc calcium concentration. 
+        Must be enabled by running simulate_dipole(...) with record_ca
+        set to 'all' to record form all sections, or 'soma' to record from
+        the soma only.
     tonic_biases : list of h.IClamp
         The current clamps inserted at each section of the cell
         for tonic biasing inputs.
@@ -391,6 +396,7 @@ class Cell:
         self.dipole_pp = list()
         self.vsec = dict()
         self.isec = dict()
+        self.ca = dict()
         # insert iclamp
         self.list_IClamp = list()
         self._gid = None
@@ -468,6 +474,7 @@ class Cell:
         cell_data['dipole_pp'] = self.dipole_pp
         cell_data['vsec'] = self.vsec
         cell_data['isec'] = self.isec
+        cell_data['ca'] = self.ca
         cell_data['tonic_biases'] = self.tonic_biases
         return cell_data
 
@@ -743,7 +750,7 @@ class Cell:
         stim.amp = amplitude
         self.tonic_biases.append(stim)
 
-    def record(self, record_vsec=False, record_isec=False):
+    def record(self, record_vsec=False, record_isec=False, record_ca=False):
         """ Record current and voltage from all sections
 
         Parameters
@@ -754,6 +761,9 @@ class Cell:
         record_isec : 'all' | 'soma' | False
             Option to record voltages from all sections ('all'), or just
             the soma ('soma'). Default: False.
+        record_ca : 'all' | 'soma' | False
+            Option to record calcium concentration from all sections ('all'), 
+            or just the soma ('soma'). Default: False.
         """
 
         section_names = list(self.sections.keys())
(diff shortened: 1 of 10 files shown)


hnn_core/cell.py
```diff
@@ -783,9 +793,22 @@ def record(self, record_vsec=False, record_isec=False):
...
+            self.ca = dict.fromkeys(['soma'])
+        elif record_ca == 'all':
+            self.ca = dict.fromkeys(section_names)
+
+        if record_ca:     
+            for sec_name in self.ca:
+                if 'ca' in self.sections[sec_name].mechs:
+                    self.ca[sec_name] = h.Vector()
+                    self.ca[sec_name].record(
+                        self._nrn_sections[sec_name](0.5)._ref_cai) # there may be an issue here with referencing mod file
```
ntolley, line 810: keep or remove?

hnn_core/cell.py
```diff
@@ -348,6 +348,11 @@ class Cell:
         by synapse type (keys can be soma_gabaa, soma_gabab etc.).
         Must be enabled by running simulate_dipole(net, record_isec=True)
         or simulate_dipole(net, record_isoma=True)
+    ca : dict
+        Contains recording of section speicifc calcium concentration. 
+        Must be enabled by running simulate_dipole(...) with record_ca
```
jasmainak, line 353:
```suggestion
        Must be enabled by running simulate_dipole(net, record_ca=True)
```

hnn_core/tests/test_dipole.py
```diff
@@ -214,30 +217,41 @@ def test_cell_response_backends(run_hnn_core_fixture):
...
     assert len(joblib_net.cell_response.vsec[
         trial_idx][gid]['apical_1']) == n_times
     assert len(joblib_net.cell_response.isec[
                trial_idx][gid]['soma']['soma_gabaa']) == n_times
-
     assert len(mpi_net.cell_response.vsec) == n_trials
     assert len(mpi_net.cell_response.isec) == n_trials
     assert len(mpi_net.cell_response.vsec[trial_idx][gid]) == 8  # num sec
-    assert len(mpi_net.cell_response.isec[trial_idx][gid]) == 1
+    assert len(mpi_net.cell_response.isec[trial_idx][gid]) == 1
```
jasmainak, line 237: this must be a pep8 issue

jasmainak:
@ntolley monthly issue metrics fails !

hnn_core/cell.py
katduecker: It's not True, it's either 'all' or 'soma'

ntolley:
> @ntolley monthly issue metrics fails !

Fixed!

hnn_core/cell.py
ntolley: I think @jasmainak is copying that from the docstring above for `record_isoma`, which is out of date.

But yes the docstring should read `Must be enabled by running simulate_dipole(..., record_ca='all')`

hnn_core/dipole.py
```diff
@@ -37,6 +37,9 @@ def simulate_dipole(net, tstop, dt=0.025, n_trials=None, record_vsec=False,
     record_isec : 'all' | 'soma' | False
         Option to record voltages from all sections ('all'), or just
         the soma ('soma'). Default: False.
+    record_ca : 'all' | 'soma' | False
```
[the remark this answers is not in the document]
ntolley: Can you also replace the incorrect record isoma/vsoma docstrings in `cell.py` with the isec/vsec docstrings here?

hnn_core/cell.py
```diff
@@ -783,9 +793,22 @@ def record(self, record_vsec=False, record_isec=False):
...
+        
+        # calcium concentration
+        if record_ca == 'soma':
+            self.ca = dict.fromkeys(['soma'])
+        elif record_ca == 'all':
+            self.ca = dict.fromkeys(section_names)
+
+        if record_ca:     
+            for sec_name in self.ca:
+                if 'ca' in self.sections[sec_name].mechs:
```
ntolley, line 807:
```suggestion
                if hasattr(self._nrn_sections[sec_name](0.5), '_ref_cai)::
```
Your suggestion! Just making sure we don't forget

hnn_core/cell.py
jasmainak: I just wanted to say that use the function signature rather than using words to explain it.

jasmainak (approved):
PR looks clean to me. I leave it to @ntolley to check and merge :) Don't forget to update whats_new !

ntolley:
I just plotted the somatic [Ca] alongside the spike raster. There's a significant delay (~ 20ms) between the first spikes and the influx of calcium. While it's definitely expected that this current is slower, I figured that this would manifest as slow rise/decay times, and not a complete lack of current when the cell spikes

```py
from hnn_core import simulate_dipole, jones_2009_model
import numpy as np

net = jones_2009_model()
net.add_evoked_drive('evprox', mu=20.0, sigma=2.0, numspikes=1,
                     location='proximal', weights_ampa={'L5_pyramidal': 0.1})

plt.figure()
fig, ax = plt.subplots(2, 1, figsize=(5, 4), sharex=True, tight_layout=True)
plot_data = np.array([net.cell_response.ca[0][gid]['soma'] for gid in net.gid_ranges['L5_pyramidal']]).T
ax[0].plot(plot_data)
ax[0].set_title('Somatic Ca$^{2+}$')
ax[1].set_title('Spike Raster')
net.cell_response.plot_spikes_raster(ax=ax[1])
```

image

katduecker:
Thanks both! I haven't managed to look at the rest of this yet, as I have been busy with the NEURON workshop. One thing I learned today at the workshop is that calcium concentration should be at the order a few hundred nM, but our calcium concentration is at the order of 0.1 mM (mod file of ca shows that cai is in mM). This is 1000 times too high... Am I missing something obvious with a conversion of the units?

jasmainak:
We use reduced neurons acc to Bush and Sejnowski 93, so do we actually
expect physical units?

katduecker:
I asked Stephanie about this and she said that it may be possible that the conductance had to be scaled up to get the desired dynamics. To confirm this, I will read out the concentration for sodium and potassium as well. I remember @ntolley saying that this may not be a useful functionality in general, but if it allows sanity checks of the scale of the conductance, couldn't it be helpful for users to have access to this? I'd be happy to write a tutorial on this.

Issue #780, opened 2024-05-29
CI: GH Linux runner stalling on pytest


gtdang:
Our GH actions linux runners for pytest have stopped worked as of sometime last Friday/Saturday.  The runners stall for about 6hrs before they are canceled. It's happening during test_dipole and likely with `test_cell_response_backends` where it's using MPI, though that has yet to be confirmed.

PR #794, opened 2024-06-12
[MAINT] Remove conda from CI Linux  Workflows 

kmilo9999:
The Linux build in the CI is failing because conda is installing a version of neuron without MPI support. Removing conda solves this issue for the linux CI workflow.

Solves #780


.github/workflows/unit_tests.yml
```diff
@@ -12,41 +12,51 @@ jobs:
...
             sudo apt-get update
             sudo apt-get install libopenmpi-dev openmpi-bin
-          fi;
+
+      - name: Install MacOs dependencies
+        if: contains(matrix.os, 'macos')
+        shell: bash -el {0}
+        run: |
+              brew install open-mpi
+              echo "MPI_LIB_NRN_PATH=/opt/homebrew/lib/libmpi.dylib" >> $GITHUB_ENV
```
jasmainak, line 38: this is not what we recommend for the users though ...

kmilo9999: I'm open to any suggestions

jasmainak: Conda does not work? It seems to work on my mac and 2 other macs we tried on today ...

kmilo9999: The Linux build in the CI is failing because conda is installing a version of neuron without MPI support. Removing conda solves this issue, but it affects the mac-os build. This PR helps giving support to both systems without conda.

jasmainak: have you read the parallel backend guide:

https://jonescompneurolab.github.io/hnn-core/stable/parallel.html 

? You should use conda for mac but pip for linux

.github/workflows/unit_tests.yml
```diff
@@ -12,41 +12,51 @@ jobs:
...
+        python-version: [3.8,3.9]
 
     steps:
       - uses: actions/checkout@v4
-      - uses: conda-incubator/setup-miniconda@v3
-        with:
-          activate-environment: test
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        id: cp310
```
jasmainak, line 22: what is cp310?

.github/workflows/unit_tests.yml
kmilo9999: I split the os build and test on different workflows

.github/workflows/unit_tests.yml
kmilo9999: It's an id to refer the step in further steps. We dont need this anymore, I removed it.

.github/workflows/linux_unit_test.yml
```diff
@@ -0,0 +1,54 @@
...
+    branches: ['**']
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: [3.8,3.9]
```
gtdang, line 16: Support for python 3.8 will end by the end of this year. Should we consider also testing more recent' versions of python in the matrix? At some point do we plan to stop testing for older versions? How do we currently communicate with users which python versions are actually supported by hnn? 

@jasmainak @rythorpe @ntolley 

https://devguide.python.org/versions/

ntolley: Perhaps let's add up to `3.11` until we figure out what's going on with `3.12`?

Also in terms of communication, perhaps we can explicitly state in the install instructions which versions of python we have listed in the unit tests here. Something like "HNN-core has been tested on python versions 3.8 - 3.11. Other versions may work but we do not explicitly test against them".

ntolley (approved):
This looks great to me! Perhaps we can try adding 3.10 and 3.11 to the list and see if the unit tests still pass?

ntolley:
Looks like `3.11` is failing with the same error as `3.12` mentioned in #796 :face_with_diagonal_mouth:

kmilo9999:
> Looks like `3.11` is failing with the same error as `3.12` mentioned in #796 🫤

I think they fail on `hnn_core/tests/test_parallel_backends.py:test_mpi_failure` 
It's failing in 3.9, 3.10 and 3.11 on all runners.

.github/workflows/linux_unit_test.yml
gtdang: @hollandjg Do you recall how [autora](https://pypi.org/project/autora/) does this? I thought there was a badge with tested python versions at one point, I don't see it anymore. Do they have a methodology for supporting versions?

kmilo9999:
@gtdang @ntolley @jasmainak 
It appears that [NumPy 2.0.0](https://numpy.org/devdocs/release/2.0.0-notes.html) was released last week, and they have removed the deprecated function `np.in1d`. This function is used in multiple parts of our project
image
What should be the next steps? 
It doesnt fail in` python 3.8` because it uses `Numpy 1.94`

kmilo9999:
@gtdang @ntolley It seems 3.11 and 3.12 fail at the same point, but I can take care of it on another PR. 
This one works on 3.8, 3.9 and 3.10. Please review it.

ntolley (approved):
This is great @kmilo9999!! Thanks a bunch for taking care of this. Definitely agree to focus on 3.11 and 3.12 in a separate PR.

jasmainak:
Btw, in mne-python etc., we make our CIs fail if there are deprecation warnings ... it catches these issues early.

jasmainak:
Looks good. What is your thought on two separate runner scripts vs a single one? A common code path (with if-else in appropriate places) would reduce likelihood of bugs ...

kmilo9999:
@gtdang @ntolley @jasmainak I removed the `linux_unit_tests.yml `, and I put linux and mac workflows in the same file with conditionals around the required steps for each OS.
Please review.

jasmainak:
Beautiful, thanks @kmilo9999 !!

PR #807, opened 2024-07-05
[MRG] read_network_configuration not clearing drive connections

gtdang:
Reading network configurations from hierarchical json with `read_drives=False` was only clearing the external_drives and not the drive connectivities in the network connectivity information. This fix corrects that issue.

### DIFF
--- a/hnn_core/hnn_io.py
+++ b/hnn_core/hnn_io.py
@@ -167,10 +167,8 @@ def _set_from_cell_specific(drive_data):
     return drive_data['n_drive_cells']
 
 
-def _read_external_drive(net, drive_data, read_drives, read_output):
+def _read_external_drive(net, drive_data, read_output):
     """Adds drives encoded in json data to a Network"""
-    if not read_drives:
-        return None
 
     if (drive_data['type'] == 'evoked') or (drive_data['type'] == 'gaussian'):
         # Skipped n_drive_cells here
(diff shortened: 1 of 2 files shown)


hnn_core/hnn_io.py
```diff
@@ -439,7 +437,7 @@ def read_network_configuration(fname, read_drives=True):
                                         net_data['external_drives'])
     for key in external_drive_data.keys():
         _read_external_drive(net, external_drive_data[key],
-                             read_output=False, read_drives=read_drives)
+                             read_output=False)
```
ntolley, line 440: Making sure I understand, the idea is that `net.clear_drives()` updates both `net.connectivity` and `net.external_drives`? Whereas this previously only updated (rather omitted updating) the external drive

ntolley (approved):
All good on my end! let me know when it's ready to merge

hnn_core/hnn_io.py
gtdang: Yep. Previously it was built into `read_external_drive` to return None if when supplied `read_drives=false` and we weren't checking and clearing the connectivity. This change just builds everything and uses the Network api's `clear_drives` to remove the drive information (both from `net.external_drives` and `net.connectivity`).

gtdang:
@ntolley Thanks! It's good to merge!

PR #821, opened 2024-07-10
[MRG] Remove nulled drives during convert to hierarchical json function

gtdang:
Updated the hierarchical json conversion script to prune nulled drives. Legacy param files had placeholders for several drives. If unused, values out of typical ranges were specified to null those drives. Typically this occurs when a start time is set greater than the stop time or if weights are set it negatives or 0s. The hierarchical json format will not include drives with bad parameters. 

Summary of Changes: 
* Added logic to remove drives that did not have at least 1 positive weight or had invalid _tstop_ and _tstart_ values. 
* Removed _model_template_ argument from `convert_to_json` and associated tests - After discussion that we are using this function internally to convert tutorial files. And they will all be the jones2009 model. 

**Questions**
1. Are the filters for nulled drives correct/complete? 
    * Check for start < stop 
    * Checks nmda and amps weights for at least one >0 value

### DIFF
--- a/hnn_core/params.py
+++ b/hnn_core/params.py
@@ -735,7 +735,6 @@ def remove_nulled_drives(net):
 
 def convert_to_json(params_fname,
                     out_fname,
-                    model_template='jones_2009_model',
                     include_drives=True,
                     overwrite=True):
     """Converts legacy json or param format to hierarchical json format
@@ -746,11 +745,6 @@ def convert_to_json(params_fname,
         Path to file
     out_fname: str
         Path to output
-    model_template: str or None, default:' jones_2009_model'
-         Options: ['jones_2009_model', 'law_2021_model', 'calcium_model', None]
-         Neocortical network model to use. Models are defined in
-         network_models. If None, the base Network object with no defined
-         network connectivity will be used.
     include_drives: bool, default=True
         Include drives from params file
     overwrite: bool, default=True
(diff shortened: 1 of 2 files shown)


hnn_core/params.py
```diff
@@ -662,9 +662,79 @@ def compare_dictionaries(d1, d2):
...
+    ranges. This function removes drives on the following conditions:
+        1. Start time is larger than stop time
+        2. All weights are non-positive
+
+    Parameters
+    ----------
+    net : Network object
+
+    Returns
+    -------
```
ntolley, line 696:
```suggestion
    -------
    net : Network object
```

hnn_core/tests/test_params.py
```diff
@@ -79,6 +78,36 @@ def test_base_params():
...
+    drives_removed = ['bursty1', 'bursty2', 'extgauss', 'extpois']
+    assert all([drive not in net_removed.external_drives.keys()
+                for drive in drives_removed])
+
+    # Connections were removed
+    conn_src_types = set([conn['src_type']
+                          for conn in net_removed.connectivity])
+    assert all([drive not in conn_src_types for drive in drives_removed])
+
+    # gid ranges were updated
```
ntolley, line 106: You can also add a check for the `pos_dict`

hnn_core/params.py
```diff
@@ -662,9 +662,79 @@ def compare_dictionaries(d1, d2):
...
+        return True
+    else:
+        return False
+
+
+def _get_n_cells(net):
+    """ Get number of neocortical cells  """
+    cell_names = net.cell_types.keys()
+    n_cells = max([net.gid_ranges[cell_name].stop for cell_name in cell_names])
+    return n_cells
```
ntolley, line 679: There's technically and `_n_cells` attribute that I believe is updated separately for `_n_gids`.

Do you think it's better to use the attribute or calculate the number of cells directly?

gtdang: Ah thanks for pointing this out. The function isn't used anymore so I deleted it.

hnn_core/tests/test_params.py
gtdang: Great, just added that assert.

gtdang:
@ntolley
Can you check to make sure the logic I'm using to identify nulled drives is correct? There's a check on the start/stop time. And for the weights, it's checking for at least one positive ampa and nmda weight. Could there be other weight conditions we should exclude or include?

Issue #824, opened 2024-07-16
GUI - Load data first throws exception


kmilo9999:
It seems the dropdown list of simulated data to download it's including loaded data

Steps to replicate:
- Load data
- Run simulation
- Error printed in output console

ntolley:
I chatted with @rythorpe and the original GUI did not allow exporting of data that was loaded in

kmilo9999:
> I chatted with @rythorpe and the original GUI did not allow exporting of data that was loaded in

Thank you Nick. I'll proceed to fix this.

PR #825, opened 2024-07-16
[MRG]: Fixed simulations dropdown list adding loaded data

kmilo9999:
The simulations dropdown list was adding loaded data, and hnn-gui cant export non-simulated data.

solves #824

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -1727,7 +1727,7 @@ def run_button_clicked(widget_simulation_name, log_out, drive_widgets,
                 'finished']
 
             sim_names = [sim_name for sim_name in simulation_data
-                               if simulation_data[sim_name]['net'] is not None]
+                         if simulation_data[sim_name]['net'] is not None]
 
             simulations_list_widget.options = sim_names
             simulations_list_widget.value = sim_names[0]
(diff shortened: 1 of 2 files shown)


hnn_core/gui/gui.py
```diff
@@ -1726,10 +1726,11 @@ def run_button_clicked(widget_simulation_name, log_out, drive_widgets,
             simulation_status_bar.value = simulation_status_contents[
                 'finished']
 
-            sim_names = [sim_name for sim_name
-                         in simulation_data]
-            simulations_list_widget.options = sim_names
-            simulations_list_widget.value = sim_names[0]
+            run_simulations = [sim_name for sim_name in simulation_data
+                               if simulation_data[sim_name]['net'] is not None]
```
ntolley, line 1730: I think I prefer `sim_names` here still, but otherwise this looks great!

ntolley (approved):
Thanks @kmilo9999! 

Is it possible to unit test this? Perhaps just check that there isn't loaded data in this field, I imagine it could be a oneliner added to an existing GUI test that loads data

kmilo9999:
> Thanks @kmilo9999!
> 
> Is it possible to unit test this? Perhaps just check that there isn't loaded data in this field, I imagine it could be a oneliner added to an existing GUI test that loads data

Yes, I thought about it too. I'll work on it and add it to this PR

kmilo9999:
@ntolley 
I added the unit test. One last review and I think you can merge it. I need it in order to continue working in the list of gui bugs.

hnn_core/gui/gui.py
kmilo9999: done

PR #808, opened 2024-07-08
hierarchical json loading to gui 

gtdang:
Replacing the param and flat-json param files with hierarchical json for network configuration.

PR #828, opened 2024-07-22
[MRG] Simulate upload from local

gtdang:
Creates the ability to use local files for simulating GUI file uploads for unit-testing. Previously there was only the ability to load from URL sources. This will allow the use of local files so developers are not required an internet connection to run tests and to make use of default configuration files that are included in the repo.

Extracted from #808

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -794,15 +794,15 @@ class HNNGUI:
 
     # below are a series of methods that are used to manipulate the GUI
     def _simulate_upload_data(self, file_url):
-        uploaded_value = _prepare_upload_file_from_url(file_url)
+        uploaded_value = _prepare_upload_file(file_url)
         self.load_data_button.set_trait('value', uploaded_value)
 
     def _simulate_upload_connectivity(self, file_url):
-        uploaded_value = _prepare_upload_file_from_url(file_url)
+        uploaded_value = _prepare_upload_file(file_url)
         self.load_connectivity_button.set_trait('value', uploaded_value)
 
     def _simulate_upload_drives(self, file_url):
-        uploaded_value = _prepare_upload_file_from_url(file_url)
+        uploaded_value = _prepare_upload_file(file_url)
         self.load_drives_button.set_trait('value', uploaded_value)
 
     def _simulate_left_tab_click(self, tab_title):
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -881,7 +881,7 @@ def _prepare_upload_file(path):
     """
     try:
         uploaded_value = _prepare_upload_file_from_local(path)
-    except FileNotFoundError:
+    except (FileNotFoundError, OSError):
         uploaded_value = _prepare_upload_file_from_url(path)
 
     return uploaded_value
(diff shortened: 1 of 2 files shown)


ntolley:
I think there might need to be some logic for setting up windows specific file paths:
image

kmilo9999:
> I think there might need to be some logic for setting up windows specific file paths: image


I have something similar in test_gui.py:621  where the path is modified for windows systems. Maybe it works for you too.

gtdang:
@ntolley @kmilo9999 
Ok finally got the checks to pass. There were two issues to get them working on Windows. 

1. The code first tries to load whatever path you give it (URL or local path) locally, if that doesn't work it tries to use it as a URL. It does this by catching exceptions in a try/except clause. Giving a URL to a the `open` function on Unix will result in a `FileNotFound` error but on Windows it will be a `OSError` because the  `/` characters in URLs. Basically I just needed to also catch the `OSError`. 
2. I was doing asserts with file sizes. I learned that text files saved to disk on Windows are larger than Unix text files because their carriage returns/next line characters are larger.  This will result in a larger file on disk than if stream directly to memory from a URL. I changed the asserts to check if the files are decoded to a dictionary the same way.

hnn_core/gui/gui.py
```diff
@@ -836,21 +837,55 @@ def _simulate_viz_action(self, action_name, *args, **kwargs):
...
     }]
 
+    return upload_structure
+
+
+def _prepare_upload_file(path):
+    """ Simulates output of the FileUpload widget for testing.
+
+    Unit tests for the GUI simulate user upload of files. File source can
+    either be locally or from a URL. This function returns the data structure
```
ntolley, line 878:
```suggestion
    either be local or from a URL. This function returns the data structure
```

hnn_core/gui/gui.py
```diff
@@ -836,21 +837,55 @@ def _simulate_viz_action(self, action_name, *args, **kwargs):
         action(*args, **kwargs)
 
 
+def _prepare_upload_file_from_local(path):
+    with open(path, 'rb') as file:
+        content = memoryview(file.read())
+    last_modified = datetime.fromtimestamp(path.stat().st_mtime)
+
+    upload_structure = [{
+        'name': path.name,
```
ntolley, line 846: Is there any significance to using the full path for `name` in the local file vs. that last characters after a `'/'` in the URL upload? Is it just metadata so it doesn't matter that much?

hnn_core/tests/test_gui.py
```diff
@@ -50,6 +54,36 @@ def test_gui_load_params():
...
+    assert (content_from_url['name'] ==
+            content_from_local['name'] ==
+            'default.json')
+    assert (content_from_url['type'] ==
+            content_from_local['type'] ==
+            'application/json')
+    # Check that the size attribute is present. Cannot do an equivalency check
+    # because file systems may add additional when saving to disk.
+    assert content_from_url.get('size')
+    assert content_from_local.get('size')
```
ntolley, line 79:
```suggestion
    assert 'size' in content_from_url
    assert 'size' in content_from_local
```
Little more readable

hnn_core/gui/gui.py
gtdang: The name attribute of a Path object is the filename. Though this reminds me that all to add a conversion to Path in this function for if a string path is passed as an argument.

ntolley (approved):
All good on my end! merging now since this is pretty compact/self-contained

PR #829, opened 2024-07-23
[MRG] Fix update gui default dicts

gtdang:
Dictionaries that are nested within dictionaries are not updated with the update method. A recursive function was added to search for dictionary keys of nested dictionaries for updating. This is used for updating default values for GUI widgets. 

Changes:
* Added a recursive update function that will search nested dictionaries for keys to update
* Updated default values for poisson widget

split from #808

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -1099,10 +1099,10 @@ def _get_poisson_widget(name, tstop_widget, layout, style, location, data=None,
         'tstop': 0.0,
         'seedcore': 14,
         'rate_constant': {
-            'L5_pyramidal': 8.5,
-            'L2_pyramidal': 8.5,
-            'L5_basket': 8.5,
-            'L2_basket': 8.5,
+            'L5_pyramidal': 40.,
+            'L2_pyramidal': 40.,
+            'L5_basket': 40.,
+            'L2_basket': 40.,
         }
     }
     if isinstance(data, dict):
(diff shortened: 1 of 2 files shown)


hnn_core/gui/gui.py
```diff
@@ -888,6 +888,43 @@ def _prepare_upload_file(path):
     return uploaded_value
 
 
+def _update_nested_dict(original, new, skip_none=True):
+    """ Updates dictionary values from another dictionary
```
ntolley, line 892:
```suggestion
    """Updates dictionary values from another dictionary
```

hnn_core/gui/gui.py
```diff
@@ -888,6 +888,43 @@ def _prepare_upload_file(path):
...
+        None values in the new dictionary are not passed to the updated
+        dictionary by when True. If False None values will be passed to the
+        updated dictionary.
+
+    Returns
+    -------
+
+    """
+    updated = original.copy()
+    for key, value in new.items():
```
ntolley, line 915: Would this behave correctly if an empty dictionary is passed in new? Like say you wanted to clear the data in one of the dictionaries

gtdang: An empty dict would not change the original because you must have a matching key to update. But if you had matching keys and had null values you could update with null values by setting `skip_none` to False. 

Thanks for brining up this question though. This reminds me that I need to make a change to the logic. Right now it will not update 0 values in the 'new' dict by default because they evaluate as False at 920. It should be changed to an explicit `value != None`

gtdang:
@ntolley This one is rebased and ready to merge now.

ntolley (approved):
Looks great!! Merging now

PR #811, opened 2024-07-09
[MRG] ENH: Advanced Features and Optimization in `BatchSimulate`

samadpls:
A follow-up PR to extend and refine the functionality of `BatchSimulate`.

PR #830, opened 2024-07-24
[MRG] ENH: `BatchSimulate` Class

samadpls:
A follow-up PR to extend and refine the functionality of `BatchSimulate`.

- [x] Add options to specify which simulation results should be saved (dipole, spiking, LFP, voltages, currents, and calcium)
- [x] Add ability to save parameter values alongside simulation results
- [x] Add plot to example the shows dipole waveforms overlaid
- [x] Update x-axis on example to show the parameter value being varied (i.e. "Synaptic Strength (nS)")

- Linkcheck fails when BatchSimulate is added to the hnn_core __init__
 >  [To-do for me: figure out why linkcheck fails when BatchSimulate is added to the hnn_core __init__](https://github.com/jonescompneurolab/hnn-core/pull/811#issuecomment-2236774984)

### DIFF
--- a/examples/howto/plot_batch_simulate.py
+++ b/examples/howto/plot_batch_simulate.py
@@ -110,27 +110,32 @@ simulation_results = batch_simulation.run(param_grid,
 # backend='dask' if installed
 print("Simulation results:", simulation_results)
 ###############################################################################
-# Extract and Plot Results
-###############################################################################
+# This plot shows an overlay of all smoothed dipole waveforms from the
+# batch simulation. Each line represents a different set of parameters,
+# allowing us to visualize the range of responses across the parameter space.
 
-# Extract and overlay dipole waveforms
+
 dpl_waveforms = []
 for data_list in simulation_results['simulated_data']:
     for data in data_list:
-        dpl_waveforms.append(data['dpl'][0].data['agg'].copy().smooth(30))
+        window_len = 30
+        dpl_smooth = data['dpl'][0].copy().smooth(window_len)
+        dpl_waveforms.append(dpl_smooth.data['agg'])
 
 plt.figure(figsize=(10, 6))
 for waveform in dpl_waveforms:
-    plt.plot(waveform, alpha=0.5)
+    plt.plot(waveform, alpha=0.5, linewidth=3)
 plt.title('Overlay of Dipole Waveforms')
 plt.xlabel('Time (ms)')
 plt.ylabel('Dipole Amplitude (nAm)')
 plt.grid(True)
 plt.tight_layout()
-plt.savefig("image.png")
 plt.show()
+###############################################################################
+# This plot displays the minimum and maximum dipole peaks across
+# different synaptic strengths. This allows us to see how the range of
+# dipole activity changes as we vary the synaptic strength parameter.
 
-# Extract min and max peaks for plotting
 min_peaks, max_peaks, param_values = [], [], []
 for summary_list, data_list in zip(simulation_results['summary_statistics'],
                                    simulation_results['simulated_data']):
--- a/examples/howto/plot_batch_simulate.py
+++ b/examples/howto/plot_batch_simulate.py
@@ -86,7 +86,8 @@ def summary_func(results):
     """
     summary_stats = []
     for result in results:
-        dpl_data = result['dpl'][0].data['agg']
+        dpl_smooth = result['dpl'][0].copy().smooth(window_len=30)
+        dpl_data = dpl_smooth.data['agg']
         min_peak = np.min(dpl_data)
         max_peak = np.max(dpl_data)
         summary_stats.append({'min_peak': min_peak, 'max_peak': max_peak})
@@ -152,5 +153,6 @@ plt.ylabel('Dipole Peak Magnitude')
 plt.title('Min and Max Dipole Peaks across Simulations')
 plt.legend()
 plt.grid(True)
+plt.xscale('log')
 plt.tight_layout()
 plt.show()
--- a/examples/howto/plot_batch_simulate.py
+++ b/examples/howto/plot_batch_simulate.py
@@ -37,9 +37,6 @@ def set_params(param_values, net=None):
     net : instance of Network, optional
         If None, a new network is created using the specified model type.
     """
-    if net is None:
-        net = jones_2009_model()
-
     weights_ampa = {'L2_basket': param_values['weight_basket'],
                     'L2_pyramidal': param_values['weight_pyr'],
                     'L5_basket': param_values['weight_basket'],
@@ -102,7 +99,8 @@ def summary_func(results):
 
 
 # Run the batch simulation and collect the results.
-batch_simulation = BatchSimulate(set_params=set_params,
+batch_simulation = BatchSimulate(net=jones_2009_model(mesh_shape=(3, 3)),
+                                 set_params=set_params,
                                  summary_func=summary_func)
 simulation_results = batch_simulation.run(param_grid,
                                           n_jobs=n_jobs,
--- a/examples/howto/plot_batch_simulate.py
+++ b/examples/howto/plot_batch_simulate.py
@@ -118,8 +118,7 @@ print("Simulation results:", simulation_results)
 dpl_waveforms = []
 for data_list in simulation_results['simulated_data']:
     for data in data_list:
-        window_len = 30
-        dpl_smooth = data['dpl'][0].copy().smooth(window_len)
+        dpl_smooth = data['dpl'][0].copy().smooth(window_len=30)
         dpl_waveforms.append(dpl_smooth.data['agg'])
 
 plt.figure(figsize=(10, 6))
--- a/examples/howto/plot_batch_simulate.py
+++ b/examples/howto/plot_batch_simulate.py
@@ -99,7 +99,8 @@ def summary_func(results):
 
 
 # Run the batch simulation and collect the results.
-batch_simulation = BatchSimulate(net=jones_2009_model(mesh_shape=(3, 3)),
+net = jones_2009_model(mesh_shape=(3, 3))
+batch_simulation = BatchSimulate(net=net,
                                  set_params=set_params,
                                  summary_func=summary_func)
 simulation_results = batch_simulation.run(param_grid,
--- a/examples/howto/plot_batch_simulate.py
+++ b/examples/howto/plot_batch_simulate.py
@@ -117,7 +117,7 @@ print("Simulation results:", simulation_results)
 dpl_waveforms = []
 for data_list in simulation_results['simulated_data']:
     for data in data_list:
-        dpl_waveforms.append(data['dpl'][0].data['agg'])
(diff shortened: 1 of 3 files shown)


hnn_core/batch_simulate.py
```diff
@@ -19,7 +19,10 @@ def __init__(self, set_params, net_name='jones', tstop=170,
                  dt=0.025, n_trials=1, record_vsec=False,
                  record_isec=False, postproc=False, save_outputs=False,
                  save_folder='./sim_results', batch_size=100,
-                 overwrite=True, summary_func=None):
+                 overwrite=True, summary_func=None,
+                 save_dpl=True, save_spiking=False,
+                 save_LFP=False, save_voltages=False,
```
ntolley, line 24:
```suggestion
                 save_lfp=False, save_voltages=False,
```
Just to keep consistent as we generally reserve uppercase characters for classes

hnn_core/batch_simulate.py
```diff
@@ -70,10 +73,22 @@ def __init__(self, set_params, net_name='jones', tstop=170,
...
+            If True, save voltages results. Default: False.
+        save_currents : bool
+            If True, save currents results. Default: False.
+        save_calcium : bool
+            If True, save calcium concentrations. Default: False.
         Notes
         -----
         When `save_output=True`, the saved files will appear as
-        `sim_run_{start_idx}-{end_idx}.npy` in the specified `save_folder`
+        `sim_run_{start_idx}-{end_idx}.npz` in the specified `save_folder`
```
ntolley, line 91: I remember finding a workaround at some point for combining`spike_<times/spikes/gids>` in a single `.npy` file, however this seems a bit more explicit

hnn_core/batch_simulate.py
```diff
@@ -110,6 +125,12 @@ def __init__(self, set_params, net_name='jones', tstop=170,
         self.batch_size = batch_size
         self.overwrite = overwrite
         self.summary_func = summary_func
+        self.save_dpl = save_dpl
```
ntolley, line 128: Add type checks and unit tests for `bool`

hnn_core/batch_simulate.py
```diff
@@ -70,10 +73,22 @@ def __init__(self, set_params, net_name='jones', tstop=170,
         summary_func : callable, optional
```
ntolley, line 73: look into if it makes sense if we should allow exporting a `.pkl` of these callables, or just have users be responsible for knowing how the parameters in the array are used

hnn_core/batch_simulate.py
```diff
@@ -70,10 +73,22 @@ def __init__(self, set_params, net_name='jones', tstop=170,
...
             A function to calculate summary statistics from the simulation
             results. Default is None.
+        save_dpl : bool
+            If True, save dipole results. Default: True.
```
ntolley, line 77:
```suggestion
            If True, save dipole results. Note, `save_outputs` must be True.
            Default: True.
```

hnn_core/batch_simulate.py
```diff
@@ -307,13 +356,135 @@ def _save(self, results, start_idx, end_idx):
...
+                                 f'sim_run_{start_idx}-{end_idx}.npz')
         if os.path.exists(file_name) and not self.overwrite:
             raise FileExistsError(
                 f"File {file_name} already exists and "
                 "overwrite is set to False.")
 
-        np.save(file_name, sim_data)
+        np.savez(file_name, **save_data)
+
+    def _save_parameters(self, param_combinations, start_idx, end_idx):
```
ntolley, line 378: Decide if you want to have a single file both data and parameters, or split between a data file and a dedicated parameters file

hnn_core/tests/test_batch_simulate.py
```diff
@@ -210,3 +218,79 @@ def test_save_load_and_overwrite(batch_simulate_instance,
...
+    assert 'dpl' in loaded_results
+    assert len(loaded_results['dpl']) == len(results)
+
+    original_data = np.stack([result['dpl'][0].data['agg']
+                              for result in results])
+    loaded_data = np.stack([dpl[0].data['agg']
+                            for dpl in loaded_results['dpl']])
+    assert np.array_equal(original_data, loaded_data)
+
+    for key in ['spiking', 'LFP', 'voltages', 'currents', 'calcium']:
```
ntolley, line 250:
```suggestion
    for key in ['spiking', 'lfp', 'voltages', 'currents', 'calcium']:
```

hnn_core/batch_simulate.py
```diff
@@ -70,10 +73,22 @@ def __init__(self, set_params, net_name='jones', tstop=170,
...
+        save_spiking : bool
+            If True, save spiking results. Default: False.
+        save_lfp : bool
+            If True, save local field potential (lfp) results. Default: False.
+        save_voltages : bool
+            If True, save voltages results. Default: False.
+        save_currents : bool
+            If True, save currents results. Default: False.
+        save_calcium : bool
+            If True, save calcium concentrations. Default: False.
```
ntolley, line 87: Add `self.clear_cache=False` to the `__init__()`, this will allow you to store simulation results in a class attribute

ntolley:
Here's what to add to make plotting with this class easier
1) Add `clear_cache` as a class attribute 
2) If `clear_cache is False`, then `batch_simulate.run()` should return a dictionary that contains summary statistics, and the simulated data that is supposed to be saved (dpl, spiking, etc.)
3) If `clear_cache is True`, then `batch_simulate.run()` returns a dictionary where the only entry contains the summary statistics

hnn_core/batch_simulate.py
ntolley, line 73: Also is this is `None`, and `clear_cache=True`, then there should be no return from `batch_simulate.run()`

examples/howto/plot_batch_simulate.py
```diff
@@ -113,18 +113,37 @@ def summary_func(results):
 # Extract and Plot Results
 ###############################################################################
 
+# Extract and overlay dipole waveforms
```
ntolley, line 116: Break into two cells with a short sentence describing the two plots

examples/howto/plot_batch_simulate.py
```diff
@@ -113,18 +113,37 @@ def summary_func(results):
...
+dpl_waveforms = []
+for data_list in simulation_results['simulated_data']:
+    for data in data_list:
+        dpl_waveforms.append(data['dpl'][0].data['agg'])
```
ntolley, line 120:
```suggestion
        dpl_waveforms.append(data['dpl'][0].data['agg'].copy().smooth(30))
```

examples/howto/plot_batch_simulate.py
```diff
@@ -113,18 +113,37 @@ def summary_func(results):
...
+
+plt.figure(figsize=(10, 6))
+for waveform in dpl_waveforms:
+    plt.plot(waveform, alpha=0.5)
```
ntolley, line 124:
```suggestion
    plt.plot(waveform, alpha=0.5, linewidth=2)
```
Try and see if thicker lines look a little better

hnn_core/batch_simulate.py
```diff
@@ -133,21 +175,20 @@ def run(self, param_grid, return_output=True,
...
-            Default is False
 
         Returns
         -------
-        results : list
-            List of simulation results if return_output is True.
+        results : dict
+            Dictionary containing summary statistics and simulation data
+            if return_output is True and clear_cache is False. Otherwise,
+            a dictionary containing only summary statistics.
```
ntolley, line 184: Add list of all possible keys ('dpl', 'lfp', etc.)

hnn_core/batch_simulate.py
```diff
@@ -252,15 +299,39 @@ def _run_single_sim(self, param_values):
             net = calcium_model()
 
         self.set_params(param_values, net)
```
ntolley, line 301: We should consider how to integrate #763 with this PR such that users can load in base networks of their own choosing

ntolley: Read some of the code in #837 to see how instantiating networks from the `json` format looks

ntolley: Here is the function that converts exported `json` files to a network https://github.com/jonescompneurolab/hnn-core/blob/773a86d9955eadb6ab4eae39bd9c3f2e91c8f403/hnn_core/hnn_io.py#L384

examples/howto/plot_batch_simulate.py
```diff
@@ -110,21 +110,45 @@ def summary_func(results):
...
-min_peaks, max_peaks = [], []
-for res in simulation_results:
-    summary_stats = res[0]
-    min_peaks.append(summary_stats['min_peak'])
-    max_peaks.append(summary_stats['max_peak'])
+dpl_waveforms = []
+for data_list in simulation_results['simulated_data']:
+    for data in data_list:
+        window_len = 30
+        dpl_smooth = data['dpl'][0].copy().smooth(window_len)
```
ntolley, line 122:
```suggestion
        dpl_smooth = data['dpl'][0].copy().smooth(window_len=30)
```

ntolley: Can you add `.copy().smooth(window_len=30)` to the summary statistic function as well?

Right now the values of the min/max peaks don't match what you see in the dipole waveform plot

hnn_core/batch_simulate.py
```diff
@@ -252,15 +305,39 @@ def _run_single_sim(self, param_values):
             net = calcium_model()
```
ntolley, line 305: @samadpls I'm realizing that this may be difficult to maintain in the future.

Instead of a `net_name` attribute passed to the init, could we just pass in a `Network` object directly? Then it is stored as `self.net` in the class, and you can call `net.copy()` whenever you run a single simulation

ntolley: This came to mind because I was going to recommend running a smaller network using the following command
```py
net = jones_2009_model(mesh_shape(3,3))
```
but this isn't possible with the current API

examples/howto/plot_batch_simulate.py
ntolley, line 1: This example nearly doubles the circleCI build time because it runs so many simulations, can you try passing `net = jones_2009_model(mesh_shape=(3, 3))` to the BatchSimulate class? This should speed up the simulation time a lot

If it's still slow we can rename this file to `batch_simulate.py` so that it is skipped over during the circleCI build

examples/howto/plot_batch_simulate.py
```diff
@@ -101,7 +99,8 @@ def summary_func(results):
 
 
 # Run the batch simulation and collect the results.
-batch_simulation = BatchSimulate(set_params=set_params,
+batch_simulation = BatchSimulate(net=jones_2009_model(mesh_shape=(3, 3)),
```
ntolley, line 102:
```suggestion
net = jones_2009_model(mesh_shape=(3, 3))
batch_simulation = BatchSimulate(net=net),
```

ntolley:
Moving forward with merge since the linkcheck failure is a separate issue

Great work @samadpls!!

PR #834, opened 2024-07-24
fix: missing n_drive_cells argument during configuration to Network conversion

gtdang:
Added `n_drives_cells` argument to `add_evoked_drive` in the `read_external_drive` function used when converting a network configuration file to Network.

split from #808


ntolley:
Great finding this @gtdang !!

Issue #832, opened 2024-07-24
GUI: Incomplete Plot types after loading data


kmilo9999:
Third Bug reported by Dylan

After loading data, the plot types dropdown in the figures control tab doesnt contain all the set of possible plot types.

current dipole
layer2 dipole
layer5 dipole
PSD
spectrogram

Missing:
'spikes', 'input histogram', 'network'

kmilo9999:
@ntolley @gtdang 

This problem comes from PR #616 [this ](https://github.com/jonescompneurolab/hnn-core/pull/616/commits/8b9af574ae8c54ce6fac4c303944a88dfe212b9b) commit.

```
     sim_name_default = simulation_names[-1]

    if data['simulations'][sim_name_default]['net'] is None:
        valid_plot_types = [
            pt for pt in _plot_types if pt not in _ext_data_disabled_plot_types
        ]
    else:
        valid_plot_types = _plot_types
```

The plot types dropdown list values depend on the type of the last simulation in the  `data['simulations']` dictionary.
This code is executed after running a simulation, loading data or pressing the 'Make figure' button.

ntolley:
Trying to think of what a clean solution is here, The only plots that make sense to overlay are 
- current dipole
- layer2 dipole
- layer5 dipole
- PSD

The spectrogram should be excluded since it used `imshow` and can't simultaneously show two sets of data

I think the correct behavior would be to have `valid_plot_types` depend on the currently selected `Data to compare`, so perhaps a callback that updates the `plot types` whenever you selected a different `Data to compare`

What do you think?

gtdang:
I think we should remove this logic. It's making an assumption of what a user's next action will definitely be adding observed data to a plot. That's not always the case.

gtdang:
We discussed this more at today's dev-sync. 

* The limiting of the plot types logic is in place for when selecting simulation vs loaded data to plot. We do want the limiting of plot types for loaded data. 
* The problem is that when creating a new figure. It decides to set defaults based on the last item stored in the `data` object, which stores both simulation and loaded data. If the data is loaded last, all new figure panes will start with plot types for loaded data. 

The solution is to initialize new figure panes with the last simulation and not loaded data.

Issue #831, opened 2024-07-24
GUI - Plot templates dont work after loading data


kmilo9999:
After running simulation and loading data, plot templates dont work

PR #836, opened 2024-07-28
[MRG] Bug: Incomplete Plot types list in visualization tab

kmilo9999:
Fixing bug on the available plot type list in the viz tab
solves #832 
solves #831

### DIFF
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -1034,6 +1034,4 @@ class _VizManager:
 
 def _is_simulation(data):
     """Determines if saved data is a simulation."""
-    if data['net'] is not None:
-        return True
-    return False
+    return data['net'] is not None
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -138,7 +138,7 @@ data_templates = {
 
 def check_sim_plot_types(
         new_sim_name, plot_type_selection, target_selection, data):
-    if data["simulations"][new_sim_name]['net'] is None:
+    if not _is_simulation(data["simulations"][new_sim_name]):
         plot_type_selection.options = [
             pt for pt in _plot_types if pt not in _ext_data_disabled_plot_types
         ]
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -489,32 +489,29 @@ def _get_ax_control(widgets, data, fig_idx, fig, ax):
     analysis_style = {'description_width': '200px'}
     layout = Layout(width="98%")
     simulation_names = tuple(data['simulations'].keys())
-    sim_name_default = simulation_names[-1]
-
-    if len(simulation_names) == 0:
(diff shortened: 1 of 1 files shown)


kmilo9999:
@ntolley @gtdang Some loaded data doesnt produce `layer2 dipole `and `layer5 dipole` figures (i.e `yes_trial_S1_ERP_all_avg`). I dont know if this is always the case though. Should we add a message in the console indicating that no figure is produced?

hnn_core/gui/_viz_manager.py
```diff
@@ -489,32 +489,29 @@ def _get_ax_control(widgets, data, fig_idx, fig, ax):
...
+    sim_index = 0
+    if not simulation_names:
+        simulation_names = ("None",)
+    else:
+        # Find the last simulation with a non-None 'net'
+        sim_index = next(
+            (idx for idx, sim_name in enumerate(simulation_names)
+             if data['simulations'][sim_name]['net'] is not None),
+            0  # Default value if no such simulation is found
+        )
```
gtdang, line 501:
```suggestion
        # Find the last simulation with a non-None 'net'
        sim_index = next(
            (idx for idx, sim_name in 
             reversed(list(enumerate(simulation_names)))
             if data['simulations'][sim_name]['net'] is not None),
            0  # Default value if no such simulation is found
        )
```
This will find the last simulation instead of the first.

hnn_core/gui/_viz_manager.py
gtdang, line 1: We use the `data["simulations"][new_sim_name]['net'] is None` logic several times in this file. Adding a helper function would abstract this logic and help readability. 
```python
def _is_simulation(data):
    """Determines if saved data is a simulation."""
    if data['net'] is not None:
        return True
    return False
```

Then the statements can be: 
```python
if _is_simulation(data["simulations"][new_sim_name]):
    ...
# OR
if not _is_simulation(data["simulations"][new_sim_name]):
    ...
```

gtdang (changes requested):
I've played around it with it and it's looking good. I have a few changes suggested.  Thanks!

gtdang (approved):
Looks great!

gtdang (approved):
Looks great! Thanks Camilo!

hnn_core/gui/_viz_manager.py
```diff
@@ -1026,3 +1030,10 @@ def _simulate_edit_figure(self, fig_name, ax_name, simulation_name,
             buttons.children[0].click()
         elif operation == "clear":
             buttons.children[1].click()
+
+
+def _is_simulation(data):
+    """Determines if saved data is a simulation."""
+    if data['net'] is not None:
+        return True
+    return False
```
samadpls, line 1039:
```suggestion
    return data['net'] is not None
```
Great work! Just a few minor refactoring suggestions. Feel free to accept them if you agree.

ntolley (approved):
All fine with me! I'll merge once you make a decision on @samadpls's suggestion

hnn_core/gui/_viz_manager.py

PR #837, opened 2024-07-29
[MRG] GUI new json loading

gtdang:
Changed the configuration/parameter file format support of the GUI. 
* The loading of connectivity and drives use the new multi-level json structure that mirrors the structure of the Network object. 
* **BREAKING:** Flat parameter and json configuration files are no longer supported by the GUI

split from #808


hnn_core/gui/gui.py
```diff
@@ -481,11 +486,10 @@ def data(self):
     @staticmethod
     def load_parameters(params_fname=None):
```
ntolley, line 487: Does it make sense to have a default of `None` anymore? Does it always need a valid file name?

gtdang: Ah right. Yeah, I'll get rid of that default value. Thanks!

hnn_core/gui/gui.py
```diff
@@ -1493,10 +1498,8 @@ def get_cell_param_default_value(cell_type_key, param_dict):
 
 def add_drive_tab(params, log_out, drives_out, drive_widgets, drive_boxes,
                   tstop, layout):
-    net = jones_2009_model(params)
-
-    drive_specs = _extract_drive_specs_from_hnn_params(
```
ntolley, line 1498: We can likely deprecate (or just remove since it's private) this function in the future

hnn_core/param/jones2009_base.json
ntolley, line 1: Perhaps there should be a test to make sure this is synchronized with the file in `hnn_data`?

hnn_core/tests/test_gui.py
```diff
@@ -85,49 +91,67 @@ def _import_json(content):
...
+
+    # Load connectivity and make sure drives did not change
+    gui._simulate_upload_connectivity(file1_url)
+    assert len(gui.drive_widgets) == 1
 
-    # file2: connectivity file1: drives
-    gui._simulate_upload_connectivity(file2_url)
-    # now connectivity is refreshed.
-    assert gui.connectivity_widgets[0][0].children[1].value == 0.01
-    assert gui.drive_widgets[-1]['tstop'].value == 250.
```
ntolley, line 128: this would be a good one to keep for the poisson drive in `file2` to make sure it's loaded properly

ntolley (approved):
Happy with the changes on my end! Just a few test suggestions otherwise good to go

hnn_core/gui/gui.py
gtdang: Sure. Just need to check if it's only use is in the GUI.

hnn_core/param/jones2009_base.json
gtdang: That's a good idea. I'll make it an issue.

hnn_core/tests/test_gui.py
gtdang: The weight check is currently preserved on line 116. However the stop time check is not because tstop isn't currently saved with the Network configurations.

ntolley: I was thinking that this was referring to the stop time of the drive?

gtdang: oh you're right! hmm... I can add it back in. I'm curious if the drive widget will set it to the tstop widget if the config tstop is greater than the the widget...

gtdang: I added the test but set it to 170 instead of 250. The logic sets the drive tstop to the GUI's widget tstop value if the drive tstop value is larger. We currently don't save a universal tstop value to Network object, so the GUI doesn't set its tstop widget from Network configuration files. 

This may change in the future if we want to save a stop value with a Network.

ntolley (approved):

PR #843, opened 2024-07-31
[WIP] Feat GUI add connection probability

gtdang:
Add connection probability to connection accordion boxes.

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -937,7 +937,7 @@ def _get_connectivity_widgets(conn_data):
 
     style = {'description_width': '150px'}
     style = {}
-    sliders = list()
+    conn_widgets = list()
     for receptor_name in conn_data.keys():
         w_text_input = BoundedFloatText(
             value=conn_data[receptor_name]['weight'], disabled=False,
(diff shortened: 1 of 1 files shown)


hnn_core/gui/gui.py
```diff
@@ -937,17 +937,23 @@ def _get_connectivity_widgets(conn_data):
...
-    sliders = list()
+    conn_widgets = list()
     for receptor_name in conn_data.keys():
-        w_text_input = BoundedFloatText(
+        weight_text_input = BoundedFloatText(
             value=conn_data[receptor_name]['weight'], disabled=False,
             continuous_update=False, min=0, max=1e6, step=0.01,
             description="weight", style=style)
+        probability_text_input = BoundedFloatText(
+            value=conn_data[receptor_name]['probability'], disabled=False,
```
ntolley, line 947: A simple test to run would be to reduce the probability of one of the connections, and assert that there are fewer `gid_pairs`

gtdang:
This PR is on hold. The problem is that if probability is changed, gid_pairs also needs to be recalculated. Work from this PR can be used once a means to update connection probabilities and associated parameters is implemented. This can either be on an API level, or changing the logic of the GUI to rebuild connections on Network init.

A patch fix to remove the GUI probability assignment logic was made in https://github.com/jonescompneurolab/hnn-core/pull/844.

PR #845, opened 2024-07-31
[MRG] fix: removed unnecessary display call

gtdang:
The extra display call was not needed because the display is called within the function to render the cell params. Also the function it was wrapping had no return, thus causing a "None" to print to the console.


ntolley (approved):

ntolley:
Perfect!! Thanks a bunch @gtdang

PR #846, opened 2024-07-31
[MRG] GUI load confirmation message

gtdang:
Added a confirmation message when connectivity or drives are loaded from file. 

Screenshot 2024-07-31 at 5 06 42 PM


ntolley (approved):
Perfect! When building the tutorials we can brainstorm about what other messages might be useful to print

PR #848, opened 2024-08-01
[MRG] make L5/L2 pyr have different geometry options in GUI

ntolley:
The current implementation of the cell parameter widget skips over a compartment of L5 pyramidal cells

This PR correctly updates the compartment parameters, as well as improves testing to make sure the GUI network matches the API instantiated network

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -14,6 +14,7 @@ import urllib.request
 from collections import defaultdict
 from pathlib import Path
 from datetime import datetime
+from functools import partial
 from IPython.display import IFrame, display
 from ipywidgets import (HTML, Accordion, AppLayout, BoundedFloatText,
                         BoundedIntText, Button, Dropdown, FileUpload, VBox,
@@ -29,6 +30,7 @@ from hnn_core.dipole import _read_dipole_txt
 from hnn_core.params_default import (get_L2Pyr_params_default,
                                      get_L5Pyr_params_default)
 from hnn_core.hnn_io import dict_to_network
+from hnn_core.cells_default import _exp_g_at_dist
 
 import base64
 import zipfile
@@ -41,7 +43,7 @@ default_network_configuration = (hnn_core_root / 'param' /
 
 cell_parameters_dict = {
 
-    "Geometry":
+    "Geometry L2":
     [
         ('Soma length', 'micron', 'soma_L'),
         ('Soma diameter', 'micron', 'soma_diam'),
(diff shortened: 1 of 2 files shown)


kmilo9999:
@ntolley The code looks for variables in the `params_default:get_L2Pyr_params_default()` script and match them with the 3rd parameter in the large gui.py dictionary. You removed  ` # ('Apical Dendrite 2 length', 'micron', '**apical2_L**'),()` , but that one still exist in `params_defaul`   
Let me know the next steps.

ntolley:
@kmilo9999 the original code actually didn't have an `apical_2` key, the commented out line was just for my own reference

L2 and L5 pyramidal cells differ in that L5_pyramidal cells have an `apical_2` section, whereas L2_pyramidal cells do not (which matches with the entries in `get_L2Pyr_params_default() `)

What I'm trying to do is have the `apical_2` key be looked up only when an L5 pyramidal cell is selected

kmilo9999:
I think the loops in the `add_cell_parameters_tab` needs to be reworked. It's not doing what it supposed and the logic it's hard to follow.
Should I work it on a different PR or on this one ?

ntolley:
@kmilo9999 I agree we can revamp the logic in a followup PR

In any case, I think I've updated the mappings that needed to be added and wrote some tests that they match what we get from making a network with the CLI

Do you mind reviewing to make sure everything is correct?

kmilo9999:
Are  you getting the right dipoles if you run with/out the UI cell params?

This is what I got:

Using UI cell params
image

Not using the UI cell params (default values)
image

The UI is just loading the default values from `params_default` and send it to the simulation. Running the default UI values should throw the same graph as if there was no UI cell params.

hnn_core/gui/gui.py
```diff
@@ -1669,7 +1701,12 @@ def _init_network_from_widgets(params, dt, tstop, single_simulation_data,
                 cell_type = vbox_key.split()[0]
                 update_function(single_simulation_data['net'], cell_type,
                                 cell_param_list.children)
-                break
+                # break  # why is this here?
```
ntolley, line 1704: @kmilo9999 let me know when you have time to look at this

kmilo9999: It's a small optimization.
Once the conditions are meet and a function is called, the loop doesnt need to continue to go through unnecessary iterations. This helps on correctness to guaranty no other calls to the same conditions are made.

hnn_core/gui/gui.py
```diff
@@ -1963,7 +2031,7 @@ def update_common_dendrite_sections(sections, mechs_params):
         name for name in sections.keys() if name != 'soma'
     ]
     for section in dendrite_sections:
-        sections[section].mechs.update(mechs_params)
+        sections[section].mechs.update(deepcopy(mechs_params))
```
kmilo9999, line 2034: Is this deepcopy necessary ? This operations hit the performance of the app

kmilo9999 (approved):
I left a comment in this PR. If you think everything is running correctly then I believe is fine.

hnn_core/gui/gui.py
ntolley: it is necessary because a nested dictionary is being added, I tried a regular `.copy()` and found that the inner dictionary was still pointing to the same variable

This is only necessary for the `ar` mechanism since `gbar_ar` is a partial function who's outputs depend on the geometry of each `section`

ntolley:
Sounds great @kmilo9999! 

I'm good with the current changes so feel free to merge when you're ready. I have some ideas for variable name changes to increase clarity but we can work that out separately

hnn_core/gui/gui.py
kmilo9999: I'll revisit this in a future PR. I can improve this piece of code for sure.

Issue #850, opened 2024-08-06
GUI: Network configuration different after GUI initialization


gtdang:
The GUI does not reconstruct the same network connectivity and drives after reconstruction from widget values. 

__Test:__
1. Load test network configuration `hnn_core/tests/assets/jones2009_3x3_drives.json`. This config has all drive types (evoked, bursty, poisson) and tonic bias added. 
2. Run a simulation 
3. Compare the Network used to run the simulation (derived from GUI widget values) to a Network created by the API `read_network_connectivity` function.

__Problems:__
1. Number of connections is not the same. There are more bursty and poisson connections.
    - The GUI passes weights and constants dictionaries by cell types to the add_x_drive functions. Because there is an input box for each cell type, even weights with 0 are passed into the drive construction. This creates additional unnecessary null-weight connections.
    -  The GUI doesn't have an input widget for the n_drive_cells parameter. It uses the defaults 'n_cells' for evoked and poisson, and 1 for bursty. The synchronous inputs checkbox does send a n_drive_cells value to drive construction but otherwise the parameter is hidden from user input. But with the default bursty drives are always synchronous. 
2. The order of the drives is not the same. They are sorted alphabetically when read into the GUI. 
    - gid_ranges and pos_dict are not the same as a consequence of 1 and 2
3. Tonic bias is not in the GUI-derived Network 
    - This is because there isn't a way to load tonics into the GUI from file yet. 
4. Cell parameters are not the same. This is being addressed in #848  

__To Do:__
- [x]  Fix drive order when loading drives https://github.com/jonescompneurolab/hnn-core/pull/851
- [x]  Add tonic loading from file #852 
- [x]  Filter weights and constants for only cells with non-zero weights during drive construction. #853
- [x]  Add n_drive_cells input widget #861

gtdang:
@ntolley 
I'm working on the n_drive_cells and sync_input rework today. I just want to make sure I understand the API correctly. Is this understanding correct?

1.`n_drive_cells=1 and cell_specific=False` - *Synchronous input*: 1 drive cell is connected to all network cells
2.`n_drive_cells=3 and cell_specific=False` - 3 drive cells are connected to all network cells
3.`n_drive_cells='n_cells' and cell_specific=True` - a unique drive cell gets assigned to each available simulated cell in the network with 1-to-1 connectivity
4.`n_drive_cells='n_cells' and cell_specific=False` - This is invalid
[...]

PR #851, opened 2024-08-06
[MRG] GUI: fix drive sorting

gtdang:
Removed the sorting of drive widgets alphabetically.

Discussed in #850

### DIFF
--- a/hnn_core/tests/test_gui.py
+++ b/hnn_core/tests/test_gui.py
@@ -141,6 +141,7 @@ def test_gui_upload_drives():
     # simulate upload default.json
     file1_url = Path(hnn_core_root, 'param', 'jones2009_base.json')
     file2_url = Path(assets_path, 'gamma_L5weak_L2weak_hierarchical.json')
+    file3_url = Path(assets_path, 'jones2009_3x3_drives.json')
 
     # check if parameter reloads
     gui._simulate_upload_drives(file1_url)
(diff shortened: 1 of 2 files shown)


hnn_core/tests/test_gui.py
```diff
@@ -164,6 +165,13 @@ def test_gui_upload_drives():
     gui._simulate_upload_connectivity(file1_url)
     assert len(gui.drive_widgets) == 1
 
+    # Load file with more drives and make sure it's in the right order
+    gui.delete_drive_button.click()
+    gui._simulate_upload_drives(file3_url)
+    drive_names = [widget['name'] for widget in gui.drive_widgets]
+    assert drive_names == ['evdist1', 'evprox1', 'evprox2',
```
ntolley, line 172: One test that comes to mind is 
1) add drives through the GUI
2) export to file
3) load back in and ensure order matches

I'm assuming this isn't possible until the load button is actually implemented?

ntolley (approved):
Thanks @gtdang!!

Let me know if the test I recommended makes sense, and if we should wait until the load button is set up to implement it

hnn_core/tests/test_gui.py
gtdang: I think that test would be more appropriate for an export test. Since it's really testing that the gui is able to write a file with the correct structure. This one is testing that the file upload respects the order of the input file's drives.

PR #852, opened 2024-08-07
[MRG] feat: GUI tonic loading

gtdang:
Updated the GUI load drive widget to be able to load tonic biases from a network config file.

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -1254,33 +1254,38 @@ def _get_evoked_widget(name, layout, style, location, data=None,
     return drive, drive_box
 
 
-def _get_tonic_widget(name, layout, style, data=None):
-
-    default_data = {
+def _get_tonic_widget(name, tstop_widget, layout, style, data=None):
+    cell_types = ['L2_basket', 'L2_pyramidal', 'L5_basket', 'L5_pyramidal']
+    default_values = {
         'amplitude': 0,
         't0': 0,
-        'tstop': 0
+        'tstop': tstop_widget.value
     }
+    t0 = default_values['t0']
+    tstop = default_values['tstop']
+    default_data = {cell_type: default_values for cell_type in cell_types}
 
-    cell_types = ['L2_basket', 'L2_pyramidal', 'L5_basket', 'L5_pyramidal']
-
-    if isinstance(data, dict):
-        default_data.update(data)
     kwargs = dict(layout=layout, style=style)
+    if isinstance(data, dict):
+        default_data = _update_nested_dict(default_data, data)
 
     amplitudes = dict()
     for cell_type in cell_types:
-        if cell_type in data:
-            default_data.update(data[cell_type])
+        amplitude = default_data[cell_type]['amplitude']
         amplitudes[cell_type] = BoundedFloatText(
-            value=deepcopy(default_data['amplitude']),
-            description=cell_type, min=0, max=1e6, step=0.01, **kwargs)
+            value=amplitude, description=cell_type,
+            min=0, max=1e6, step=0.01, **kwargs)
+        # Reset the global t0 and stop with values from the 'data' keyword.
+        # It should be same across all the cell-types.
+        if amplitude > 0:
+            t0 = default_data[cell_type]['t0']
+            tstop = default_data[cell_type]['tstop']
 
     start_times = BoundedFloatText(
-        value=deepcopy(default_data['t0']), description="Start time",
+        value=t0, description="Start time",
         min=0, max=1e6, step=1.0, **kwargs)
     stop_times = BoundedFloatText(
-        value=deepcopy(default_data['tstop']), description="Stop time",
+        value=tstop, description="Stop time",
         min=-1, max=1e6, step=1.0, **kwargs)
 
     widgets_dict = {
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -7,11 +7,13 @@ import io
 import logging
 import mimetypes
 import multiprocessing
+import numpy as np
 import sys
 import json
 import urllib.parse
 import urllib.request
 from collections import defaultdict
+from copy import deepcopy
 from pathlib import Path
 from datetime import datetime
 from functools import partial
@@ -34,15 +36,12 @@ from hnn_core.cells_default import _exp_g_at_dist
 
 import base64
 import zipfile
-import numpy as np
-from copy import deepcopy
 
 hnn_core_root = Path(hnn_core.__file__).parent
 default_network_configuration = (hnn_core_root / 'param' /
                                  'jones2009_base.json')
 
 cell_parameters_dict = {
-
     "Geometry L2":
     [
         ('Soma length', 'micron', 'soma_L'),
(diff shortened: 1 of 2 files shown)


hnn_core/tests/test_gui.py
```diff
@@ -716,9 +730,12 @@ def test_gui_upload_csv_simulation(setup_gui):
                 ['dpls'][0].data['L5']) == data_lengh)
 
 
-def test_gui_add_tonic_input(setup_gui):
+def test_gui_add_tonic_input():
     """Test if gui add different type of drives."""
-    gui = setup_gui
+    gui = HNNGUI()
+    _ = gui.compose()
```
gtdang, line 736: Changed this because the setup_gui already has a tonic and we want to test adding a tonic.

gtdang:
I'm not sure why linkcheck keeps failing with this pr. It works locally...

hnn_core/gui/gui.py
```diff
@@ -1254,33 +1253,38 @@ def _get_evoked_widget(name, layout, style, location, data=None,
...
-
-    if isinstance(data, dict):
-        default_data.update(data)
     kwargs = dict(layout=layout, style=style)
+    if isinstance(data, dict):
+        default_data = _update_nested_dict(default_data, data)
 
     amplitudes = dict()
     for cell_type in cell_types:
-        if cell_type in data:
```
ntolley, line 1273: `update_nested_dict()` gets rid of the need for this logic?

gtdang: Yep. The default_data dict is built on line 1265.  It is a nested dict by cell_type. Then update_nested_dict updates that structure with the data, which is also in that structure.

kmilo9999:
> I'm not sure why linkcheck keeps failing with this pr. It works locally...

I think is something related to:

```
* AttributeError: module 'hnn_core' has no attribute 'read_network'
* ModuleNotFoundError: No module named 'hnn_core.read_network'
```

gtdang:
> > I'm not sure why linkcheck keeps failing with this pr. It works locally...
> 
> I think is something related to:
> 
> ```
> * AttributeError: module 'hnn_core' has no attribute 'read_network'
> * ModuleNotFoundError: No module named 'hnn_core.read_network'
> ```

Hm thanks for finding that. That's strange that function hasn't been around for some time... Wonder why it's deciding to error now.

gtdang:
Well that error goes away but it doesn't seem like that's what causes the test to fail... 

```
(auto_examples/workflows/plot_simulate_beta: line  330) broken    https://doi.org/10.1101/2021.04.16.440210 - 403 Client Error: Forbidden for url: https://www.biorxiv.org/lookup/doi/10.1101/2021.04.16.440210
```
The log shows the above as the broken link consistently. The links work on my end though. And when I run the linkcheck locally it shows it as working...

hnn_core/gui/gui.py
```diff
@@ -1254,33 +1253,38 @@ def _get_evoked_widget(name, layout, style, location, data=None,
...
-            default_data.update(data[cell_type])
+        amplitude = default_data[cell_type]['amplitude']
         amplitudes[cell_type] = BoundedFloatText(
-            value=deepcopy(default_data['amplitude']),
```
ntolley, line 1276: is the deepcopy no longer necessary because the value is assigned to variable above?

hnn_core/gui/gui.py
```diff
@@ -1393,7 +1397,7 @@ def add_drive_widget(drive_type, drive_boxes, drive_widgets, drives_out,
 
             for idx, drive in enumerate(drive_widgets):
                 tab_name = drive['name']
-                if "Tonic" not in tab_name:
+                if "tonic" not in tab_name.lower():
```
ntolley, line 1400: any reason to prefer not matching by case?

hnn_core/gui/gui.py
```diff
@@ -1530,23 +1534,38 @@ def add_drive_tab(params, log_out, drives_out, drive_widgets, drive_boxes,
...
+    tonic_specs = net.external_biases
 
     # clear before adding drives
     drives_out.clear_output()
     while len(drive_widgets) > 0:
         drive_widgets.pop()
         drive_boxes.pop()
 
-    drive_names = sorted(drive_specs.keys())
+    drive_names = list(drive_specs.keys())
```
ntolley, line 1545: This might cause a merge conflict with #851

hnn_core/tests/test_gui.py
```diff
@@ -734,20 +752,33 @@ def test_gui_add_tonic_input(setup_gui):
...
+    assert last_drive['tstop'].value == 170.0
+    assert last_drive['amplitude']['L5_pyramidal'].value == 0
+
+    # Set new widget values
+    last_drive['t0'].value = 0
+    last_drive['tstop'].value = 15
+    last_drive['amplitude']['L5_pyramidal'].value = 10
+
+    # Check that you can't add more than one tonic
+    gui.add_drive_button.click()
```
ntolley, line 767: Is the current behavior that the previous tonic drive is overwritten?

ntolley:
We can use linkcheck ignore if this issue persists, not sure what changed in the runner:
https://github.com/sphinx-doc/sphinx/issues/5051

hnn_core/tests/test_gui.py
gtdang: Current behavior is that you cannot add more than one tonic. If you click the button again nothing will happen. It won't overwrite the previous tonic widget. There should probably be a console message though...

hnn_core/gui/gui.py
gtdang: Usually good practice to match on a standard case if it's something that can be variable. In this case the title of the drive can be passed in from an input file. That said...this if-condition should be using the drive _type_ not the drive _name_, as the _type_ is a standard set that's not variable like _title_.

hnn_core/gui/gui.py
gtdang: I think the difference is the use and update of the default_data dictionary. Previously we were using a small single-level default dict containing the amplitude. And it was updated in-place in the for-loop if data was passed in for that cell-type. Because the values were pointing to the same part of the default_data dict, an update would change the other widget values, thus necessitating the deep copy. 

The refactored code constructs a nested default_data dict cell-type as keys. Because the structure has a unique amplitude for each cell-type that doesn't get updated, there isn't a need for the deep copy.

gtdang:
> We can use linkcheck ignore if this issue persists, not sure what changed in the runner: [sphinx-doc/sphinx#5051](https://github.com/sphinx-doc/sphinx/issues/5051)

I think it could be similar to this issue. https://github.com/sphinx-doc/sphinx/issues/10343
For some reason the website doesn't accept whatever `user-agent` identity the runner is using. The user-agent is an identifier for the browser so websites can adjust to different browsers/applications.

hnn_core/gui/gui.py
gtdang: I rebased and resolved conflicts.

ntolley (approved):
@gtdang just tested this locally and everything is running as expected

The only change I think is necessary is renaming the drive title from `Tonic0` to `Tonic`. Other than that happy to merge!

image

gtdang:
The tonics currently follow the same numbering style as the drives when you add them. A value is appended based on the number of drives already added. I think this is just to make sure that the drives are uniquely named, since there's no way to create custom names for added drives. I think there was discussion at some point of eventually allowing multiple tonic biases, in which case this numbering would be necessary. 

Screenshot 2024-08-12 at 2 38 22 PM

PR #854, opened 2024-08-08
Fix linkcheck

gtdang:
Linkcheck started failing because of a certain link this week. The link works locally on my machine but always fails on the GitHub runner.
```
(auto_examples/workflows/plot_simulate_beta: line  330) broken    https://doi.org/10.1101/2021.04.16.440210 - 403 Client Error: Forbidden for url: https://www.biorxiv.org/lookup/doi/10.1101/2021.04.16.440210
```

I attempted different user-agent headers to see if I could get past the failing URL. Nothing worked and so I accepted defeat and added the link to the ignore list. 

Other changes: 
* Updated the runner to use py3.11. There isn't a reason to have it as an older and soon to be unsupported py3.8. 
* Removed an outdated function from api.rst that was giving an error message


ntolley (approved):
Great work @gtdang !!

PR #855, opened 2024-08-08
[MRG] Update GUI color

ntolley:
As discussed in #767 we are going to use the following theme for the GUI:

image

Also @gtdang I added `dt` and `tstop` as private attributes to the network object. This will allow us to override the default `dt` and `tstop` to saved values when we load a network back in.


gtdang (approved):
Looks nice!

PR #857, opened 2024-08-10
ENH: Refactor `BatchSimulate` Example and Improve Documentation

samadpls:
Adjusted parameter descriptions and added more informative explanations for waveforms and simulation outputs.

### DIFF
--- a/examples/howto/plot_batch_simulate.py
+++ b/examples/howto/plot_batch_simulate.py
@@ -116,8 +116,7 @@ def summary_func(results):
 net = jones_2009_model(mesh_shape=(3, 3))
 batch_simulation = BatchSimulate(net=net,
                                  set_params=set_params,
-                                 summary_func=summary_func,
-                                 n_trials=10)
+                                 summary_func=summary_func)
 simulation_results = batch_simulation.run(param_grid,
                                           n_jobs=n_jobs,
                                           combinations=False,
--- a/examples/howto/plot_batch_simulate.py
+++ b/examples/howto/plot_batch_simulate.py
@@ -107,10 +107,6 @@ def summary_func(results):
 ###############################################################################
 # Run the batch simulation and collect the results.
 
-# Uncomment this code if dask and distributed Python packages are installed.
-# from dask.distributed import Client
-# client = Client(threads_per_worker=1, n_workers=5, processes=False)
-
 
 # Initialize the network model and run the batch simulation.
 net = jones_2009_model(mesh_shape=(3, 3))
@@ -121,7 +117,7 @@ simulation_results = batch_simulation.run(param_grid,
                                           n_jobs=n_jobs,
                                           combinations=False,
                                           backend='loky')
-# backend='dask' if installed
+
 print("Simulation results:", simulation_results)
 ###############################################################################
 # This plot shows an overlay of all smoothed dipole waveforms from the
--- a/examples/howto/plot_batch_simulate.py
+++ b/examples/howto/plot_batch_simulate.py
@@ -24,7 +24,7 @@ from hnn_core.batch_simulate import BatchSimulate
 from hnn_core import jones_2009_model
 
 # The number of cores may need modifying depending on your current machine.
-n_jobs = 10
+n_jobs = 4
 ###############################################################################
 # The `add_evoked_drive` function simulates external input to the network,
 # mimicking sensory stimulation or other external events.
@@ -116,7 +116,8 @@ def summary_func(results):
 net = jones_2009_model(mesh_shape=(3, 3))
 batch_simulation = BatchSimulate(net=net,
                                  set_params=set_params,
-                                 summary_func=summary_func)
+                                 summary_func=summary_func,
+                                 n_trials=10)
 simulation_results = batch_simulation.run(param_grid,
                                           n_jobs=n_jobs,
                                           combinations=False,
--- a/examples/howto/plot_batch_simulate.py
+++ b/examples/howto/plot_batch_simulate.py
@@ -26,11 +26,24 @@ from hnn_core import jones_2009_model
 # The number of cores may need modifying depending on your current machine.
 n_jobs = 10
 ###############################################################################
+# The `add_evoked_drive` function simulates external input to the network,
+# mimicking sensory stimulation or other external events.
+#
+# - `evprox` indicates a proximal drive, targeting dendrites near the cell
+#   bodies.
+# - `mu=40` and `sigma=5` define the timing (mean and spread) of the input.
+# - `numspikes=1` means it's a single, brief stimulation.
+# - `weights_ampa` and `synaptic_delays` control the strength and
+#   timing of the input.
+#
+# This evoked drive causes the initial positive deflection in the dipole
+# signal, triggering a cascade of activity through the network and
+# resulting in the complex waveforms observed.
 
 
 def set_params(param_values, net=None):
     """
-    Set parameters in the network drives.
+    Set parameters for the network drives.
 
     Parameters
     ----------
@@ -57,16 +70,16 @@ def set_params(param_values, net=None):
                          synaptic_delays=synaptic_delays)
 
 ###############################################################################
-# Define a parameter grid for the batch simulation.
+# Next, we define a parameter grid for the batch simulation.
 
 
 param_grid = {
-    'weight_basket': np.logspace(-4 - 1, 10),
+    'weight_basket': np.logspace(-4, -1, 10),
     'weight_pyr': np.logspace(-4, -1, 10)
 }
 
 ###############################################################################
-# Define a function to calculate summary statistics
+# We then define a function to calculate summary statistics.
 
 
 def summary_func(results):
@@ -95,12 +108,12 @@ def summary_func(results):
 ###############################################################################
 # Run the batch simulation and collect the results.
 
-# Comment off this code, if dask and distributed Python packages are installed
+# Uncomment this code if dask and distributed Python packages are installed.
 # from dask.distributed import Client
 # client = Client(threads_per_worker=1, n_workers=5, processes=False)
 
 
-# Run the batch simulation and collect the results.
+# Initialize the network model and run the batch simulation.
 net = jones_2009_model(mesh_shape=(3, 3))
 batch_simulation = BatchSimulate(net=net,
                                  set_params=set_params,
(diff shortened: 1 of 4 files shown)


hnn_core/batch_simulate.py
```diff
@@ -97,9 +81,23 @@ def __init__(self, set_params, net=jones_2009_model(), tstop=170,
             If True, save calcium concentrations.
             Note, `save_outputs` must be True.
             Default: False.
+        record_vsec : 'all' | 'soma' | False
```
ntolley, line 84: While these do get checked when `simulate_dipole()` is run, I think it'd be worth adding a `check_option()` test here because you won't know that this value is incorrect until `batch_simulate.run()` is called

hnn_core/hnn_io.py
```diff
@@ -388,12 +388,12 @@ def dict_to_network(net_data,
 
     Parameters
     ----------
-    fname : str or Path
-        Path to configuration file
-    read_drives : bool
-        Read-in drives to Network object
-    read_external_biases
-        Read-in external biases to Network object
+    net_data : dict
```
ntolley, line 391: good catch!

hnn_core/batch_simulate.py
```diff
@@ -40,31 +41,15 @@ def __init__(self, set_params, net=jones_2009_model(), tstop=170,
             The network model to use for simulations. Must be an instance of
             jones_2009_model, law_2021_model, or calcium_model.
             Default is jones_2009_model().
+        net_json : str, optional
```
ntolley, line 44: Since a `Network` is already being passes in during the init, it might be easier to have users create the network with a file outside of `BatchSimulate`, and then pass it to the init.

The comment I made was mainly referring to whether or not it is worth exporting these `net_json` files after `batch_simulate.run()` is called. There's a lot of redundant information in those files so it's definitely not worth saving 100,000 copies of the same network that changes by a handful of parameters...

However, there should be some metadata that helps people understand what the exported parameter values mean (it's not clear to me how to best achieve that right now)

hnn_core/batch_simulate.py
samadpls: Thanks for the suggestion!  `check_option()` is already in place, but I’ve added a unit test for validation to catch any issues earlier. 😉

hnn_core/batch_simulate.py
```diff
@@ -41,9 +39,6 @@ def __init__(self, set_params, net=jones_2009_model(), net_json=None,
             The network model to use for simulations. Must be an instance of
```
ntolley, line 39:
```suggestion
            The network model to use for simulations. Examples include
```

examples/howto/plot_batch_simulate.py
```diff
@@ -111,18 +124,47 @@ def summary_func(results):
...
         dpl_smooth = data['dpl'][0].copy().smooth(window_len=30)
         dpl_waveforms.append(dpl_smooth.data['agg'])
+        param_values.append(data['param_values']['weight_basket'])
 
 plt.figure(figsize=(10, 6))
-for waveform in dpl_waveforms:
-    plt.plot(waveform, alpha=0.5, linewidth=3)
+cmap = plt.get_cmap('viridis')
+param_values = np.array(param_values)
+norm = plt.Normalize(param_values.min(), param_values.max())
```
ntolley, line 163: I would mess with the plotting here, it seems like the log scale is throwing off the colormap (most of the colors are dark)

examples/howto/plot_batch_simulate.py
```diff
@@ -111,18 +124,47 @@ def summary_func(results):
...
-# allowing us to visualize the range of responses across the parameter space.
-
-dpl_waveforms = []
+# batch simulation. Each line represents a different set of synaptic strength
+# parameters (`weight_basket`), allowing us to visualize the range of responses
+# across the parameter space.
+# The colormap represents different synaptic strengths, with purple indicating
+# lower strengths and yellow indicating higher strengths.
+#
+# Key observations:
```
ntolley, line 133: This is a great description! But it might be a little too verbose for this notebook. I would recomend cutting down everything below here to 2 sentences that describe the progression of dipole responses as drive strength increases.

examples/howto/plot_batch_simulate.py
```diff
@@ -111,18 +124,47 @@ def summary_func(results):
...
+#
+# - The dipole signal reflects the net current flow in the cortical column.
+# - Initially, we see a positive deflection as excitatory input arrives at
```
ntolley, line 136: Like replace this sentence to describe how weak vs strong drives impact the simulation as a whole

ntolley:
@samadpls I tried testing `BatchSimulate` with trial based parallelism (as opposed to MPI). Locally I'm seeing it run each trial sequence, can you confirm if this is happening on your computer?

samadpls:
Yes, sorry for the delay. It looks like Joblib wasn’t being used for `simulate_dipole` before. I’ve added it now, and it’s running in parallel properly. It should work for you too!

asoplata:
Ideally we should add significantly more testing, and test all batch simulation backends. In particular, both `multiprocessing` and `dask` are not tested as backends for batch simulation, but that was true before this issue. However, AFAIK I believe we need to merge this before the Steering committee meeting next week, and in its current form it seems to work.

asoplata:
Re: batch simulations in general: I am no longer doing simulations for my own research projects, but when I did, I relied strongly on parameter-sweeping Batch style simulations. This was essential for mapping out the parameter space landscapes, especially around behavior regime changes, and investigating unexpected behavior regimes, or using it to identify what channel might be more important than expected to a particular behavior, etc. That said, I am not expecting to be a significant user of the current `BatchSimulation` myself, since I'm more on the software-itself side of things.

Re: our particular `BatchSimulation`: I agree that it needs a refactor, but I think that's outside the scope of this PR. There are too many simulation-scope arguments, such as saving various data output, but these are all valid options that I can understand a user would want. Furthermore, many of these options are shared with `simulate_dipole`, and many of the ones not shared are also valid for the latter. The number of simulation-specific settings (and experience) suggests to me that we need to switch to a "simulation-config" style of setting all of these, where they are packaged in a dictionary. I have written a little bit on how we may want to start thinking about separating things like Network-config, Network-instantiated-object, simulation-config, etc. in here https://www.notion.so/jonescompneurolab/REF-Re-architect-configurations-vs-object-storage-f83530861109473ba036e5c211d9998c 

All that said, I think this PR is ready for merge on its own accord.

jasmainak:
@asoplata agree that the refactor is outside the scope of this PR. If you can open an issue on github, we can discuss what's the optimal way to implement it ... notion page is not accessible to those who are not Jones lab members.

you should begin by identifying a "developer user" who will provide feedback, otherwise the development is happening in a vacuum. One way to do this would be to incorporate the new class in @ntolley's SBI repo. When you incorporate it there, you will hit roadblocks and the fixes can be made upstream to the hnn-core repo. If the class is not being used by the developers of HNN, can we expect it to be used by the general user? :)

asoplata:
We had some good discussion on refactoring `BatchSimulation`'s API and handling simulation-configuration options at today's HNN Dev meeting. I will make a separate Issue here on GH for discussion about changing configuration APIs, but for larger issues like that, I prefer to wait to post the issue until I've got a proposal that I've though through. I personally plan to focus more on refactors in the spring, once we've gotten through more outstanding PRs / issues. Good idea on the developer-user, I will try to incorporate that into our workflows

ntolley:
@jasmainak truthfully a lot of my guidance on this PR came from simplifying the batch simulate code I implemented in the SBI repo (minus the API checks). So in terms of me actually using it, I've been using a worse verion :smile: 

With your example I feel it's a little misleading, the code you write to update the parameters will be specific to every batch simulate class, and the run function which saves outputs will also be specific depending on the outputs you want to save. After working for a few years on these sorts of problem, I feel I've written the same code blocks to save different kinds of output, and update different parameters, many many times. In that sense the current class is a huge improvement for my own work.

ntolley:
^not to say that we shouldn't iterate on the API, just wanted to give some background on how this fits with an actual project dependent on batch simulation

ntolley:
@samadpls apologies for the slow review on this, but I'm pretty sure this is still running serially:
image

It seems like it's spinning up an instance with 10 workers, but only completing a single task before moving to the next one

hnn_core/batch_simulate.py
```diff
@@ -301,14 +297,15 @@ def _run_single_sim(self, param_values):
...
         if self.save_dpl:
-            dpl = simulate_dipole(net,
-                                  tstop=self.tstop,
-                                  dt=self.dt,
-                                  n_trials=self.n_trials,
-                                  record_vsec=self.record_vsec,
-                                  record_isec=self.record_isec,
-                                  postproc=self.postproc)
-            results['dpl'] = dpl
+            with JoblibBackend(n_jobs=n_jobs):
```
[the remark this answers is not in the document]
ntolley: Yeah I don't think this piece is necessary, you really only need it to parallelize over trials

ntolley:
@samadpls just opened a PR on your branch: https://github.com/samadpls/hnn-core/pull/5

I think the issue was just indexing on the start/stop positions of the batch weren't being calculated correctly. @samadpls @asoplata @dylansdaniels any ideas on how to test that batches are actually being run in parallel and not serially? Doesn't need to be super fancy as we should get this PR merged

asoplata:
> @samadpls just opened a PR on your branch: [samadpls#5](https://github.com/samadpls/hnn-core/pull/5)
> 
> I think the issue was just indexing on the start/stop positions of the batch weren't being calculated correctly. @samadpls @asoplata @dylansdaniels any ideas on how to test that batches are actually being run in parallel and not serially? Doesn't need to be super fancy as we should get this PR merged

I've been going through the Joblib docs, and unfortunately I'm not sure there's a trivial way to measure the number of active workers (or total workers at all) inside a simple `Parallel` call like we use in `BatchSimulate.simulate_batch`. For testing multiple simultaneous workers, we may have to wrap our `simulate_batch` call inside a block specific to each backend, such as https://github.com/joblib/loky?tab=readme-ov-file#usage . I'm assuming that, in that example, Loky would constrain the `Parallel` instantiation, but I'm not sure. It may actually be easier to mock `simulate_batch` and test that, instead of actually calling it.

samadpls:
> @samadpls just opened a PR on your branch: [samadpls#5](https://github.com/samadpls/hnn-core/pull/5)
> 
> I think the issue was just indexing on the start/stop positions of the batch weren't being calculated correctly. @samadpls @asoplata @dylansdaniels any ideas on how to test that batches are actually being run in parallel and not serially? Doesn't need to be super fancy as we should get this PR merged

I have added timing around the start and end of `simulate_batch` to track execution duration, which should help us understand parallelism better. Let me know if you think we should take a different approach or add more specific checks.

asoplata:
Thanks for reverting the joblib addition in `_run_single_sim` @samadpls and all your work with the debugging and testing! Since this PR is mostly about the example Howto and not about the `BatchSimulation` config API or its tests, I've made a separate issue here #955 for refactoring `BatchSimulation`'s tests (I have not yet made an issue for the config API). Once the "dust has settled" with the commits and we've taken time to manually test this, it will probably be ready for merge soon.

examples/howto/plot_batch_simulate.py
```diff
@@ -95,36 +107,59 @@ def summary_func(results):
...
 
 
-# Run the batch simulation and collect the results.
+# Initialize the network model and run the batch simulation.
 net = jones_2009_model(mesh_shape=(3, 3))
 batch_simulation = BatchSimulate(net=net,
                                  set_params=set_params,
-                                 summary_func=summary_func)
+                                 summary_func=summary_func,
+                                 n_trials=10)
```
ntolley, line 120:
```suggestion
                                 n_trials=1)
```

ntolley, line 120:
```suggestion
                                 summary_func=summary_func)
```

hnn_core/batch_simulate.py
```diff
@@ -16,15 +16,14 @@
 
 
 class BatchSimulate(object):
-    def __init__(self, set_params, net=jones_2009_model(), tstop=170,
-                 dt=0.025, n_trials=1, record_vsec=False,
-                 record_isec=False, postproc=False, save_outputs=False,
+    def __init__(self, set_params, net=None,
```
asoplata, line 19: For the default `net` object here, is there a specific reason why you changed Lines 19, 109, and 132 (corresponding to the later version of `batch_simulate.py`)? As far as I tested, both versions work the same: they use `jones_2009_model()` if nothing is passed, but they successfully use a passed `Network` if the user passes a different one. If there is no use-case where the changes to `net` in Lines 19, 109, and 132  are required, then would you please revert them to the original? The original is cleaner.

examples/howto/plot_batch_simulate.py
```diff
@@ -95,36 +107,58 @@ def summary_func(results):
 ###############################################################################
 # Run the batch simulation and collect the results.
 
-# Comment off this code, if dask and distributed Python packages are installed
+# Uncomment this code if dask and distributed Python packages are installed.
 # from dask.distributed import Client
 # client = Client(threads_per_worker=1, n_workers=5, processes=False)
```
asoplata, line 112: Concerning all Dask lines (in the later version, Lines 110-112 and 124 of `plot_batch_simulate.py`): After installing the packages, uncommenting these lines, and changing the `backend`, I could not get this to work at all. On both Macos on Apple Silicon and Linux on Intel chips via Oscar, I could not get the `dask` usage case to work. I tested different numbers of workers and some options, but in every case I either got NEURON errors about actual simulations or just segmentation faults. Also, we have no tests for the `dask` cases at all. @ntolley @samadpls Do y'all use, or are you aware of anyone who uses, the `dask` backends for batch simulation, or at all?

I think that, for now, we should remove any mention of `dask` from this Example. If we want to truly support `dask`, then there is a good amount of work that needs to be done (illustrating that it works, writing tests for it including across platforms, justification to use it in place of JobLib, etc.).

hnn_core/batch_simulate.py
```diff
@@ -201,16 +196,13 @@ def run(self, param_grid, return_output=True,
...
         batch_size = min(self.batch_size, total_sims)
 
         results = []
         simulated_data = []
-        for i in range(batch_size):
-            start_idx = i * num_sims_per_batch
-            end_idx = start_idx + num_sims_per_batch
-            if i == batch_size - 1:
-                end_idx = len(param_combinations)
+        for i in range(0, total_sims, batch_size):
```
asoplata, line 203: I like this, it is clearer

asoplata (changes requested):
I did some testing with this and I think it needs only some very minor changes. Once these are done, I would be willing to merge it. Thanks @samadpls !

asoplata (approved):
Retested with new changes, and it's ready to go!

PR #861, opened 2024-08-14
[MRG] GUI n_drive_cells widget

### LINKED ISSUE #850: GUI: Network configuration different after GUI initialization

gtdang:
**Needs #853 merged first**
Commits for this work start at [4789e9e](https://github.com/jonescompneurolab/hnn-core/pull/861/commits/4789e9e6c7d475749f2c0cf4c010a7e0a2aaae25). Previous are in #853. 

Adds "number of drive cells" to the drive widgets. 

Screenshot 2024-08-14 at 10 06 04 AM

Work done:
1. Created widget for n_drive_cells parameter
2. Changed the Synchronous inputs checkbox to Cell Specific to align with the API keyword 
3. Added observer to disable the n_dive_cells widget if the cell specific checkbox is checked 

closes #850


hnn_core/gui/gui.py
```diff
@@ -1048,21 +1048,32 @@ def _get_cell_specific_widgets(layout, style, location, data=None):
...
+                         weights_nmda=None, delays=None,
+                         n_drive_cells=None, cell_specific=None
+                         ):
     default_data = {
         'tstart': 0.,
         'tstart_std': 0.,
         'tstop': tstop_widget.value,
         'burst_rate': 7.5,
         'burst_std': 0,
-        'repeats': 1,
```
ntolley, line 1061: any reason to remove `repeats`?

hnn_core/gui/gui.py
```diff
@@ -1048,21 +1048,32 @@ def _get_cell_specific_widgets(layout, style, location, data=None):
...
+        'numspikes': 1,
+        'n_drive_cells': 1,
+        'cell_specific': False,
         'seedcore': 14,
     }
-    if isinstance(data, dict):
-        default_data = _update_nested_dict(default_data, data)
+    data.update({'n_drive_cells': n_drive_cells,
```
ntolley, line 1074: Why not define these directly in `default_data`?

hnn_core/gui/gui.py
gtdang: I changed this to 'numspikes' to match the api for drive creation.

hnn_core/tests/test_gui.py
```diff
@@ -551,22 +551,58 @@ def test_gui_synchronous_inputs(setup_gui):
     """Test if the GUI creates plot using synchronous_inputs."""
     gui = setup_gui
 
+    # Set cell_specific to False
+    gui.drive_widgets[0]['is_cell_specific'].value = False
+    # Check that the n_drive_cells is not disabled
+    assert not gui.drive_widgets[0]['n_drive_cells'].disabled
+
+    # set synch inputs to first driver in simulation
```
ntolley, line 559:
```suggestion
    # set sync inputs to first driver in simulation
```

PR #868, opened 2024-08-21
[MRG] fix GUI MPI test

gtdang:
Updated `test_gui_run_simulation_mpi`.  The test was not actually testing if any dipoles were constructed. 

* Added assert that the dipole list is not empty 
* Updated the function to use the default network instead of gui test fixture (smaller mesh grid) because it was not working with MPI.  Presumably because the smaller network used in the fixture was not compatible with the parameterization of multiprocessing subprocess call.
    *  We were encountering an error at the `h.finitialize` step of `network_builder`>`_simulate_single_trial` when using the fixture network. The error was within neuron: "NEURON: mindelay is 0 (or less than dt for fixed step method)".


gtdang:
@dylansdaniels @ntolley @jasmainak 
Can you review and merge this?

dylansdaniels (approved):
Thanks George, this looks good. And good to know about the issue with the test fixture / MPI as well.

PR #880, opened 2024-09-06
build: update setup.py to remove deprecated distutils import

### LINKED ISSUE #879: build: distutils depreciation
We currently use `distutils` in our setup.py. It has been [deprecated and supposedly removed in python 3.12](https://docs.python.org/3.10/whatsnew/3.10.html#distutils). This might be why Neuron didn't create the mod files for this user https://github.com/jonescompneurolab/hnn-core/discussions/835 - they were using 3.12. 

`setuptools` has classes to replace the distutils classes we are using.

gtdang:
Replaces distutils imports with corollary setuptools classes because distutils is deprecated and supposedly removed from python 3.12.

closes #879

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,10 +5,8 @@ import os
 import subprocess
 import shutil
 
-from setuptools import setup, find_packages
-
-from distutils.command.build_py import build_py
-from distutils.cmd import Command
+from setuptools import setup, find_packages, Command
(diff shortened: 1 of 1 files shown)


jasmainak (approved):
LGTM ! @ntolley please merge if you're happy

ntolley:
Great work @gtdang!! Merging now

PR #890, opened 2024-09-17
[MRG] GUI: delete single drive button

### LINKED ISSUE #887: GUI: Drive deletion
Add a feature to delete a single drive.

gtdang:
Added a button to delete a single drive.

optimized

I decided to refactor previous functions that changed the GUI's state as class methods. The key challenge I faced was how to close a parent object based a nested child delete button. Keeping the external functions would have required creating additional logic and data structures to keep track of drive indices upon creation and editing.  As an added benefit I found that the refactor made the method calls more straight-forward and easier to understand. 

closes #887

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -926,96 +926,97 @@ class HNNGUI:
             return
 
         style = {'description_width': '125px'}
-        self._drives_out.clear_output()
         if not prespecified_drive_data:
             prespecified_drive_data = {}
         prespecified_drive_data.update({"seedcore": max(event_seed, 2)})
 
-        with self._drives_out:
-            if not prespecified_drive_name:
-                name = drive_type + str(len(self.drive_boxes))
-            else:
-                name = prespecified_drive_name
-            if drive_type in ('Rhythmic', 'Bursty'):
-                drive, drive_box = _get_rhythmic_widget(
-                    name,
-                    self.widget_tstop,
-                    layout,
-                    style,
-                    location,
-                    data=prespecified_drive_data,
-                    weights_ampa=prespecified_weights_ampa,
-                    weights_nmda=prespecified_weights_nmda,
-                    delays=prespecified_delays,
-                    n_drive_cells=prespecified_n_drive_cells,
-                    cell_specific=prespecified_cell_specific,
-                )
-            elif drive_type == 'Poisson':
-                drive, drive_box = _get_poisson_widget(
-                    name,
-                    self.widget_tstop,
-                    layout,
-                    style,
-                    location,
-                    data=prespecified_drive_data,
-                    weights_ampa=prespecified_weights_ampa,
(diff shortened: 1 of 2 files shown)


hnn_core/gui/gui.py
```diff
@@ -417,6 +417,7 @@ def __init__(self, theme_color="#802989",
         # Add drive section
         self.drive_widgets = list()
         self.drive_boxes = list()
+        self.drive_accordion = Accordion()
```
ntolley, line 420: In the old code, was this information not stored persistently meaning you couldn't access the parent tree structure for deleting the drives?

hnn_core/tests/test_gui.py
```diff
@@ -1109,3 +1109,12 @@ def test_update_nested_dict():
...
     assert updated == expected
+
+
+def test_delete_single_drive(setup_gui):
+    """Deleting a single drive."""
+    gui = setup_gui
+    assert len(gui.drive_accordion.children) == 6
+
+    gui._simulate_delete_single_drive(0)
+    assert len(gui.drive_accordion.children) == 5
```
ntolley, line 1120: possible to check that the correct drive was deleted?

hnn_core/gui/gui.py
```diff
@@ -532,11 +533,11 @@ def _handle_backend_change(backend_type):
...
         def _add_drive_button_clicked(b):
-            return add_drive_widget(self.widget_drive_type_selection.value,
-                                    self.drive_boxes, self.drive_widgets,
-                                    self._drives_out, self.widget_tstop,
-                                    self.widget_location_selection.value,
-                                    layout=self.layout['drive_textbox'])
+            return self.add_drive_widget(
+                self.widget_drive_type_selection.value,
+                self.widget_location_selection.value,
+                layout=self.layout['drive_textbox']
```
ntolley, line 539: able to remove and call directly in side the function?

hnn_core/tests/test_gui.py
gtdang: Updated this to also have explicit checks for titles. Turns out titles weren't being updated correctly, so thanks for bringing it up!

hnn_core/gui/gui.py
gtdang: That's correct. It wasn't being stored. It was just being generated and displayed in the output. That happens here:
https://github.com/jonescompneurolab/hnn-core/blob/4abfa6f7da1eefff3af1d14a8be6d8103614f3db/hnn_core/gui/gui.py#L1467-L1480

Previously the only action we needed to do was add drives or delete them all. With add drive, the Accordion would just be rebuilt with the new drive and redisplayed in the output. With the delete function we need to access the Accordion to determine the selected accordion to close, which necessitated saving the Accordion to the class.

ntolley (approved):
Look great @gtdang!!

Going to mess with this locally a bit but will merge soon

PR #892, opened 2024-09-18
[MRG] GUI Dipole load data figure scaling

gtdang:
The plot generated when loading data had the same default scaling and smoothing that is applied to simulation data. This PR removes smoothing and scaling so the plot respects the raw loaded data. The user can then choose how to adjust the scaling and smoothing as they see fit. 

Screenshot 2024-09-18 at 3 16 36 PM


ntolley (approved):
Looks great @gtdang !!

PR #893, opened 2024-09-18
[MRG] feat: post-processing function to adjust figures 

gtdang:
Added a post-processing function to adjust figures after they have been constructed

This was created for a special case where L2 and L5 dipole plots should have the same axis limits. There wasn't a simple way to set axis limits upon plot creation so the post-process implementation seemed like the most simple solution while keeping the current APIs intact.

Admittedly this solution is a bit hacky as it renders the first set a plots and quickly re-renders the adjustment (looks like a flicker). Perhaps adding keywords to not render certain plot calls would be a solution. 

Screenshot 2024-09-18 at 5 08 55 PM

### DIFF
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -767,6 +767,28 @@ def _add_figure(b, widgets, data, template_type, scale=0.95, dpi=96):
     data['fig_idx']['idx'] += 1
 
 
+def _postprocess_template(template_name, fig, idx,
+                          use_ipympl=True, widgets=None):
+    """Post-processes and re-renders plot templates with determined styles
+
+    Templates are constructed on panel-by-panel basis. If adjustments need to
+    be made based on information other plots in the figure, it is adjusted with
+    this function. For example, L2 and L5 dipole plots should have the same
+    y-axis range.
+    """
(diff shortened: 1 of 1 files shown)


hnn_core/gui/_viz_manager.py
```diff
@@ -958,6 +983,15 @@ def add_figure(self, b=None):
                 # paint fig in axis
                 self._simulate_edit_figure(fig_name, ax_name, sim_name,
                                            plot_type, {}, "plot")
+            # template post-processing
+            fig_key = self.data['fig_idx']['idx'] - 1
+            _postprocess_template(template_name,
+                                  fig=self.figs[fig_key],
```
gtdang, line 989: @ntolley 
This was the PR I was showing you that was glitching and preventing further figure creation after deleting figures a few times. The issue was at this line where I was previously indexing `self.figs` with a the length of itself, because I assumed it was a list and I wanted the latest created figure. Turns out that`self.figs` is a dict with ints as keys and the length of the dict doesn't correspond to the keys after you delete from and add to it.  Using the int stored in `self.data['fig_idx']['idx']` as the key fixed the issue.

ntolley:
@gtdang are we sure all the plots are being iterated over to find ymin/ymax? The 3rd plot seems to have different ylims here:
image

PR #899, opened 2024-09-26
[MRG] GUI fix minor glitches

gtdang:
1. Removed tight_layout method in favor of the constrained layout argument
    * The tight layout was displaying a warning for spectrogram plots. This change removes that message.

2. fixed set selected index for figure controls after new figure is generated
    * When creating a blank figure, the GUI was not automatically opening/selecting the new figure tab. This was because the selected index was only being set for the figure tabs but not the figure control tabs. This made them not synchronous, and when the two tab groups are later linked, it reverted to the un-updated index.

### DIFF
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -763,9 +763,8 @@ def _add_figure(b, widgets, data, template_type, scale=0.95, dpi=96):
         with plt.ioff():
             fig = plt.figure(figsize=figsize, dpi=dpi, layout='constrained')
             axd = fig.subplot_mosaic(mosaic, **kwargs)
-
-        fig.canvas.header_visible = False
-        fig.canvas.footer_visible = False
+            fig.canvas.header_visible = False
+            fig.canvas.footer_visible = False
 
         if data['use_ipympl'] is False:
             plt.show()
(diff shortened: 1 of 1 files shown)


hnn_core/gui/_viz_manager.py
```diff
@@ -762,13 +760,10 @@ def _add_figure(b, widgets, data, template_type, scale=0.95, dpi=96):
...
         mosaic = template_type['mosaic']
         kwargs = template_type['kwargs']
-        plt.ioff()
-        fig, axd = plt.subplot_mosaic(mosaic,
-                                      figsize=figsize,
-                                      dpi=dpi,
-                                      **kwargs)
-        plt.ion()
-        fig.tight_layout()
+        with plt.ioff():
```
ntolley, line 763: what's the significance of this context manager?

gtdang: It turns off interactive mode. With interactive on the plot may render each time a change is made. Switching it off will construct the figure object but won't display/render it until a plt.show() is called. 

That said... the fig.canvas commands after this context should be within the context manager here....

ntolley:
Everything looks good on me end, thanks @gtdang!!

ntolley (approved):

PR #902, opened 2024-10-09
[MRG] Update Network eq method

gtdang:
The eq method was refactored to be more flexible to new methods added to the object. 

The eq method loops over class attributes and compares the values to assess equality. 
* Previously we used the dir() function to list all of the classes attributes and methods and then filtered for only the attributes.
    * We used a hard-coded list of attributes and methods we wanted to pass over. This was annoying because each new method added to the Network class had to be manually added to this list. 
* This change uses the var() function to grab just the list of attributes from the class so new methods don't need to be coded manually. 
* The change was also made to the test_gui.py>check_equal_networks function, which had logic adapted from the eq method.

### DIFF
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -465,20 +465,16 @@ class Network:
             return False
 
         # Check all other attributes
-        all_attrs = dir(self)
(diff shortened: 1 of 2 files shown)


ntolley (approved):
Looks good to me!

PR #913, opened 2024-10-18
[MRG] fix synapses properties rendering

### LINKED ISSUE #911: GUI BUG: Synapse radio button does not do anything
Very easy to reproduce, just click the button!

@gtdang any thoughts on this? I know @kmilo9999 coded these parameters so hopefully this is an easy fix

gtdang:
Changed logic in `_update_cell_params_vbox` to properly build the `cell_parameters_key`.

The problem was here: 
https://github.com/jonescompneurolab/hnn-core/blob/7a42650dbe41e0a740612fe374e5b19e3e8e8d7c/hnn_core/gui/gui.py#L1974-L1975

`if "Biophysics"` will always evaluate to `True`.

closes #911


asoplata:
Just double-checked this, yes this solves the issue. Merging.

Issue #910, opened 2024-10-18
GUI: Enhancements to improve tutorials


ntolley:
Going through the tutorials has revealed some modifications that would make using the GUI much smoother for users:

- [ ] Add global smoothing and scaling option (temporarily set default smoothing to zero (right now these are set manually for every new plot you make)
- [ ] Allow setting min/max freq on PSD
- [ ] Disable spectrogram specific parameters when other plots are selected
- [ ] Add E/I synaptic gains button (already implemented in #885)
- [ ] Add ability to set simulation data to `None` so that you can just load in and plot a data file

For discussion

- [ ] Disable simulation data in `Data to compare` box, you can plot multiple simulations by hitting the `Add plot` button multiple times

dylansdaniels:
See #914 for proposed PSD fix

dylansdaniels:
[] add observer to monitor change in tstop in Simulation tab to ensure simulation length can be correctly specified for rhythmic, poisson, and tonic external drives

gtdang:
#918

Issue #882, opened 2024-09-10
GUI: Min frequency input for spectrogram


gtdang:
Add a minimum frequency selector for spectrogram plots.

PR #914, opened 2024-10-18
Fix Freq in GUI PSD Plot

dylansdaniels:
Update PSD plot in gui _viz_manager so that min/max frequency are set… according to the plot config rather than being hard coded

See issue #910

### DIFF
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -252,8 +252,10 @@ def _update_ax(fig, ax, single_simulation, sim_name, plot_type, plot_config):
 
     elif plot_type == 'PSD':
         if len(dpls_copied) > 0:
+            min_f = plot_config['min_spectral_frequency']
+            max_f = plot_config['max_spectral_frequency']
             color = ax._get_lines.get_next_color()
-            dpls_copied[0].plot_psd(fmin=0, fmax=50, ax=ax, color=color,
+            dpls_copied[0].plot_psd(fmin=min_f, fmax=max_f, ax=ax, color=color,
                                     label=sim_name, show=False)
 
     elif plot_type == 'spectrogram':


hnn_core/gui/_viz_manager.py
```diff
@@ -252,8 +252,10 @@ def _update_ax(fig, ax, single_simulation, sim_name, plot_type, plot_config):
 
     elif plot_type == 'PSD':
         if len(dpls_copied) > 0:
+            min_f = plot_config['min_spectral_frequency']
+            max_f = plot_config['max_spectral_frequency']
             color = ax._get_lines.get_next_color()
-            dpls_copied[0].plot_psd(fmin=0, fmax=50, ax=ax, color=color,
+            dpls_copied[0].plot_psd(fmin=min_f, fmax=max_f, ax=ax, color=color,
```
ntolley, line 258: This is great! Can you checkout the spectrogram tests associated with #882 to verify that the PSD plot is rendered correctly?

gtdang:
Might consider renaming the widgets to remove the 'spectral' from the widget names. Before they were specific to the spectrogram plots, but with this change they are not.

asoplata (commented):
Tested it myself and LGTM. I'll note that for the PSD plot, changing the Min or Max Spectral Frequency then clicking "Add plot" does successfully update the view. However, for the spectrogram, the view is not updated. This may be a separate bug.

gtdang:
> Tested it myself and LGTM. I'll note that for the PSD plot, changing the Min or Max Spectral Frequency then clicking "Add plot" does successfully update the view. However, for the spectrogram, the view is not updated. This may be a separate bug.

I think it's over-plotting a new line when you click add-plot and then updating the axis. There's no over plotting functionality with the spectrogram, so you need to clear then "add plot" to re-plot the axis.

asoplata:
Yep, you're right, clearing the axis and then re-adding the plot successfully changes the frequency range for the spectrogram too.

Because this is needed for Thursday's gamma tutorial and I can't find problems with it, merging.

Issue #915, opened 2024-10-18
GUI: Add E/I synaptic gains button


asoplata:
Task 4 of https://github.com/jonescompneurolab/hnn-core/issues/910

`hnn_core/network:Network.update_weights()` ( 
https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/network.py#L1430 ) currently applies universally across all subsets of `{Excitatory, Inhibitory} -> {Excitatory, Inhibitory}`. If we assume that most users are going to want to *universally* increase the gain of a particular synapse connection, this means that both the `Network > Connectivity` or `Network > Cell parameters` VBoxes are not good places to add this. In the Connectivity case, a user would have to add the gain to every type of connection, which is annoying and time-consuming. In the Cell parameters case, we cannot select all Excitatory connections to update, which `update_weights()` currently requires.

For the sake of the tutorial, I think this should be solved in three steps:

1. Add a new VBox named `Network > Universal synaptic gain` with 4 adjustable values: the 4 combinations that `update_weights()` supports.
2. Later, either add another function or refactor `update_weights()` so that you can add a unitless gain modifier to connections the "lowest" level of the hierarchy. 
3. Then, add entries in the `Network > Connectivity` vbox to customize these per every connection, if the user wants: e.g. add a gain box to `L2_pyramidal -> L5_basket (soma)` and similar.
4. At this point, if we want, we could also add additional per-`cell_type` synaptic gain entries in `Network > Cell parameters` vbox

asoplata:
@dylansdaniels For the zero-ing out of the L2/3 synapses in the Gamma tutorial (near Fig 7 of https://jonescompneurolab.github.io/hnn-tutorials/gamma/gamma ), it may be fine for Thursday if the weights are changed manually. I think in the instance in the example, they would only need to change the outgoing L2/3 Pyramidal weights to 0, which is only about 15 clicks (shown below, they don't need to adjust the inhibitory weights).

![Screenshot_20241021_114540](https://github.com/user-attachments/assets/4a97a8f8-2f14-4362-aa15-2b8e450d1ea8)

I suggest this because the current `update_weights()` currently does not discriminate between L2 and L5, and will have to be changed before it could be used on Thursday, not to mention where it gets used and tested in the GUI would have to be implemented and tested, etc.

dylansdaniels:
[...]

PR #918, opened 2024-10-22
[MRG] GUI synaptic gains implementation

gtdang:
Added synaptic gains widgets to the GUI.

Changes:
1. Adjusted `Network` class methods 
    2. Renamed `update_weights` method to `set_synaptic_gains` for more clarity 
    3. Added `get_synaptic_gains` method and test
4. Added new "Synaptic gains" child tab under the Network tab with widgets for each weight 
5. Updated the `_init_network_from_widgets` function to update synaptic gains 
6. Added tests for GUI synaptic gains 

Screenshot 2024-10-22 at 2 27 42 PM

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -1739,10 +1739,13 @@ def add_cell_parameters_tab(cell_params_out, cell_pameters_vboxes,
 
 def add_synaptic_gain_tab(net, syn_gain_out, syn_gain_textfields, layout):
 
+    gain_values = net.get_synaptic_gains()
     gain_types = ('e_e', 'e_i', 'i_e', 'i_i')
     for gain_type in gain_types:
         gain_widget = BoundedFloatText(
-            value=1, description=f'{gain_type}', min=0, max=1e6, step=.1,
+            value=gain_values[gain_type],
+            description=f'{gain_type}',
+            min=0, max=1e6, step=.1,
             disabled=False, layout=layout)
         syn_gain_textfields[gain_type] = gain_widget
 
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -424,6 +424,8 @@ class HNNGUI:
 
         # Cell parameter list
         self.cell_pameters_widgets = dict()
+        # Cell parameter dict
+        self.cell_parameters_widgets = dict()
 
         self._init_ui_components()
         self.add_logging_window_logger()
@@ -570,6 +572,7 @@ class HNNGUI:
                 self._simulation_status_bar, self._simulation_status_contents,
                 self.connectivity_widgets, self.viz_manager,
                 self.simulation_list_widget, self.cell_pameters_widgets)
+                self.simulation_list_widget, self.cell_parameters_widgets,
 
         def _simulation_list_change(value):
             # Simulation Data
@@ -612,13 +615,13 @@ class HNNGUI:
 
         def _cell_type_radio_change(value):
             _update_cell_params_vbox(self._cell_params_out,
-                                     self.cell_pameters_widgets,
+                                     self.cell_parameters_widgets,
                                      value.new,
                                      self.cell_layer_radio_buttons.value)
 
         def _cell_layer_radio_change(value):
             _update_cell_params_vbox(self._cell_params_out,
-                                     self.cell_pameters_widgets,
+                                     self.cell_parameters_widgets,
                                      self.cell_type_radio_buttons.value,
                                      value.new)
 
@@ -688,8 +691,10 @@ class HNNGUI:
 
         connectivity_configuration.children = [connectivity_box,
                                                cell_parameters]
+                                               cell_parameters,
         connectivity_configuration.titles = ['Connectivity',
                                              'Cell parameters']
+                                             'Cell parameters',
 
         drive_selections = VBox([
             self.add_drive_button, self.widget_drive_type_selection,
@@ -902,7 +907,7 @@ class HNNGUI:
                                  self._connectivity_out,
                                  self.connectivity_widgets,
                                  self._cell_params_out,
-                                 self.cell_pameters_widgets,
+                                 self.cell_parameters_widgets,
                                  self.cell_layer_radio_buttons,
                                  self.cell_type_radio_buttons,
                                  self.layout)
@@ -1034,7 +1039,7 @@ class HNNGUI:
             if load_type == 'connectivity':
                 add_connectivity_tab(
                     params, self._connectivity_out, self.connectivity_widgets,
-                    self._cell_params_out, self.cell_pameters_widgets,
+                    self._cell_params_out, self.cell_parameters_widgets,
                     self.cell_layer_radio_buttons,
                     self.cell_type_radio_buttons, layout)
             elif load_type == 'drives':
@@ -1598,7 +1603,7 @@ def _build_drive_objects(drive_type, name, tstop_widget, layout, style,
 
 
 def add_connectivity_tab(params, connectivity_out, connectivity_textfields,
-                         cell_params_out, cell_pameters_vboxes,
+                         cell_params_out, cell_parameters_vboxes,
                          cell_layer_radio_button, cell_type_radio_button,
                          layout):
     """Add all possible connectivity boxes to connectivity tab."""
@@ -1609,7 +1614,7 @@ def add_connectivity_tab(params, connectivity_out, connectivity_textfields,
                                  connectivity_textfields)
 
     # build cell parameters tab
-    add_cell_parameters_tab(cell_params_out, cell_pameters_vboxes,
+    add_cell_parameters_tab(cell_params_out, cell_parameters_vboxes,
                             cell_layer_radio_button, cell_type_radio_button,
                             layout)
     return net
@@ -1819,7 +1824,6 @@ def _init_network_from_widgets(params, dt, tstop, single_simulation_data,
                     'nc_dict']['A_weight'] = vbox_key.children[1].value
 
     # Update cell params
-
     update_functions = {
         'L2 Geometry': _update_L2_geometry_cell_params,
         'L5 Geometry': _update_L5_geometry_cell_params,
@@ -1934,6 +1938,7 @@ def run_button_clicked(widget_simulation_name, log_out, drive_widgets,
                                    simulation_data[_sim_name], drive_widgets,
                                    connectivity_textfields,
                                    cell_pameters_widgets)
+                                   cell_parameters_widgets,
 
         print("start simulation")
         if backend_selection.value == "MPI":
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -422,10 +422,11 @@ class HNNGUI:
         # Connectivity list
         self.connectivity_widgets = list()
 
-        # Cell parameter list
-        self.cell_pameters_widgets = dict()
         # Cell parameter dict
         self.cell_parameters_widgets = dict()
+
+        # Synaptic Gains dict
+        self.synaptic_gain_widgets = dict()
 
         self._init_ui_components()
         self.add_logging_window_logger()
@@ -475,6 +476,7 @@ class HNNGUI:
         self._drives_out = Output()  # tab to add new drives
         self._connectivity_out = Output()  # tab to tune connectivity.
         self._cell_params_out = Output()
+        self._syn_gain_out = Output()
 
         self._log_out = Output()
 
@@ -571,8 +573,9 @@ class HNNGUI:
                 self.widget_mpi_cmd, self.widget_n_jobs, self.params,
                 self._simulation_status_bar, self._simulation_status_contents,
                 self.connectivity_widgets, self.viz_manager,
-                self.simulation_list_widget, self.cell_pameters_widgets)
                 self.simulation_list_widget, self.cell_parameters_widgets,
+                self.synaptic_gain_widgets
+            )
 
         def _simulation_list_change(value):
             # Simulation Data
@@ -689,12 +692,14 @@ class HNNGUI:
             self._cell_params_out
         ])
 
+        syn_gain = VBox([self._syn_gain_out])
+
         connectivity_configuration.children = [connectivity_box,
-                                               cell_parameters]
                                                cell_parameters,
+                                               syn_gain]
         connectivity_configuration.titles = ['Connectivity',
-                                             'Cell parameters']
                                              'Cell parameters',
+                                             'Synaptic gains']
 
         drive_selections = VBox([
             self.add_drive_button, self.widget_drive_type_selection,
@@ -910,6 +915,8 @@ class HNNGUI:
                                  self.cell_parameters_widgets,
                                  self.cell_layer_radio_buttons,
                                  self.cell_type_radio_buttons,
+                                 self._syn_gain_out,
+                                 self.synaptic_gain_widgets,
                                  self.layout)
 
             # Add drives
@@ -1041,7 +1048,9 @@ class HNNGUI:
                     params, self._connectivity_out, self.connectivity_widgets,
                     self._cell_params_out, self.cell_parameters_widgets,
                     self.cell_layer_radio_buttons,
-                    self.cell_type_radio_buttons, layout)
+                    self.cell_type_radio_buttons,
+                    self._syn_gain_out, self.synaptic_gain_widgets,
+                    layout)
             elif load_type == 'drives':
                 self.add_drive_tab(params)
             else:
@@ -1605,7 +1614,7 @@ def _build_drive_objects(drive_type, name, tstop_widget, layout, style,
 def add_connectivity_tab(params, connectivity_out, connectivity_textfields,
                          cell_params_out, cell_parameters_vboxes,
                          cell_layer_radio_button, cell_type_radio_button,
-                         layout):
+                         syn_gain_out, syn_gain_textfields, layout):
     """Add all possible connectivity boxes to connectivity tab."""
     net = dict_to_network(params)
 
@@ -1617,6 +1626,10 @@ def add_connectivity_tab(params, connectivity_out, connectivity_textfields,
     add_cell_parameters_tab(cell_params_out, cell_parameters_vboxes,
                             cell_layer_radio_button, cell_type_radio_button,
                             layout)
+
+    # build synaptic gains tab
+    add_synaptic_gain_tab(net, syn_gain_out, syn_gain_textfields, layout)
+
     return net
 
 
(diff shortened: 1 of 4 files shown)


asoplata:
Did some initial simulation tests using the new tab, and the simulated changes do appear to have the expected effects. Still need to actually review both the code and the tests.

asoplata:
Please see https://github.com/jonescompneurolab/hnn-core/issues/915 for some discussion on planning this for the GUI. 

Also, @dylansdaniels had a great idea: Instead of creating a new `Synaptic gains` Network tab, we could instead add the values to the top of `Connectivity`. This way, once we have each individual connection's `gain` also displayed in the `Connectivity` tab (currently a TODO), everything for the gains would be in the same location and with less clutter. Additionally, we could also add a button nearby with something like "Apply global synaptic gain changes" which applies a change via one of the four above gains, but then the user can inspect the current, actual value of the gain below on a per-connection basis.

I will try to add the per-connection `gain` to the `Connectivity` tab on this branch.

gtdang:
> Please see #915 for some discussion on planning this for the GUI.
> 
> Also, @dylansdaniels had a great idea: Instead of creating a new `Synaptic gains` Network tab, we could instead add the values to the top of `Connectivity`. This way, once we have each individual connection's `gain` also displayed in the `Connectivity` tab (currently a TODO), everything for the gains would be in the same location and with less clutter. Additionally, we could also add a button nearby with something like "Apply global synaptic gain changes" which applies a change via one of the four above gains, but then the user can inspect the current, actual value of the gain below on a per-connection basis.
> 
> I will try to add the per-connection `gain` to the `Connectivity` tab on this branch.

Would that involve a change to the API to set per-connection gain?

asoplata:
I don't think it would need a change to the API, since it should be analogous to changing `nc_dict`'s `A_weight` value. Related question: is this https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/gui/gui.py#L1818-L1819 where the GUI actually *applies* weight changes from the GUI to the Net object?

asoplata:
Noticed a small bug: Loading in a new network connectivity file adds another full set of the 4 connection gain types: 
Screenshot_20241022_172217

gtdang:
Yes, but you don't want to be applying the gain and updating the weight there. The nc_dict as a gain parameter. The multiplication should be left to the Network at simulation runtime.

hnn_core/network.py
```diff
@@ -299,6 +299,40 @@ def pick_connection(net, src_gids=None, target_gids=None,
     return sorted(conn_set)
 
 
+def _get_cell_index_by_synapse_type(net):
+    """Returns the indexes of excitatory and inhibitory cells in the network.
```
jasmainak, line 303:
```suggestion
    """Returns the indices of excitatory and inhibitory cells in the network.
```

hnn_core/network.py
```diff
@@ -299,6 +299,40 @@ def pick_connection(net, src_gids=None, target_gids=None,
...
+
+    This function extracts the source GIDs (Global Identifiers) of excitatory
```
jasmainak, line 305: "Global Identifiers" isn't too descriptive either ... it's a bit of a misnomer since it's mainly a programming construct. I like to say "cell ID" to be clear.

hnn_core/network.py
```diff
@@ -299,6 +299,40 @@ def pick_connection(net, src_gids=None, target_gids=None,
...
+    tuple: A tuple containing two lists:
+        - e_cells (list): The source GIDs of excitatory cells.
+        - i_cells (list): The source GIDs of inhibitory cells.
+    """
+
+    def list_src_gids(indices):
+        return np.concatenate([list(net.connectivity[conn_idx]['src_gids'])
+                               for conn_idx in indices]).tolist()
+
+    e_conns = pick_connection(net, receptor=['ampa', 'nmda'])
```
jasmainak, line 327:
```suggestion
    picks_e = pick_connection(net, receptor=['ampa', 'nmda'])
```

I like to use `e_conn` after the indexing, i.e.:

```py
e_conns = [net.connectivity[p] for p in picks_e]
```

so it's clear from the variable name that one is an element of `net.connectivity` and the other is simply an index

hnn_core/gui/gui.py
```diff
@@ -1719,6 +1737,24 @@ def add_cell_parameters_tab(cell_params_out, cell_pameters_vboxes,
                              cell_layer_radio_button.value)
 
 
+def add_synaptic_gain_tab(net, syn_gain_out, syn_gain_textfields, layout):
```
jasmainak, line 1740: missing docstring!

hnn_core/network.py
```diff
@@ -1497,6 +1525,49 @@ def update_weights(self, e_e=None, e_i=None,
...
+        - excitatory to inhibitory (e_i)
+        - inhibitory to excitatory (i_e)
+        - inhibitory to inhibitory (i_i)
+
+        The gain is assumed to be uniform within each connection type, and only
+        the first connection's gain value is used for each type.
+
+        Returns
+        -------
+        dict: A dictionary with the connection types ('e_e', 'e_i', 'i_e',
```
jasmainak, line 1543:
```suggestion
        values : dict
             A dictionary with the connection types ('e_e', 'e_i', 'i_e',
```

to be consistent with the docstring formatting standard

hnn_core/network.py
```diff
@@ -1497,6 +1525,49 @@ def update_weights(self, e_e=None, e_i=None,
...
+        # Define the connection types and source/target cell indexes
+        conn_types = {
+            'e_e': (e_cells, e_cells),
+            'e_i': (e_cells, i_cells),
+            'i_e': (i_cells, e_cells),
+            'i_i': (i_cells, i_cells)
+        }
+
+        # Retrieve the gain value for each connection type
+        for conn_type, (src_indexes, target_indexes) in conn_types.items():
```
jasmainak, line 1558:
```suggestion
        for conn_type, (src_idxs, target_idxs) in conn_types.items():
```

just for sake of consistency

hnn_core/network.py
```diff
@@ -1497,6 +1525,49 @@ def update_weights(self, e_e=None, e_i=None,
...
+            conn_indices = pick_connection(self,
```
jasmainak, line 1559:
```suggestion
            picks = pick_connection(self,
```

see above comment too ... would try to be consistent with naming

hnn_core/network.py
```diff
@@ -1497,6 +1525,49 @@ def update_weights(self, e_e=None, e_i=None,
...
+                                           src_gids=src_indexes,
+                                           target_gids=target_indexes)
+
+            if conn_indices:
```
jasmainak, line 1563:
```suggestion
            if len(picks) > 0:
```

to be explicit that `picks` is a list, not a `bool`

jasmainak (commented):
Looks good! Just left tiny nitpicks ... I didn't test though. I am hoping @dylansdaniels or @asoplata took care of that ;)

hnn_core/network.py
gtdang: I think using the truthyness of collections is more concise and pythonic.  If we really wanted to be explicit we could also name the variable "picks_list". 

But I'll open it up to the rest of the group to define a style guide. @ntolley @asoplata @dylansdaniels

jasmainak (approved):

jasmainak:
@asoplata @ntolley I see the PR has [MRG] in the title. Please feel free to merge if you're happy with the changes.

hnn_core/network.py
```diff
@@ -299,6 +299,40 @@ def pick_connection(net, src_gids=None, target_gids=None,
     return sorted(conn_set)
 
 
+def _get_cell_index_by_synapse_type(net):
```
asoplata, line 302: Is there a reason to put this outside of the `Network` class as a standalone function if it requires an existing `Network`?

asoplata:
This looks good, but @dylansdaniels and I have separately discussed that we would like to move these gain settings from their own subtab to the top of `Connectivity` subtab. I will try to make this change myself, and make a PR to your branch.

hnn_core/network.py
jasmainak: my general rule is to keep methods of a class minimal, so the structure of the class/object is clear. Just because it takes the object as input doesn't mean it should be a method. The method should have an explicit purpose. For example:

```py
ica = ICA()
ica.fit(X)
y = ica.predict(X)
```

once something becomes a method, it's very hard to disentangle it from the class

asoplata:
Back to working on this in prep for our next workshop. This branch has gone through significant changes.

First, it has been rebased on top of #1119 in order to pass a test using the new (unfinished) synaptic gains API and GUI code, which utilizes a network file at `tests/assets/gamma_L5weak_L2weak_hierarchical.json`.

Secondly, all of George's substantial work in prior commits were squashed into df3b1ac or 899de6c . This was done to make it easier to rebase onto `master`, for which there were substantial conflicts, especially due to the original work being in a pre-`ruff format` state. The rebase onto master (and #1119) is now completed.

Third, a few of the refactors like the fixing of frequent `cell_pameters_...` typos were reverted for the sake of a cleaner diff.

This is still a work in progress, but tests are passing. I've finished moving George's gains boxes to the `Network > Connectivity` subtab, and have added per-connection Gain float boxes and the eventual live-updated-gain multiplier text to their right: 

image

There's still a fair bit of work to do:
- [ ] To get and set gains for individual synapse types in the "Connectivity accordion", I need to rename the existing synaptic gain functions to "global" versions of the name, and then create new "single" versions of synaptic gain getters and setters to be used directly by the GUI.
- [ ] Then add live HTML text update whenever a user changes either a single or global gain value.
- [ ] Also need to test that the GUI use of all synaptic gain setters *actually works* and can reproduce simulation dipole output that is identical between API and GUI use of the same functions.
- [ ] Discovered that `Network.add_connection` cannot read in synaptic gain values, instead setting them all to 1.0. Since this is ultimately used as part of `hnn_io.read_network_configuration`, this is a bug and needs to be fixed, since custom synaptic gain values are *lost* on read in.
- I discovered that, prior to #1099, our `Network.update_weights` (which updates the synaptic GAINS, NOT the weights) would actually update all gains for both celltype connections AND external drives of a given type, since it only filtered by receptor type ([see here](https://github.com/jonescompneurolab/hnn-core/blob/7f3b7a932a9f25652b8580800cce3a7dfd9efd62/hnn_core/network.py#L1892)). I think this was a prior oversight, but #1099 's change was a desired one, since I personally think that changing any kind of global synaptic gains (like E-to-E) should NOT affect drives, and instead only affect celltype connections. In other words, the current behavior of `Network.update_weights` on master is slightly different from stable, since it now no longer affects external drives. I have added documentation as part of this PR to highlight that fact.

asoplata:
Props to Claude AI for helping with ipywidgets! Initial linking between single and global gains is working (in test case) 
image

asoplata:
Had to rebase / force-push again due to #1153

hnn_core/tests/test_gui.py
```diff
@@ -1346,3 +1364,206 @@ def test_default_frequencies(setup_gui):
...
+    """Test that synaptic gain changes are reproducible between the GUI and API."""
+    # Config
+    # --------------------------------
+    # Need a fully-detailed simulation to ensure spikes present (aka good data)
+    local_dt = 0.025
+    local_tstop = 100.0
+    global_custom_gain = 8.0
+    single_custom_gain = 2.0
+
+    # # AES: For some strange reason, the 3x3_drives network file does NOT produce
```
ntolley: should we try regenerating the .json file?

hnn_core/tests/test_network.py
```diff
@@ -1911,3 +1923,152 @@ def test_update_weights_metadata():
...
+def test_check_global_synaptic_gains_uniformity():
+    """Test _check_global_synaptic_gains_uniformity function."""
+    net = jones_2009_model()
+
+    # Set some global gains
+    net.set_global_synaptic_gains(e_e=0.5)
+
+    # With uniform gains (default), should return False (no warning)
+    result = _check_global_synaptic_gains_uniformity(net)
+    assert result is False
```
ntolley: To me the title of this function implies that `result` should be true when global synaptic gains are uniform

hnn_core/hnn_io.py
```diff
@@ -329,6 +329,16 @@ def _read_connectivity(net, conns_data):
             weight=conn_data["nc_dict"]["A_weight"],
             delay=conn_data["nc_dict"]["A_delay"],
             lamtha=conn_data["nc_dict"]["lamtha"],
+            threshold=(
```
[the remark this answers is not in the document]
asoplata: If I'm understanding what you mean about this, then yes: this is essentially a backwards-compatibility layer, since at some unknown point in the last two years, "threshold" and "gain" were added to Network files' "nc_dict"s. This is essentially a "versioning" problem of our Network files. Eventually (as part of #962) I plan to add both overt versioning of our Network files (by my estimate, we're around version 6-ish), along with a new module that handles exactly these kinds of backwards-compatibility issues.

However, I don't think it's worth time investing into any other possible bug-prevention strategy here that isn't tied to comprehensive changes with our input/output Network file formats.

hnn_core/network.py
```diff
@@ -313,6 +313,37 @@ def pick_connection(net, src_gids=None, target_gids=None, loc=None, receptor=Non
...
+    -------
+    e_cell_gids : list
+        The source GIDs of excitatory cells.
+    i_cell_gids : list
+        The source GIDs of inhibitory cells.
+    """
+    e_cell_gids = list()
+    i_cell_gids = list()
+    for cell_type_name, cell_data in net.cell_types.items():
+        if cell_data["cell_metadata"].get("electro_type") == "excitatory":
```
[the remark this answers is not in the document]
asoplata: @Chetank99 gets credit for this one ;)

hnn_core/network.py
```diff
@@ -2361,3 +2441,59 @@ def _add_cell_type_bias(
         cell_type_bias["section"] = section
 
     network.external_biases[bias_name][cell_type] = cell_type_bias
+
+
+def _check_global_synaptic_gains_uniformity(net):
+    """Check whether gain values for uniform within their type of connection.
```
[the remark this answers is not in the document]
asoplata: Resolved in one of the new commits.

hnn_core/tests/test_gui.py
```diff
@@ -1346,3 +1364,206 @@ def test_default_frequencies(setup_gui):
...
+    )
+
+    gains_default = _single_simulation["net"].get_global_synaptic_gains()
+    assert gains_default == {"e_e": 1.0, "e_i": 1.0, "i_e": 1.0, "i_i": 1.0}
+
+    # Change the synaptic weight widgets
+    gui.global_gain_widgets["e_e"].value = 0.5
+    gui.global_gain_widgets["e_i"].value = 0.5
+    gui.global_gain_widgets["i_i"].value = 1.1
+    gui.global_gain_widgets["i_e"].value = 1.1
```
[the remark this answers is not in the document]
asoplata: done in new commit

hnn_core/tests/test_gui.py
asoplata: Unfortunately, no. Regenerating the 3x3 json file does not appear to currently change it in any way, so the problem would still be there. Additionally, that test file is already known to have other issues that need investigation, see #960. For now, only testing the file *actually* used by the GUI is worthwhile (we should be testing it always anyways).

hnn_core/tests/test_gui.py
```diff
@@ -1346,3 +1364,206 @@ def test_default_frequencies(setup_gui):
...
+                widget._belongsto["src_gids"] == "L2_pyramidal"
+                and widget._belongsto["target_gids"] == "L2_pyramidal"
+            ):
+                conn_widget = widget
+                break
+        if conn_widget is not None:
+            break
+
+    assert conn_widget is not None, (
+        "Could not find L2_pyramidal->L2_pyramidal connection"
```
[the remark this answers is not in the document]
asoplata: Yeah that is weird and unnecessary. Removing in new commit.

hnn_core/tests/test_gui.py
```diff
@@ -1346,3 +1364,206 @@ def test_default_frequencies(setup_gui):
...
+    assert single_gain_widget.value >= 0  # Non-negative gain
+
+    # Check initial combined gain indicator value includes both gains
+    initial_single_gain = single_gain_widget.value
+    initial_global_gain = 1.0  # default
+    expected_initial = initial_single_gain * initial_global_gain
+    assert f"{expected_initial:.4f}" in combined_indicator.value
+
+    # Change global gain and check that combined indicator updates
+    gui.global_gain_widgets["e_e"].value = 2.0
```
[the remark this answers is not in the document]
asoplata: Yes, that is correct: this recalculation is executed by way of the combined indicator "observing" that widget field, which you can see here: https://github.com/brown-ccv/hnn-core-ccv/blob/53c3e207d18b26543b7756f2ef73045cec83c22a/hnn_core/gui/gui.py#L1526 . That's how the combined indicator can get a "live update" from two different input thingies

asoplata:
Okay @ntolley , I believe I've addressed all your comments. Most of them were implemented and then Resolved. The ones I left un-Resolved are worth a look to see if you agree or if the explanation or solution makes sense.

hnn_core/network.py
```diff
@@ -2361,3 +2448,59 @@ def _add_cell_type_bias(
...
+        "i_i": (i_gids, i_gids),
+    }
+
+    output_indicator = True
+    # Retrieve the gain value for each connection type
+    for conn_type, (src_idxs, target_idxs) in conn_types.items():
+        picks = pick_connection(net, src_gids=src_idxs, target_gids=target_idxs)
+
+        first_value = net.connectivity[picks[0]]["nc_dict"]["gain"]
+        for other_idx in range(1, len(picks)):
```
dylansdaniels: Is there a reason we don't do this above in _get_global_synaptic_gains when we extract the gain from the first connection?

It looks to me like we need to do this check every time anyway, and all this function does is reinstantiate the same setup to then do the check. It seems a bit duplicative to me

asoplata: We do use this check every time `Network.get_global_synaptic_gains` is called, here https://github.com/brown-ccv/hnn-core-ccv/blob/b30147dd326362c3262786b067bffd605719a3a3/hnn_core/network.py#L2078

hnn_core/gui/gui.py
```diff
@@ -2014,29 +2106,119 @@ def add_connectivity_tab(
...
+            title_box,
+            HBox(
+                [
+                    left_box,
+                    right_box,
+                ]
+            ),
+        ],
+        layout=Layout(
+            border="1px solid black",  # border width, style, and color
```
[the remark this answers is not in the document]
asoplata: Do you mean `"1px solid gray"`?

dylansdaniels:
I was immediately confused by the presentation of the gains here, which is probably not a great sign that our users will grasp it easily ... 

image

This reads to me like the local and global gains should add together. So the calculated weight would be 0.02 x (1+1) = 0.04, but obviously the weight stays 0.02 when both gains are set to 1

asoplata:
Relatedly, since the Weights are always multiplied by the Gains, and tied to your point, maybe it makes sense to *also* include indicator text to the right of the Weights textbox, which says something like `*Gain=<XXX> Total Weight`?

dylansdaniels:
Looks great, totally happy with this!

asoplata:
Latest commit seemed to do it. Border now looks identical to nearby borders of same width.

dylansdaniels:
Enabling auto merge. Thanks everyone!

PR #920, opened 2024-10-24
[MRG] change default smoothing for dipoles to 0

gtdang:
Changed default smoothing for simulation dipoles to 0.


asoplata (approved):
LGTM, plots for the gamma tutorial now look nice by default

PR #923, opened 2024-10-30
[ENH] In GUI, flip the distal drives when plotting the input histogram

dylansdaniels:


dylansdaniels:
@ntolley I haven't added tests yet but otherwise this is ready for class

PR #924, opened 2024-10-31
[MRG] New widget to adjust default smoothing value in gui

dylansdaniels:
Title says it all

### DIFF
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -906,9 +906,6 @@ class _VizManager:
             button_style="primary",
             style={'button_color': self.viz_layout['theme_color']},
             layout=self.viz_layout['btn'])
-        # self.make_fig_button.on_click(
-        #     lambda b: self.add_figure(self.default_smoothing)
-        # )
         self.make_fig_button.on_click(self.add_figure)
 
         self.datasets_dropdown = Dropdown(
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -549,11 +549,12 @@ def _clear_axis(b, widgets, data, fig_idx, fig, ax, widgets_plot_type,
         _dynamic_rerender(fig)
 
 
-def _get_ax_control(widgets, data, default_smoothing, fig_idx, fig, ax):
+def _get_ax_control(widgets, data, fig_default_params, fig_idx, fig, ax):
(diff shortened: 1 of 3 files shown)


hnn_core/gui/_viz_manager.py
```diff
@@ -902,6 +906,9 @@ def __init__(self, gui_data, viz_layout):
             button_style="primary",
             style={'button_color': self.viz_layout['theme_color']},
             layout=self.viz_layout['btn'])
+        # self.make_fig_button.on_click(
+        #     lambda b: self.add_figure(self.default_smoothing)
+        # )
```
ntolley, line 911:
```suggestion
```

hnn_core/gui/_viz_manager.py
```diff
@@ -937,6 +944,11 @@ def data(self):
             "figs": self.figs
         }
 
+    @property
+    def fig_default_params(self):
```
ntolley, line 948: Does this need to be `@property`? Seems like the other ones were created to regroup existing attributes for code cleanliness, whereas here you're just returning `self.fig_defaults`

hnn_core/tests/test_gui.py
```diff
@@ -1116,3 +1116,53 @@ def test_delete_single_drive(setup_gui):
...
+    new_smoothing = 0
+    gui.widget_default_smoothing.value = new_smoothing
+
+    gui.run_button.click()
+
+    # check that new default smoothing value is set everywhere
+    gui_smooth_value = gui.fig_default_params['default_smoothing']
+    viz_smooth_value = gui.viz_manager.fig_default_params['default_smoothing']
+
+    assert gui_smooth_value == new_smoothing
```
ntolley, line 1139: It'd be useful to add a similar assert for the original smooth values of the `fig_default` and `viz_manager`

hnn_core/gui/_viz_manager.py
dylansdaniels: I made it a @property so I could call it in the test:
```viz_smooth_value = gui.viz_manager.fig_default_params['default_smoothing']```

hnn_core/gui/_viz_manager.py
dylansdaniels: Good catch 😅

asoplata (approved):
LGTM. Tested and works, even for subsequent runs in-session

PR #928, opened 2024-11-06
hotfix: Change Morlet cycles for better alpha

asoplata:
This is primarily being changed for the upcoming Alpha and Beta tutorials


hnn_core/tests/test_gui.py
```diff
@@ -513,6 +513,8 @@ def test_gui_add_data_dependent_figure(setup_gui):
     axes_config_tabs = gui.viz_manager.axes_config_tabs
     assert len(fig_tabs.children) == 0
     assert len(axes_config_tabs.children) == 0
+    # Spectrogram needs longer time for wavelet analysis
+    gui.widget_tstop.value = 500
```
ntolley, line 517: I can see that this doesn't impact test length, but FYI we've discussed previously adding a hard limit to unit test runtime

ntolley (approved):
Looks great @asoplata!!

Happy to merge if it's all good with you

PR #929, opened 2024-11-07
fix: clear_axis contextually removes second axis

asoplata:
This fixes a prior bug with my prior addition to `_clear_axis` where it would always remove the histogram's inverted second Axes object. It now only removes if it the desired Axes being cleared is a spike histogram.


jasmainak (approved):
Looks good! @ntolley merge button is yours

PR #946, opened 2024-11-18
fix: log messages to the bottom of the output window instead of the top

gtdang:
Prints log messages to the bottom of the output instead of the top.

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -156,7 +156,7 @@ class _OutputWidgetHandler(logging.Handler):
             'output_type': 'stream',
             'text': formatted_record + '\n'
         }
-        self.out.outputs = (new_output, ) + self.out.outputs
+        self.out.outputs = self.out.outputs + (new_output, )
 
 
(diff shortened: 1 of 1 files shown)


gtdang:
A separate issue is that scrolling to the bottom of the output is not automatic. There is not currently a fix for that though. And it is a noted desired feature/milestone by the ipywidgets devs. 
https://github.com/jupyter-widgets/ipywidgets/issues/1815
https://github.com/jupyter-widgets/ipywidgets/milestone/2

jasmainak (approved):

asoplata (approved):
As far as I can test, this definitely solves the issue. However, something that was un-noticeable before is that if you have scrolled down the log, and then do something that prints output to the log (such as making a new figure), the log "refreshes" and automatically moves its position back up to the top. We probably never noticed it before since it was printing new messages to the top location itself. Hopefully, if we can ever get auto-scroll working, that will solve this annoyance. (That said, this is only an annoyance, not a bug, so I won't make a thing of it). Thanks George!

PR #952, opened 2024-11-22
[MRG] Add Min/Max Frequency and Add Visual Updates Ahead of 0.4

dylansdaniels:
Adding sections to Simulation tab, separating our the default values for visualization parameters that will be updated on 'run' 
Screenshot 2024-11-22 at 5 26 54 PM

Improvements to network connectivity tab, adjusting spacing of Vboxes, text indentation, adding css classes to more easily target sections, updating display of the receptor names
Screenshot 2024-11-22 at 5 27 14 PM

### DIFF
--- a/hnn_core/tests/test_gui.py
+++ b/hnn_core/tests/test_gui.py
@@ -1176,3 +1176,39 @@ def test_default_smoothing(setup_gui):
         assert gui.viz_manager.figs[figid].axes[0].has_data()
 
     plt.close('all')
+
+
+def test_default_frequencies(setup_gui):
+    """Tests that default min/max frequency are inherited correctly"""
+    gui = setup_gui
+
+    # check that the defaults are the same everywhere after running
+    # the default simulation
+    gui.run_button.click()
+
+    gui_min = gui.fig_default_params['default_min_frequency']
+    viz_min = gui.viz_manager.fig_default_params['default_min_frequency']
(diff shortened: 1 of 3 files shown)


dylansdaniels:
Note that the min/max spectral freq widgets are just placeholders here, they don't actually update the defaults like the Smoothing field does. But could add that to this or another PR

hnn_core/tests/test_gui.py
```diff
@@ -1174,3 +1176,39 @@ def test_default_smoothing(setup_gui):
...
+    # the default simulation
+    gui.run_button.click()
+
+    gui_min = gui.fig_default_params['default_min_frequency']
+    viz_min = gui.viz_manager.fig_default_params['default_min_frequency']
+    gui_max = gui.fig_default_params['default_max_frequency']
+    viz_max = gui.viz_manager.fig_default_params['default_max_frequency']
+
+    assert gui_min == viz_min == 10
+    assert gui_max == viz_max == 100
```
asoplata, line 1195: Everything looks great, the only thing I would change is removing the current default frequency values here. I think it's enough to check that they're consistent. We may want to change the default frequency values in the future, but with this check, we would also have to remember to manually change these lines in the tests.

dylansdaniels:
@asoplata made your change and similarly edited the initial equivalency check for test_default_smoothing!

asoplata (approved):
LGTM, tested the buttons in the GUI, everything seems to work

PR #977, opened 2025-01-24
[MRG] DEV: Add local codespell usage to dev install

asoplata:
This is a pretty trivial change: this adds support for automatic *local* install and usage of [codespell](https://github.com/codespell-project/codespell). This fixes an annoyance since, currently, we have a CI runner for `codespell` usage (see https://github.com/codespell-project/codespell ), but there's no automatic local way to test for such spelling mistakes before you push your commits.

This fixes the problem, in that now, `codespell` is used the same as our linting: it's installed as part of the `[test]` or `[dev]` options (e.g. `pip install "hnn_core[test]"`), and run during a developer's local `make test`. The TOML change also enables it to ignore any locally-built artifacts (which are not a problem for the CI runner).

After this, hopefully no more follow-up "`flake8 fix`" or "`spelling mistake`" commits :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,6 @@ clean :
 check-manifest:
 	check-manifest
 
-test: lint
-	pytest ./hnn_core/tests/ -m "not uses_mpi" -n auto
-	pytest ./hnn_core/tests/ -m "uses_mpi"
-
 lint:
 	@if command -v ruff > /dev/null; then \
 		echo "Running ruff check"; \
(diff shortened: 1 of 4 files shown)


jasmainak (approved):
LGTM. merge button is yours @ntolley @dylansdaniels !

asoplata:
Force-push was only for the `master` changes to unrelated dependencies in `setup.py`.

jasmainak:
Force push as much as you want on feature branches! Just don't sin on the `master` branch :)

asoplata:
Absolutely! I would never :)

ntolley:
@asoplata I'm flattered if you think this will stop my junk commits :smile: 

Do you mind rebasing and resolving the conflicts?

Makefile
```diff
@@ -29,3 +25,16 @@ lint:
 		echo "ruff not found, please install it!"; \
 		exit 1; \
 	fi;
+
+spell:
```
ntolley, line 29: just so I understand, this would be invoked by `make spell`?

Would it be possible update the contributing guide?

asoplata: Yes, and I've added that exact instruction to evoke it on the Contributing page.

ntolley (approved):
Perfect!! Just tested this locally and everything seems great

PR #971, opened 2025-01-21
[MRG] Change Sphinx theme (and fix minor typos)

asoplata:
This makes a big visual change to the "Code Website" at https://jonescompneurolab.github.io/hnn-core/stable/index.html , whereby it changes the theme from the unmaintained `sphinx-bootstrap-theme` to the actively maintained and popular `pydata-sphinx-theme`. The first commit also fixes some typos currently present in the documentation `Makefile`. This is part of the changes organized in #969.

This resolves #941 because search is provided entirely by the theme; this also enables us to remove `doc/_templates/search.html`.

This resolves #908 because version-hyperlinking is supported by the theme too, enabling us to remove `doc/_templates/navbar.html`. Instead, in the future, versions will need to be specific in `doc/_static/versions.json`.

A note about the versioning system: the PyData theme insists that the file your versioning metadata "...needs to be at a stable, persistent, fully-resolved URL" ( https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/version-dropdown.html#configure-switcher-json-url ). Thus, currently in this PR, the `versions.json` URL is set to point to an existing file I manually added to the `gh-pages` branch for the stable version, accessible here: https://jonescompneurolab.github.io/hnn-core/stable/_static/versions.json ( new code using this resource is located here: https://github.com/jonescompneurolab/hnn-core/commit/b85b222dc02c1455b12559c4a78ad99ede82011f#diff-e170e9a7d787c21095c6c11bb25f0f1ff0294a42a46d45ba6fb5ed794e457624R105 ). AFTER this PR is merged, 3 things will happen:
1. The `master` branch will then contain its own local copy of `doc/_static/versions.json`.
2. The CircleCI docs build runner will update the `dev` version of the website, which will then cause this new file to be located at `https://jonescompneurolab.github.io/hnn-core/dev/_static/versions.json`. This is the "version" of the "version file" that we should use after this point (which is what other theme users do).
3. We need to then make a second, much smaller PR, where we remove the file I added to https://jonescompneurolab.github.io/hnn-core/stable/_static/versions.json , and replace the line in `doc/conf.py` to link to the "dev" version of our "version file". After this point, anytime there is a new version, one will only need to update the `master` branch's copy of the version file, and then it will be propagated.

PR #978, opened 2025-01-24
[MRG] fix: relax test for poisson drive stats

asoplata:
This is a trivial fix for stochastic, difficult-to-reproduce test failure. Rarely, and seemingly only on Python 3.8 `Unit Test` runners, this test will fail due to the random Poisson mean interval exceeding the `approx` threshold. Here is an example case: https://github.com/asoplata/hnn-core/actions/runs/12952951618/job/36131478285#step:9:164 . I noticed the same thing happening on the first time I tried to run the tests for #971 ; however, since I told it to re-run the failed jobs, now I can't find any data on the previously-failed jobs. (It's possible that when you re-run failed jobs, maybe you can't access the old failed data using GitHub Actions?). As you can see in the above example, the test failed because the mean event interval was around 18.98, which is *just* outside of the threshold of 20 +/- 1.

This may have been a pre-existing issue that sometimes produced random failures in the tests, but only rarely and stochastically. By expanding the tolerance for this test, this should become *far* less likely to cause test failures in the future.

To be clear, scientifically, we're saying that the mean of our realized Poisson rate is within 1.5 ([Hz] I believe) of our expected value, instead of 1 [Hz], for testing purposes. We could also tighten the tolerance gap back to something like 1.1 if anyone else is uncomfortable with this change (@ntolley ?)

### DIFF
--- a/hnn_core/tests/test_drives.py
+++ b/hnn_core/tests/test_drives.py
@@ -59,7 +59,7 @@ def test_external_drive_times():
     lamtha = 50.
     event_times = _create_extpois(t0=0, T=100000, lamtha=lamtha, prng=prng)
     event_intervals = np.diff(event_times)
-    assert pytest.approx(event_intervals.mean(), abs=1.) == 1000 * 1 / lamtha
+    assert pytest.approx(event_intervals.mean(), abs=1.5) == 1000 * 1 / lamtha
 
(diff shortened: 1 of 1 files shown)


ntolley (approved):
Fascinating edge case! Thanks @asoplata!

Issue #969, opened 2025-01-14
DOC: Refresh code-website


asoplata:
Firstly, when I say the "Code-website", specifically I mean https://jonescompneurolab.github.io/hnn-core/stable/index.html . I say this to differentiate it from the "Main HNN website" here https://hnn.brown.edu/ or the "New Tutorial" website.

This is intended as a catch-all issue to track all the TODOs that Austin needs to do in order to get the website "up to date". This does not include major architectural changes like the build system, but rather dealing with smaller things that have broken over time, improving the ease of contributing to documentation by changing most of the website to use Markdown (MyST), and small text improvements. This will be done through many smaller PRs.

## Technical changes:

Note: I have done preliminary testing (unpushed) of all of the below, and everything works.

- [ ] Fix bugs in the sphinx Makefile
- [ ] Change the theme in order to fix the drop-downs ( #908 ) and search ( #941 ):
    - See my comment here: https://github.com/jonescompneurolab/hnn-core/issues/908#issuecomment-2584414827 . Unfortunately, the easiest (and best) solution is switching our Sphinx theme. Since many things in `doc/conf.py` are theme-specific, I'm experimenting with how to get the page to use the new theme to resemble the older one.
    - The obvious theme to switch TO is https://pydata-sphinx-theme.readthedocs.io/en/stable/ . This will look familiar if you've spent time in the numerical computing community of Python, since it's used for the documentation of MNE, Numpy, Scipy, Pandas, etc. Importantly, this theme has built-in version switching https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown . In my tests with applying this theme, it fixes both of the above issues.
- [ ] Transition from ReStructured Text (RST) to Markdown (where applicable), specifically via the MyST-parser ( https://www.sphinx-doc.org/en/master/usage/markdown.html#markdown ):
    - Reasons: Markdown is significantly more popular and easier to use than RST, in part because it is far less powerful. Changing our main code-website documents to also support using Markdown via MyST https://myst-parser.readthedocs.io/en/latest/ allows us to continue using RST for some files, MyST syntax that maps to RST, and raw RST inside MyST files, but also users who only know Markdown to contribute more easily. Most of our non-API documentation on the code-website already makes very little use of RST, and most of them need very little work to convert to MyST (I've already done almost all the conversion, but haven't pushed them yet).
[...]

PR #980, opened 2025-01-24
[MRG] doc: small authoring, also separate/expand install

asoplata:
In addition to some minor changes in the code-website documentation (such as putting our logo at the top of the README), the main focus of this PR is mentioned in #969:

1. Moving the installation instructions to their own page, from both the README and the Parallel pages,
2. Reformatting of the Installation instructions,
3. Adding installation documentation that was missing before but is necessary (such as Xcode Command-Line Tools on MacOS, Conda install of MPI on Linux, etc.), and
4. More hyperlinking between Usage near the Install instructions for every case.


asoplata:
Note that, due to the way our documentation system works (adding links to a not-yet-created webpage is a circular dependency), the only way to actually view this new Install page is to either:

1. Download this branch and build the docs yourself (the best way)
2. View the source Markdown file here: https://github.com/asoplata/hnn-core/blob/iss969-authoring/doc/install.md (not as good, but can be seen instantly).

README.md
```diff
@@ -174,20 +92,19 @@ recommend editors such as [VS Code](https://code.visualstudio.com) and
 
 # Bug reports
 
-Use the [github issue
+Use the [GitHub Issues
```
ntolley, line 95: in the future we might consider adding an issue template specifying what information that needs to be included

README.md
```diff
@@ -198,4 +115,8 @@ JOSS](https://doi.org/10.21105/joss.05848):
 > circuit-level interpretation of human MEG/EEG. *Journal of Open Source
 > Software*, 8(92), 5848, <https://doi.org/10.21105/joss.05848>
 
-[contributing guide]: https://jonescompneurolab.github.io/hnn-core/stable/contributing.html
+[Contributing Guide]: https://jonescompneurolab.github.io/hnn-core/stable/contributing.html
+
+<!-- TODO: Once we have a "stable" version of the install webpage, need to update this link to it: -->
```
ntolley, line 120: assuming this is the equivalent in a comment in markdown?

doc/install.md
```diff
@@ -0,0 +1,288 @@
+(install)=
+# Installation Guide
+
+If you have any questions or problems while installing `hnn_core`, feel free to
```
ntolley, line 4: do you think this should be a github issue instead of a discussion? In my head discussions are better used for conceptual questions

doc/install.md
```diff
@@ -0,0 +1,288 @@
...
+--------------
+
+# Step 3. `hnn_core` installation types
+
+Note that Step 1 and Step 2 above are required for **all installation types**. Note that you can install multiple sets of HNN features below during the install process by combining them similar to `pip install "hnn_core[gui,parallel]"`.
+
+## Basic Installation
+
+To install just the `hnn_core` API, open a Terminal or Command Prompt using a fresh virtual
+environment, and simply run:
```
ntolley, line 63:
```suggestion
environment, and enter:
```

doc/install.md
```diff
@@ -0,0 +1,288 @@
...
+
+**Basic Usage**: For how to use the `hnn_core` API, which is available for every installation
+type, see our extensive {doc}`Examples page <auto_examples/index>`.
+
+## GUI Installation
+
+To install `hnn_core` with both its API and GUI support, a simple tweak to the above command is
+needed (pay attention to the "quotes"):
+
+    $ pip install "hnn_core[gui]"
```
ntolley, line 90: does this work for zsh as well?

doc/install.md
```diff
@@ -0,0 +1,288 @@
...
+To check if everything worked fine, you can run:
+
+    $ python -c "import hnn_core"
+
+and it should not give any error messages.
+
+**Basic Usage**: For how to use the `hnn_core` API, which is available for every installation
+type, see our extensive {doc}`Examples page <auto_examples/index>`.
+
+## GUI Installation
```
ntolley, line 85:
```suggestion
## Graphical User Interface (GUI) Installation
```

ntolley (approved):

README.md
asoplata: Yes, creating standard issue templates is a task currently stored in our HNN notion https://www.notion.so/jonescompneurolab/DOC-Create-Issue-templates-17227379ff31431886fee83cc53eded6 but not on Github itself yet

README.md
asoplata: Yes, Markdown re-uses the same comment syntax as HTML, which is seen here. More context: https://github.github.com/gfm/#html-comment

doc/install.md
asoplata: Yes, as far as I can tell, this works across both bash and zsh. If we tell people to always use quotes then this should always work.

asoplata:
@ntolley I've either responded to or made changes according to each of your comments as of now, this is ready for re-review

ntolley (approved):

Issue #935, opened 2024-11-13
Bug: Github Action for Issue Tracking not always running


asoplata:
Problem: Our Issue Metrics Github Action here https://github.com/jonescompneurolab/hnn-core/blob/master/.github/workflows/issue-metrics.yml does not always run for some reason I don't understand. If you look for Issues containing the automated Monthly reports (here https://github.com/jonescompneurolab/hnn-core/issues?q=is%3Aissue+monthly ), then you see that they've only been run for the 2024 months of May, July, and September.

Solution: hack the mainframe and solve the problem. Maybe it's the Cronjob arguments, or the date definition, or something else.

More context: As an example, in #900 , at the bottom of the auto-generated report, it gives a query that you can copy and use to filter the issues you're interested in. The query given is for all days in September, month 2024-09. If we adapt this query for August for example, such as here https://github.com/jonescompneurolab/hnn-core/issues?q=is%3Aissue+created%3A2024-08-01..2024-08-31 , then we see that there were issues created in that month. This should have lead to a Monthly metrics report being created for August, however no report was automatically generated.

This is especially weird, since as far as I can tell, the Action workflow that we use was, rightly, modeled after the Github Blog's example for exactly this situation: https://github.blog/open-source/maintainers/metrics-for-issues-pull-requests-and-discussions/ . It is possible the bug is inherent to that script itself, as opposed to any customizations we made.

ntolley:
Maybe  look at the script used to process/analyze the issue metadata and modify it to run locally for reporting?

FYI here's where to look for failing jobs https://github.com/jonescompneurolab/hnn-core/actions/workflows/issue-metrics.yml

PR #983, opened 2025-01-29
[MRG] fix: copy template for monthly metrics workflow

asoplata:
This copies over the exact content of the template for the `issue_metrics` workflow from
https://github.com/github/issue-metrics/blob/main/docs/example-workflows.md EXCEPT for the `SEARCH_QUERY` and removal of `assignees`. This may help with #935 , possibly due to the change in `permissions: contents`.

### DIFF
--- a/.github/workflows/issue-metrics.yml
+++ b/.github/workflows/issue-metrics.yml
@@ -1,12 +1,11 @@
 name: Monthly issue metrics
-
 on:
   workflow_dispatch:
   schedule:
-    - cron: '3 2 1 * *'
+    - cron: "3 2 1 * *"
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


asoplata:
Here is an example of a recent failure of the issue metrics run: https://github.com/jonescompneurolab/hnn-core/actions/runs/12566527833/job/35032026617 . At least in this case, this appears to be a permissions error. This may be fixed by the appropriate change in this PR.

dylansdaniels:
So you changed "permissions: contents:" from "write" to "read", but I'm wondering how that potentially solves the issue in the failed run you shared? Just trying to wrap my head around it

```You do not have permission to view a repository from: 'jonescompneurolab/hnn-core '; Check your API Token.```

Also, I don't see `assignees` in the original file before your changes... Am I missing something?

asoplata:
Here https://github.com/github/issue-metrics/issues/166 is what appears to be the only discussion I could find that specifically deals with our issue. I initially figured that a "weakening" of the permissions, as was done in the examples given by the action repo, may help to prevent the issue, since I initially figured it to *be* a permissions error. However, now I'm not so sure: the issue I linked strongly suggests that despite the error message indicating it's a permissions issue, it is actually an API limiting issue. This would explain why it happens randomly. It may perhaps be that the only full-proof solution right now is still "just re-run it if it breaks".

Regardless, I do think we should merge this PR just in the off chance that it helps; it makes our workflow script slightly more similar to the direct example provided by the repo. However, if anyone else feels strongly that that's not enough justification, then we can close this PR without merging.

Re: `assignees`, I only meant that I removed the final line when copying from https://github.com/github/issue-metrics/blob/main/docs/example-workflows.md#calculated-time-example , NOT from our personal earlier version of the file.

PR #985, opened 2025-01-30
Adjust cell spacing

### LINKED ISSUE #968: MODEL: Adjust cell spacing
cc: @ntolley @carolinafernandezp @katduecker 

I don't know the details, but as Nick mentioned in the last dev meeting, there is a scientific issue having to do with placement of the cells in space. IIRC, the issue with the placement of the inhibitory and excitatory cells should not, in principle, affect the current-dipole output of our simulations, but it will likely affect any LFP/CSD output. In order for us to also work towards scientifically meaningful LFP/CSD, it sounded like this is an issue that needs to be addressed. I've made this issue for us to track our work.

carolinafernandezp:
Closes #968

### DIFF
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -64,7 +64,7 @@ def _create_cell_coords(n_pyr_x, n_pyr_y, zdiff, inplane_distance):
         list_coords = [pos for pos in it.product(xxrange, yyrange, [zdiff])]
         return list_coords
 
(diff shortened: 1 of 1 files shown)


carolinafernandezp:
Hi @asoplata 
Thanks for the helpful comments! 
I made the changes and it looks like all the tests passed.

PR #986, opened 2025-01-30
Adding a new plot for vizualizing drive strength across every cell types 

dikshant182004:
i have added the option for normalization (and i was thinking to use standard scaler rather than min max one ) and i have also added a feature to represent the units for the absolute and normalized values .i have  also added mplcursors for better interactivity and user engagement .

### DIFF
--- a/hnn_core/viz.py
+++ b/hnn_core/viz.py
@@ -1171,6 +1171,132 @@ def plot_connectivity_matrix(net, conn_idx, ax=None, show_weight=True,
     return ax.get_figure()
 
 
+def plot_drive_strength(net, show_weight=True, ax=None,
+                        colorbar=True, color_scale='linear',normalize=True, show=True):
+    """Plot the relative strength of drives to cell types.
+
+    Parameters
+    ----------
+    net : Network
+        Instance of a Network object.
+    ax : matplotlib.Axes | None
+        Matplotlib axes. If None, create a new figure.
+    show_weight : bool
+        If True, visualize connectivity weights as gradient.
+        If False, all weights set to constant value.
+    colorbar : bool
+        If True (default), adjust figure to include colorbar.
+    color_scale : str ('linear' or 'log')
+        Color scaling for drive strength.
+    normalize : bool
+        If True, normalize the strength values to be between 0 and 1.
+    show : bool
+        If True, show the plot immediately.
+
+    Returns
+    -------
+    fig : matplotlib.figure.Figure
+        The figure handle.
+    """
+    import numpy as np
+    import matplotlib
+    matplotlib.use('TkAgg')   
+    import matplotlib.pyplot as plt 
+    from .network import Network
+    from .cell import _get_gaussian_connection
+    import mplcursors
+
+    _validate_type(net, Network, 'net', 'Network')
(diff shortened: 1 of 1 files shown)


hnn_core/viz.py
```diff
@@ -1171,6 +1171,132 @@ def plot_connectivity_matrix(net, conn_idx, ax=None, show_weight=True,
...
+        If True, show the plot immediately.
+
+    Returns
+    -------
+    fig : matplotlib.figure.Figure
+        The figure handle.
+    """
+    import numpy as np
+    import matplotlib
+    matplotlib.use('TkAgg')
```
asoplata, line 1203: We do not set or require a specific backend for matplotlib outside of our test code. If this is necessary for your code, why?

hnn_core/viz.py
```diff
@@ -1171,6 +1171,132 @@ def plot_connectivity_matrix(net, conn_idx, ax=None, show_weight=True,
...
+    fig : matplotlib.figure.Figure
+        The figure handle.
+    """
+    import numpy as np
+    import matplotlib
+    matplotlib.use('TkAgg')   
+    import matplotlib.pyplot as plt 
+    from .network import Network
+    from .cell import _get_gaussian_connection
+    import mplcursors
```
asoplata, line 1207: Please remove this import and its usage. It is best practice to limit our dependencies as much as possible, and we should only add them if they are absolutely necessary. Adding any new dependencies contributes significantly to maintenance burden.

hnn_core/viz.py
```diff
@@ -1171,6 +1171,132 @@ def plot_connectivity_matrix(net, conn_idx, ax=None, show_weight=True,
...
+
+    ax.set_xticks(np.arange(len(cell_types)))
+    ax.set_yticks(np.arange(len(drive_names)))
+    ax.set_xticklabels(cell_types)
+    ax.set_yticklabels(drive_names)
+    ax.set_xlabel('Cell Types')
+    ax.set_ylabel('Drive Names') 
+
+    # For addition of units information 
+    unit_text = "Units: Synaptic Weight" if not normalize else "Units: Relative Strength (0-1)"
```
asoplata, line 1277: "Synaptic Weight" is not a scientific physical unit. The physical units of synaptic weights are usually in `uS` in HNN-Core: you should check that this is the case for all drives (such as in `network.py` and elsewhere).

asoplata (changes requested):
In addition to these changes, please add tests for this new plotting function and example use cases into https://github.com/jonescompneurolab/hnn-core/blob/master/examples/howto/plot_connectivity.py . After you have added examples, make sure that the examples output correctly when you build the documentation (which requires you to manually do https://jonescompneurolab.github.io/hnn-core/stable/contributing.html#building-the-documentation ).

hnn_core/viz.py
dikshant182004: that was added basically for mplcursors to support interactivity since it is not required i will remove this

hnn_core/viz.py
dikshant182004: okay i will remove this

hnn_core/viz.py
dikshant182004: i will change the units

PR #988, opened 2025-01-31
[MRG] maint: make test net generation more explicit

asoplata:
This moves the re-generation of our test network (in `hnn_core/tests/assets/jones2009_3x3_drives.json`) from `test_io.py` into its own, executable file. This also adds an easy way to run it, and documentation for when to run it. Note the test network is probably the case of the test failures in #985 ; this may help with that issue.


hnn_core/tests/regenerate_test_network.py
```diff
@@ -0,0 +1,72 @@
...
+
+from pathlib import Path
+
+from hnn_core import read_params, jones_2009_model
+
+hnn_core_root = Path(__file__).parents[1]
+assets_path = Path(hnn_core_root, 'tests', 'assets')
+
+
+def jones_2009_additional_features():
```
ntolley, line 11: Just documenting from our in person conversation, it'd be ideal to find a solution that doesn't have an identical copy the code laying around that has to match.

I do think calling this from a fixture should be fine as it's essentially the same as calling a function from a package import (which is already done in the existing fixture).

asoplata: Okay, did what we discussed in the latest commit.

asoplata:
Force-pushed to resolve `Makefile` conflict. Ready for review.

dylansdaniels (approved):
Thanks for adding the updated description of the floating point discrepancy. This is good to go now.

PR #1000, opened 2025-02-22
protoype of updated docstring

dikshant182004:
i have changed some of the funtions docstring ,if they are okay then we can update that throughout our code base.


asoplata:
Hey @dikshant182004 , you appear to be re-using an old branch you used for a different PR. Please apply your changes to a new branch that you create off of an up-to-date version of the `master` branch. The name of your new branch doesn't matter, but ideally it should be unique to its PR.

I have to close this PR because GitHub does not allow you to change the source branch of a PR. Please make a new PR with the new branch that you create.

As a sidenote, I will point out that part of the reason we want to use standalone "feature branches" is that the history of your feature branch does not necessarily match the history on `master`, even after we merge your code in. This is because we squash and rebase your feature branch into `master`.

dikshant182004:
@asoplata okay i have understood it ,i will create a new pr with new branch.

PR #1007, opened 2025-03-06
[MRG]: DOC Update whatsnew

asoplata:
This significantly adds to the `doc/whats_new.md` page with mention of both PRs and contributors that were not added at the time their work was merged. There were many missing contributions that should now be present in our changelog. I believe I've added all outstanding PRs that were merged into `master` (as opposed to the HDF5 changes in `upstream`'s `network-rw` branch, etc.).

I also re-organized them into a couple of categories (API, GUI, Bug, other).

Since this is only additions to contribution-documentation, this should be an easy merge.

Note that this is NOT the final version of the release notes for v0.4. I will write up some actual summary content similar to https://numpy.org/doc/stable/release/2.2.0-notes.html when we get closer to the v0.4 release very soon.

### DIFF
--- a/doc/whats_new.md
+++ b/doc/whats_new.md
@@ -9,173 +9,169 @@ orphan: true
 
 ### API changes
 
-- Add ability to manually define colors in spike histogram plots,
-  by [Nick Tolley][] in {gh}`640`
+- Add ability to manually define colors in spike histogram plots, by [Nick Tolley][] in
+  {gh}`640`
 
-- Connection `'src_gids'` and `'target_gids'` are now stored as set objects
-  instead of lists, by [Ryan Thorpe][] in {gh}`642`.
+- Connection `'src_gids'` and `'target_gids'` are now stored as set objects instead of
+  lists, by [Ryan Thorpe][] in {gh}`642`
 
 - {func}`~hnn_core.CellResponse.write` and {func}`~hnn_core.Cell_response.read_spikes`
-  now support hdf5 format for read/write Cell response object, by
-  [Rajat Partani][] in {gh}`644`
+  now support hdf5 format for read/write Cell response object, by [Rajat Partani][] in
+  {gh}`644`
 
 - Add ability to customize plot colors for each cell section in
   {func}`~hnn_core.Cell.plot_morphology`, by [Nick Tolley][] in {gh}`646`
 
-- {func}`~hnn_core.Dipole.write` and {func}`~hnn_core.Dipole.read_dipoles`
-  now support hdf5 format for read/write Dipole object, by
-  [Rajat Partani][] in {gh}`648`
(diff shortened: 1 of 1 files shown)


doc/whats_new.md
```diff
@@ -627,5 +803,7 @@ orphan: true
 [Yaroslav Halchenko]:  https://github.com/yarikoptic
 [Tianqi Cheng]: https://github.com/tianqi-cheng
 [Carolina Fernandez Pujol]: https://github.com/carolinafernandezp
-[Austin Soplata]: https://github.com/asoplata
+[Austin E. Soplata]: https://github.com/asoplata
 [Dikshant Jha]: https://github.com/dikshant182004
+[Dan "pynmash"]: https://github.com/pynmash
```
ntolley, line 808: @pynmash before we merge this PR, do you have a full/official name you'd like us to use?

pynmash: Thanks for asking, Dan Toms, please.

jasmainak:
@asoplata I just wanted to point out that `whats_new` page is meant to be user facing. When you make the new release, you pick the most important changes from the `whats_new` page and send the email. Internal changes that do not affect users are typically not documented. Also changes that were reverted in the same release cycle are typically not documented. It's indeed a way to give new contributors credit, but also a way to tell users "what is new". Since maintenance fixes are typically made by established contributors, it's not too bad if they don't get credit for a PR or two in the whats_new document

asoplata:
@jasmainak yes I understand that the page is primarily for users, especially for highlighing new features, compatibility/deprecations, and bugs. As I said before, this is not the final version of the release notes, and I was planning on highlighting the most important changes in a user-friendly way at the top of the changelog, similar to a combination of a full and maintenance release like https://numpy.org/doc/stable/release.html , maybe using this as a template:

```
### Deprecations
(if any)...

### New Features
(not written yet)...

### API changes
(not written yet)...

### Contributors
(done)

### Bug fixes and corrections
(done)

### PRs merged (API)
(done)

### PRs merged (GUI)
(done)

### PRs merged (Other)
(done)
```

For this release, because there were only "release-candidate" releases between v0.3 and v0.4, and there's almost 2 years of work in v0.4, I figured it would be best to include all the work in the single v0.4 release since it is considerable; as you point out, this is not the way it's normally done with maintenance vs. full releases.

An alternative option is we treat the new changelog in this PR as a single changelog for only a new release-candidate (e.g. v0.4rc5). Then, we make a separate section for main v0.4 release notes, making them independent like you suggest.

One more complex option: We could retroactively treat the different release-candidate "releases" as maintenance releases, and then segment the PR changelog according to each of them. Basically, this would involve going back and re-building the changelogs of each release-candidate that should have been added at the time of each. This would take a bit of work since there aren't tags for most of them, so I'd need to infer which commit belongs to which release-candidate. Then, similarly, we make a separate section for main v0.4 release notes, making them independent.

dylansdaniels:
lgtm, thanks Austin!

PR #1008, opened 2025-03-10
[WIP] Maint: limit py versions to 3.8>=x<3.13

asoplata:


asoplata:
Canceling this since:

1. After further investigation yesterday (see https://github.com/jonescompneurolab/hnn-core/issues/947#issuecomment-2711449405 ), this probably won't help provide a clearer error message to the user,
2. We would have to revert it anyways once NEURON supports 3.13,
3. Independent of this change, we need to update all install documentation to include that `hnn-core` does not support Python 3.13.

PR #1009, opened 2025-03-12
[test]: improve BatchSimulate test cases

### LINKED ISSUE #955: TEST: Improve `BatchSimulate` testing
Currently, testing of `BatchSimulate` is very incomplete. 

- [ ] Importantly, as @ntolley pointed out here https://github.com/jonescompneurolab/hnn-core/pull/857#issuecomment-2499288882 , we need a way to test that the `loky` backend not only creates multiple workers, but *actually uses* the multiple parallel workers, in parallel. I think we may have to actually use `loky` directly in our tests for this (see https://github.com/jonescompneurolab/hnn-core/pull/857#issuecomment-2501908460 ).

- [ ] The `threading` backend only has a single test, the `multiprocessing` and `dask` backends are not tested at all.

- [ ] Much of the testing is parameter `TypeError` testing, which is not testing `BatchSimulate`'s functionality so much as it is testing `_validate_type` and `_check_option`. I hate to say it, but I think many of these tests can be removed. Relatedly, I will make an issue for introducing typing and type-checking for the codebase as a whole. I'm fine with leaving these tests in until we get more formal type-checking implemented, however.

- [ ] Relatedly, we should consider dropping the 'multiprocessing' backend as an option, since it has quirks in Joblib and, judging from joblib's documentation, is effectively replaced by `loky` (see https://joblib.readthedocs.io/en/stable/parallel.html#old-multiprocessing-backend ).

Freedisch:
fix #955 

- added `test_loky_worker_creation`
- added `test_dask_backend`
- added `test_summary_func`

NB: Fixing bug with the multiprocessing backend testing

### DIFF
--- a/hnn_core/tests/test_batch_simulate.py
+++ b/hnn_core/tests/test_batch_simulate.py
@@ -8,6 +8,9 @@ import time
 import pytest
 import numpy as np
 import os
+from dask.distributed import Client
+import shutil
+from unittest.mock import patch
 
 from hnn_core.batch_simulate import BatchSimulate
 from hnn_core import jones_2009_model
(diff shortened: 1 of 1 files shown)


asoplata:
Hello Freedisch, and thank you for your contributions! Unfortunately, your PR helped me discover some pre-existing issues that I was not aware of with `BatchSimulate`, which I have documented in updates to #955. We need to fix things on our end before we proceed with adding tests such as the tests that you wrote. 

Don't worry, your effort has been noted for our GSoC considerations :). In the meantime, I recommend you submit a PR based off of another one of our good first issues: https://github.com/jonescompneurolab/hnn-core/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22 .

PR #1011, opened 2025-03-14
fix: add back in GUI output log print to top

### LINKED ISSUE #953: GUI: Change logging output to show in "reverse time order"
Due to the difficulties with auto-scroll (see #940 ), we have decided to have the GUI log window show output in reverse time order, meaning that new log messages are printed to the top as they occur.

This may require capturing NEURON output into a logger itself.

asoplata:
Resolve #953 for the final time.

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -164,7 +164,7 @@ class _OutputWidgetHandler(logging.Handler):
             'output_type': 'stream',
             'text': formatted_record + '\n'
         }
-        self.out.outputs = self.out.outputs + (new_output, )
+        self.out.outputs = (new_output, ) + self.out.outputs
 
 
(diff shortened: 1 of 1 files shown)


hnn_core/gui/gui.py
```diff
@@ -164,7 +164,7 @@ def emit(self, record):
             'output_type': 'stream',
             'text': formatted_record + '\n'
         }
-        self.out.outputs = self.out.outputs + (new_output, )
+        self.out.outputs = (new_output, ) + self.out.outputs
```
ntolley, line 167: Any thoughts as to why we thought this was fixed earlier?

asoplata: See here https://github.com/jonescompneurolab/hnn-core/issues/953#issuecomment-2725825119

PR #1012, opened 2025-03-14
doc: add warning about py3.13

asoplata:
Trivial PR: add a warning about not using Python 3.13 with `hnn-core`. Related to #947 .


ntolley (approved):
Perfect, thanks for adding this @asoplata!

PR #1013, opened 2025-03-15
pretty-print-network with md format

Chetank99:
i have added a new function `network_to_markdown()`, that exports a network object's config to md format. It is using the current JSON-based network export format.

Added `network_to_markdown()` to `hnn_io.py`

I used this code to test it:
```
from hnn_core import jones_2009_model
from hnn_core.hnn_io import network_to_markdown

#if we dont want to save, we can remove file param and just print
md_text = network_to_markdown(net,'file.md') 
print(md_text)
```
Attaching a sample md file: 
[file.md](https://github.com/user-attachments/files/19259632/file.md)

This is my first PR, so very open to feedback! If this is fine, I can work on making a table with differences of 2 network params

### DIFF
--- a/hnn_core/hnn_io.py
+++ b/hnn_core/hnn_io.py
@@ -15,6 +15,244 @@ from .cell import Cell, Section
 from .cell_response import CellResponse
 from .externals.mne import fill_doc
 
+def network_to_markdown(net, fname=None):
+    """converts network parameters to a markdown table."""
+    from datetime import datetime
+    
+    md = []
+    md.append("# HNN Network Configuration")
+    md.append("")
+    md.append(f"*Generated on: {datetime.now().strftime('%Y-%m-%d %H:%M:%S')}*")
+    md.append("")
+    md.append("")
+    
+    # sim params
+    md.append("## Simulation Parameters")
+    md.append("")
+    md.append("| Parameter | Value |")
+    md.append("|-----------|-------|")
+    md.append(f"| Time step (dt) | {net._params.get('dt', 'Not set')} ms |")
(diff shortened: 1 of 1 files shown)


hnn_core/hnn_io.py
```diff
@@ -15,6 +15,244 @@
 from .cell_response import CellResponse
 from .externals.mne import fill_doc
 
+def network_to_markdown(net, fname=None):
```
jasmainak, line 18: If this is a `write_xxx` function, it should follow the API in the remaining codebase

hnn_core/hnn_io.py
```diff
@@ -15,6 +15,244 @@
...
+    """converts network parameters to a markdown table."""
+    from datetime import datetime
```
jasmainak, line 20: top-level import

hnn_core/hnn_io.py
```diff
@@ -15,6 +15,244 @@
...
+    
+    md = []
```
jasmainak, line 22:
```suggestion
    md = list()
```

hard to distinguish between `[]` and `{}` otherwise

jasmainak:
can you add a test?

hnn_core/hnn_io.py
jasmainak: parameters should be documented in the docstring

Chetank99:
Hi @jasmainak , 
thank you for the review and comments, addressed them, except the test.

since it is a write_xxx function, used the syntax to have the overwrite param

documented via docstring

used list() for md

I am not sure right now how to add a test, going through `hnn_core/tests` to learn how the code base does it

jasmainak:
I will let @asoplata and @ntolley review the PR and merge if happy :)

PR #1019, opened 2025-03-18
[MRG]: Update and include `BatchSimulate` documentation

asoplata:
This PR does the following:

1. Promotes `BatchSimulate` to a top-level class befitting its status as a member of the public API. This means that its import is analogous to other public classes: just like `from hnn_core import Network`, you can now `from hnn_core import BatchSimulate` without having to import it strictly from its own module. This puts it on equal footing with the other classes described in the public API https://jonescompneurolab.github.io/hnn-core/stable/api.html . The example script for `BatchSimulate` has been updated to include this new import way.
2. Moves around the docstrings of `BatchSimulate` so they display properly on the sphinx build (i.e. the code website).
3. Adds a WARNING to the docstring of any `backend` options which indicate that only `loky` should be used right now (since the other backends do not work out of the box, see updates to #955).
4. Minor text polishing of docstrings.

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -25,8 +25,7 @@ jobs:
         shell: bash -el {0}
         run: |
           python -m pip install --upgrade pip
-          pip install numpy matplotlib scipy NEURON
-          pip install '.[gui, docs]'
+          pip install '.[docs, gui, parallel]'
 
       - name: Linkcheck
         shell: bash -el {0}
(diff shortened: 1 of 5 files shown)


asoplata:
@ntolley What do you think of the backend WARNING here: https://github.com/jonescompneurolab/hnn-core/pull/1019/commits/0889616dfce88802e1f26ba9ec5be5d572ad191d#diff-f2780322abf4ffe47adb1211d552f3809ea8e151a810b91230a42a945f08104aR198-R200 ?

dylansdaniels:
Reviewed and looks good. No code changes, just some reordering and improvements to the documentation. Will merge

PR #1022, opened 2025-03-24
[MRG] GUI layer-specific dipole axes

dylansdaniels:
Change template Layer-Specific Dipole (3x1) figure to consider all three subplots when determining the ylims of the axes

### DIFF
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -942,9 +942,9 @@ def _postprocess_template(template_name, fig, idx,
 
     if template_name == 'Dipole Layers (3x1)':
         # Make the L2 and L5 plots the same y-range
-        y_lims_list = [ax.get_ylim() for ax in fig.axes[0:2]]
+        y_lims_list = [ax.get_ylim() for ax in fig.axes]
         y_lims = (np.min(y_lims_list), np.max(y_lims_list))
-        [ax.set_ylim(y_lims) for ax in fig.axes[0:2]]
+        [ax.set_ylim(y_lims) for ax in fig.axes]
 
     # Re-render
     if not use_ipympl:
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -941,7 +941,7 @@ def _postprocess_template(template_name, fig, idx,
         return
 
     if template_name == 'Dipole Layers (3x1)':
-        # Make the L2 and L5 plots the same y-range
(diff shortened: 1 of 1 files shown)


asoplata:
Currently, this only improves the *initial* drawing of the templated figure, but not per-`axes`-object subsequent redraws. What I get currently with `master` is this:

Screenshot 2025-03-24 at 2 19 18 PM

This PR improves the initial figure drawing to use the same axes limits, as shown here:

image

However, I believe this is only solving the problem for the initial template draw, and not subsequent re-plotting of individual `axes`. Even with the current PR, if I `Clear axis` then re-draw `ax0` with `Add plot`, the y limits are re-calculated based only on this `axes` object, not all three:

image

I'm fine merging this as it is, since it makes the initial draw look better, but noting that AFAIK it doesn't fix the core issue for this figure. @dylansdaniels Does this issue still appear on your machine? Do you want to leave the work on fixing the individual `axes` redraw to later (v0.5) and merge this, or try to solve the problem in this PR? Up to you

dylansdaniels:
@asoplata I thought per our previous discussion we weren't going to explicitly handle the case of re-draws? This should always look correct when generating a figure from the template as you noted

hnn_core/gui/_viz_manager.py
```diff
@@ -942,9 +942,9 @@ def _postprocess_template(template_name, fig, idx,
 
     if template_name == 'Dipole Layers (3x1)':
         # Make the L2 and L5 plots the same y-range
```
gtdang, line 944: Perhaps update the comment to reflect the change?

dylansdaniels: Good catch! Thanks George @gtdang :)

asoplata: Thanks for taking a look George!

asoplata:
> To expand a bit, there are many things the user could do with a given subplot after generating a figure. They might replace one of the dipoles with a totally different type of figure. E.g., they might put spiking on the bottom where the 'agg' was displayed initially.
> 
> Putting a constraint on the axes bounds would require a lot of checks for these types of scenarios. I think we should guarantee the figure generates correctly from the template. But beyond that, the user has the freedom to change the figure to whatever they want if they hit the 'clear' button. We can add clarification about this in the documentation
> 
> @asoplata

Great point on not over-constraining subsequent axes draws. Also yeah, I forgot that we agreed to not address re-draws. Re-tested this and it still looks good. Merging.

PR #1024, opened 2025-03-25
[WIP]: Improve install instructions

asoplata:
@dylansdaniels Here is a very tentative series of changes we could make to the Install page so that Workshop users (and users in general) have a more straightforward way, with more explicit instructions, to install all available features on the 3 platforms (GUI+Opt on Windows, GUI+Opt+Parallel on MacOS and Linux). 

Note that this is a work in progress. The new "Comprehensive" installation methods add a lot of text, but I've run through them across all 3 platforms myself, and I think they're still simple and direct enough for people to use them in our Workshop.

However, the total lines of text on the page are IMHO too much now, to the point where the page as a whole suffers. I don't have time today, but I have some ideas about how we could trim the specific per-feature installs.

*edit: Due to the amount of investigation time on the install stuff, I started but didn't finish a comment on the Issue about conda builds. The short version is that 1. trying to build `conda` packages is probably not going to happen in time for the workshop, and 2. Much of the reason for 1 is that our possible `conda` install of HNN is going to require a lot of testing, since the conda build recipe system does NOT play well with `pip` (`conda` and `pip` are, indeed, two distinct package management ecosystems), and this is made more complex due to both the `conda` package build system and especially our NEURON dependency. I still plan on working towards eventually making hnn conda packages, but it needs more time to dance around the caveats, way more time than prep for the April workshop allows.

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -24,7 +24,8 @@ through the prompts):
 ```
 xcode-select --install
 ```
-3. Thirdly, If you already installed the [NEURON][] software system-wide using its traditional
+3. Thirdly, after installing "Xcode Command-Line Tools" above, restart your computer.
+4. Fourthly, If you already installed the [NEURON][] software system-wide using its traditional
 installer package, it is recommended to **remove it** first. We will be installing NEURON using
 its PyPI package alongside `hnn_core`. If you installed this using the default settings, then to
 uninstall it, you should delete the files located at `/Applications/NEURON`, and remove any
--- a/doc/install.md
+++ b/doc/install.md
@@ -5,59 +5,93 @@ If you have any questions or problems while installing `hnn_core`, feel free to
 ask for help [on our GitHub Discussions
 page](https://github.com/jonescompneurolab/hnn-core/discussions)!
 
-Please follow the instructions in Steps 1, 2, and 3, in that order.
+Please follow the instructions in Step 1 *before* proceeding to Step 2.
+
+Note that `hnn_core` currently only supports Python 3.8, 3.9, 3.10, 3.11, and 3.12, but *not* 3.13.
 
 --------------
 
-# Step 1. Python Environment
-
-We strongly recommend that you install `hnn_core` inside a "virtual
-environment" using software like the [Anaconda Python
-Distribution](https://www.anaconda.com/download/success). If you
-are new to Python or data science in Python, we recommend you review the
-resources here: <https://docs.anaconda.com/getting-started/>.
-
-Note that `hnn-core` currently only supports Python 3.8, 3.9, 3.10, 3.11, and 3.12, but *not* 3.13.
-
---------------
-
-# Step 2. Platform-specific requirements
+# Step 1. Platform-specific pre-requisites
 
 ## Windows
 
-- Before you install `hnn_core`, it is important that you **install** the
-  [NEURON][] software system-wide using its [Windows-specific
-  binaries](https://nrn.readthedocs.io/en/latest/install/install_instructions.html#windows). You
-  can test that NEURON was installed correctly by opening a Command Prompt (or
-  similar) via your Anaconda install, and running the following command:
-
-    ```
-    $ python -c "import neuron;"
-    ```
+- First, we **strongly* recommend you install the [Anaconda Python
+Distribution](https://www.anaconda.com/download/success). Use the default settings
+during the installation process. If you are new to Python, virtual environments in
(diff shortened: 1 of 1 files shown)


asoplata:
Also, note that both the location of the `environment-windows.yml` file and location of the `pip` install from github are all currently set for *testing* of these instructions. If we want to push these changes for v0.4, then these lines will need to be changed accordingly (i.e. the `environment-windows.yml` moved somewhere else or converted strictly into commands that the user is run, and the Github location of the hnn pip install changed to use the to-be-pushed v0.4 regular `pip` install).

doc/install.md
```diff
@@ -5,112 +5,225 @@ If you have any questions or problems while installing `hnn_core`, feel free to
...
+uninstall it, you should delete the files located at `/Applications/NEURON`, and remove any
+lines which mention NEURON in the files located at `~/.bashrc` and `~/.zshrc`.
+
+### Part 2: MacOS Comprehensive HNN Install steps:
+
+- Recommended for new users: To do a "comprehensive" install of HNN, including all features
+  available for your platform (for MacOS: API, GUI, Optimization, and Parallelism via both Joblib
+  and MPI), use the following instructions in this section.
+    - Alternatively, if you instead want a "minimal" installation with only specific features, see
+      the [Minimal Installation instructions](#minimal-installation-instructions) section below.
```
dylansdaniels, line 40: I think this anchor should be #minimal-installations

doc/install.md
```diff
@@ -5,112 +5,225 @@ If you have any questions or problems while installing `hnn_core`, feel free to
...
+
+1. Start the program called "Terminal". If you have recently installed Anaconda like in
+  Part 1, then make sure to start a **new** window of the "Terminal" program. You should
+  see `(base)` somewhere in text that is displayed.
```
dylansdaniels, line 44: somewhere in *the* text

doc/install.md
```diff
@@ -5,112 +5,225 @@ If you have any questions or problems while installing `hnn_core`, feel free to
...
+echo "export OLD_DYLD_FALLBACK_LIBRARY_PATH=\$DYLD_FALLBACK_LIBRARY_PATH" >> $CONDA_PREFIX/etc/conda/activate.d/env_vars.sh
+echo "export DYLD_FALLBACK_LIBRARY_PATH=\$DYLD_FALLBACK_LIBRARY_PATH:\${CONDA_PREFIX}/lib" >> $CONDA_PREFIX/etc/conda/activate.d/env_vars.sh
+echo "export DYLD_FALLBACK_LIBRARY_PATH=\$OLD_DYLD_FALLBACK_LIBRARY_PATH" >> $CONDA_PREFIX/etc/conda/deactivate.d/env_vars.sh
+echo "unset OLD_DYLD_FALLBACK_LIBRARY_PATH" >> $CONDA_PREFIX/etc/conda/deactivate.d/env_vars.sh
+export OLD_DYLD_FALLBACK_LIBRARY_PATH=$DYLD_FALLBACK_LIBRARY_PATH
+export DYLD_FALLBACK_LIBRARY_PATH=$LD_LIBRARY_PATH:${CONDA_PREFIX}/lib
+conda install -y -q "openmpi>5" mpi4py -c conda-forge
+pip install "hnn_core[gui,opt,parallel] @ git+https://github.com/jonescompneurolab/hnn-core.git"
+```
+- [TODO: after v0.4 publication, change the above to pip install "hnn_core[gui,opt,parallel]"]
```
dylansdaniels, line 60: already done?

doc/install.md
```diff
@@ -5,112 +5,225 @@ If you have any questions or problems while installing `hnn_core`, feel free to
...
+    - Alternatively, if you instead want a "minimal" installation with only specific features, see
+      the [Minimal Installation instructions](#minimal-installation-instructions) section below.
+
+1. Start the program called "Terminal". If you have recently installed Anaconda like in
+  Part 1, then make sure to start a **new** window of the "Terminal" program. You should
+  see `(base)` somewhere in text that is displayed.
+2. Copy and paste the following text directly into your Terminal:
+```
+conda activate base
+conda create -y -q -n hnn-core-env python=3.12
```
dylansdaniels, line 48: Check for an env with this name first, and delete if exists?

asoplata:
Delaying this past the v0.4 milestone release, since this content is subject to change according to the eventual attempts of building a conda package (see https://github.com/jonescompneurolab/hnn-core/issues/950 ).

asoplata:
This PR is deprecated by the work in #1051 . Closing.

PR #1027, opened 2025-03-27
[MRG] doc: draft v0.4 release notes, add missing api

asoplata:
As promised before, here are additions to the what's new page with proper, user-friendly release notes, in addition to updating all the PRs merged so far.

This is just a draft of course, and I am interested in if anyone else has writing changes they want to make, or especially if I left out describing some new feature that is included.

PR #1028, opened 2025-03-29
[WIP]: update testing files to use local assets

Freedisch:
Fix issue #936 
- Add downloaded files into the assets folders
- Updates testing files to use them


Freedisch:
Oh I see why the test CIs are falling it's because the downloaded `dpl.txt` has been added to `.gitignore` (I didn't noticed)

Freedisch:
So, since the `dpl.txt` seems to be a big file, and is currently added to .gitignore, I have reverted the `test_parallel_backends.py ` testing file

ntolley:
Ah it's a good point that `dpl.txt` takes up a lot of space!

@asoplata what do you think of saving a binarized format like `.npy` or `.hdf5`?

asoplata:
Re: `txt` files specifically, we may want to exclude them from the work in this PR for now, since it's not clear what data format we may want them to eventually take (see https://github.com/jonescompneurolab/hnn-core/pull/1027#issuecomment-2766506145 ). For example, if we eventually (this year?) plan on phasing out `txt` read/write support, then we may not want to include the files in `assets` (slightly contradicting what I wrote in #936)

PR #1030, opened 2025-03-31
[MRG] fix: expand custom x-axis to psd layer plots

### LINKED ISSUE #1023: GUI: Bug: `PSD layer` x-axes limits are set by simulation time, not frequencies
There is currently a bug with the GUI's `PSD Layers` plot template, which is probably due to this template inheriting code from the `Dipole Layers` template. The minimum and maximum of the x-axes shown in the layer-specific, but not aggregate, PSDs of the plot appear to be derived from the simulation time parameters, NOT the frequency ranges as they should. Here are two examples of simulations where the `PSD Layers` plot is shown, and the only difference between them were simulation name and the length of simulation time. It is pretty obvious that x-axis maximum is set to, and changes similarly to, the `tstop`:

Image

second example:

Image

asoplata:
Resolves #1023 , fixing the PSD layer GUI plots' x-axes.

### DIFF
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -442,8 +442,14 @@ def _update_ax(fig, ax, single_simulation, sim_name, plot_type, plot_config):
                 io_buf.close()
                 _ = ax.imshow(img_arr)
 
-    # set up alignment
-    if plot_type not in ['network', 'PSD']:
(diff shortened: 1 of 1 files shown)


dylansdaniels:
Tested on my end - works with the PSD layers template, blank figs, axes updates. I see no issues

PR #1034, opened 2025-04-01
maint: patch: v0.4

asoplata:
This is an emergency patch due to several issues that our tests did not catch. This will become the new v0.4 release commit.

This commit does two things:
1. Reverts `BatchSimulate` from a top-level import to how it was imported previously, which now requires manual import, as shown in its example. I've added relevant instructions in the example that indicates it requires some extra dependencies to function.
2. Adds a line in `setup.py` which is required for `setup.py` use (but not `pyproject.toml`) that indicates to the relevant Python Packaging repository that the README is in markdown format. I've also be re-writing the "How to make a release" guide a bunch, including reversing the order of instructings for `twine` uploading vs testing the distributable.

It turns out that the reason `BatchSimulate` was not added to the existing API documentation is related to a "blind spot" in our tests: neither our units tests nor documentation tests actually test a "minimal install" of simply:
`pip install hnn_core`
*without* including various extras. It turns that doing the above command in a fresh environment of the previous master branch is actually broken: this is due to `BatchSimulate`'s import dependence on `joblib`, which is only included in the `[parallel]` dependencies. Because I recently added `BatchSimulate` as a top-level import, but all of our tests automatically install the `joblib` dependency, this was not caught. I apologize for the error, and I will make a future PR that adds additional steps to our unit tests that *progressively* install the dependencies of minimal and, later, extra dependencies of hnn-core, before proceeding to the existing unit tests.

Normally, it would be best to propagate this change as a "patch" release such as v0.4.1, but we currently do not follow Semantic Versioning, and furthermore it has yet to be tested if our documentation systems works with Semantic Versioning indexing (or how it can be made to be). This is another change we should make soon in the future.

Depending on the tests and more install testing, this will likely become the new v0.4 release commit because the Python-packaged version had not yet been pushed to Pypi (immutable), while other metadata (such as the documentation and git tags) are mutable.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -24,7 +24,6 @@ Simulation (:py:mod:`hnn_core`):
    Cell
    CellResponse
    pick_connection
-   BatchSimulate
 
 Network Models (:py:mod:`hnn_core`):
 ------------------------------------
@@ -100,7 +99,6 @@ Parallel backends (:py:mod:`hnn_core.parallel_backends`):
 
    MPIBackend
    JoblibBackend
-
 
 Input and Output:
 -----------------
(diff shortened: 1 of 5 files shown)


asoplata:
btw, you can test the fact that the default install for master currently is broken by the joblib dependency using the following two commands:

```
pip install git+https://github.com/jonescompneurolab/hnn-core.git@9e14b998836a8202777486093d62ec2de31793e2
python -c "import hnn_core"
```

This has probably been broken ever since I merged in the `BatchSimulate` import change.

PR #1043, opened 2025-04-24
[ENH]: Update Linkcheck Schedule to Reduce Frequency

### LINKED ISSUE #1031: MAINT: Decrease frequency of running Linkcheck
Currently, Linkcheck ( https://github.com/jonescompneurolab/hnn-core/blob/master/.github/workflows/linkcheck.yml ), which checks that all the external URLs of our documentation resolve correctly, runs every time that PRs are tested (frequently), and on every push to `master`.

However, as our documentation increases (especially via references to Github in our `whats_new.md`), we will gradually be increasing the number of external links that need to be checked. Recently, in my personal experience, Linkcheck has been failing more frequently than in the past. The vast majority of Linkcheck failures are due to external URL uptime, *not* due to mistakes on our part. I am frequently seeing 502 gateway issues or unavailable URLs from Github, CircleCI, Codecov (which AFAIK is not currently working due to config updates on our part), and Brown's HNN servers. Very frequently, a manual re-run of the Linkcheck job is needed to get the tests to pass. An example failure can be found here: https://github.com/jonescompneurolab/hnn-core/actions/runs/14178037937/job/39717517721 .

Similar to changes to testing frequency we have discussed at the dev meetings, I think we may want to consider reducing the frequency of Linkcheck, since most code does not change the public API or external links, and I think running Linkcheck on every PR is unnecessary. Instead, I propose the following:

- Change Linkcheck to only run:
    1. On every push/merge to `master` (like currently)
    2. NOT on every PR (including subsequent updates)
    3. On a half-month basis

Myrausman:
### Summary
This pull request updates the Linkcheck workflow to modify its frequency and reduce unnecessary runs.
### Changes Made
1. The workflow now:
   - Runs only on every push/merge to the `master` branch.
   - Executes on a scheduled basis on the 1st and 15th of each month at midnight UTC.
   - No longer runs on every pull request (including subsequent updates).

Closes #1031

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -2,9 +2,10 @@ name: Linkcheck
 
 on:
   push:
-    branches: ['**']
-  pull_request:
-    branches: ['**']
+    branches:
+      - master
+  schedule:
+    - cron: "0 0 1,15 * *"
 
 jobs:
   build:
--- a/README.md
+++ b/README.md
@@ -118,7 +118,4 @@ JOSS](https://doi.org/10.21105/joss.05848):
 > Software*, 8(92), 5848, <https://doi.org/10.21105/joss.05848>
 
 [Contributing Guide]: https://jonescompneurolab.github.io/hnn-core/stable/contributing.html
-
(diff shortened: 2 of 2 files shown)


asoplata:
Thanks Maira!

PR #1045, opened 2025-04-24
Simplify rename_cell_type

jasmainak:
Follow up from #1044 ... hopefully the correct branches now. Tests seem to pass locally

### DIFF
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -124,6 +124,37 @@ def _compare_lists(s, t):
     return not t
 
 
+def _update_dict_keys(input_dict, old_key, new_key):
+    """Recursively update keys and values in a dictionary.
+
+    Parameters
+    ----------
+    input_dict : dict
+        The dictionary to update.
+    old_key : str
+        The key or value to replace.
+    new_key : str
+        The new key or value to replace with.
+
+    Returns
+    -------
+    updated_dict : dict
+        The updated dictionary with keys and values replaced.
+    """
+    updated_dict = dict()
+    for key, value in input_dict.items():
+        # Update the key if it matches old_key
+        new_key_name = new_key if key == old_key else key
+        # Recursively update values if they are dictionaries
+        if isinstance(value, dict):
+            updated_dict[new_key_name] = _update_dict_keys(
+                value, old_key, new_key)
+        else:
+            # Update the value if it matches old_key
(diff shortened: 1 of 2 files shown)


hnn_core/network.py
```diff
@@ -124,6 +124,37 @@ def _compare_lists(s, t):
     return not t
 
 
+def _update_dict_keys(input_dict, old_key, new_key):
```
jasmainak, line 127: should go into `utils` as you suggested @asoplata

hnn_core/network.py
```diff
@@ -124,6 +124,37 @@ def _compare_lists(s, t):
...
+    updated_dict = dict()
+    for key, value in input_dict.items():
+        # Update the key if it matches old_key
+        new_key_name = new_key if key == old_key else key
+        # Recursively update values if they are dictionaries
+        if isinstance(value, dict):
+            updated_dict[new_key_name] = _update_dict_keys(
+                value, old_key, new_key)
+        else:
+            # Update the value if it matches old_key
```
jasmainak, line 153:
```suggestion
```

superfluous comment

hnn_core/network.py
```diff
@@ -124,6 +124,37 @@ def _compare_lists(s, t):
...
+    new_key : str
+        The new key or value to replace with.
+
+    Returns
+    -------
+    updated_dict : dict
+        The updated dictionary with keys and values replaced.
+    """
+    updated_dict = dict()
+    for key, value in input_dict.items():
```
jasmainak, line 145: note after python 3.7 or 3.8 (don't remember the exact version) this will preserve order

asoplata:
Much more elegant! I will double check that it works for the sensitive `gid_ranges` etc., then move the recursion into utils, then merge into the `refactor...` branch.

asoplata:
@jasmainak I pushed some commits to your branch here. Updated some documentations, moved the function to utils, and most importantly changed the behavior of the replace function so that it also performs the replacement for any *list* elements inside the top-level dictionary. The last part was necessary because there were still replacements not being performed that should have been, as documented in the final commit message.

If you are okay with these changes, then please let me know and I will squash+merge this PR.

asoplata:
Note: Linkcheck fails can be ignored, since these branches were based off of old `master` versions that include out-of-date documentation links (which have been replaced since).

jasmainak:
Looks good. Please feel free to merge!

PR #1046, opened 2025-04-25
[MRG] doc: add cloud install instructions

asoplata:
This adds a section to the beginning of the Install page with information on how to use HNN through both NSG and Google CoLab. This is related to our grant milestones of improving ease of use for NSG.

I have explicitly added the version note about HNN v0.3 on NSG, but I am in contact with NSG about them updating the version. Hopefully we should be able to remove that note soon.

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -1,4 +1,23 @@
 (install)=
+# Using HNN in the cloud (no installation necessary)
+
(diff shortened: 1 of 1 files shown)


dylansdaniels:
Looks good to me

PR #1048, opened 2025-05-06
[MRG] Replace "HNN-cloud usage" instructions with link

asoplata:
As per discussion with @dylansdaniels  recently, this reverses the addition of content to the Install page that describes how to run HNN in the cloud without a local install.

However, instead of removing it entirely, this still leaves in a link to the Cloud guidance on the current Textbook website. I think this is the best solution for two reasons:

1. Any user who is going to the Code-website's install page may be interested in test-driving HNN in the cloud before they install it, and it is good that they are made aware that *that is possible*. Frequently, for software packages, some users (such as myself) tend to learn about a software package, then immediately look up the Github repo before reviewing the non-code website for a package. In this case, all users reviewing any installation pages (either on our Code website or Textbook website) will be shown that it is possible to run HNN in the cloud before installation.

2. This also obeys the "once and only once" guideline, as local installation instructions should only be on the Code website's install page, while Cloud usage should only be on the Textbook's Install page.

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -1,28 +1,11 @@
 (install)=
-# Using HNN in the cloud (no installation necessary)
-
-If you want to try out HNN *without installing it locally* on your computer first, there are several methods:
-
-### NSG cloud method
-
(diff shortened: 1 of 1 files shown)


dylansdaniels:
This works for me. I'll merge

PR #1051, opened 2025-05-21
[MRG] DOC: Add `conda` install instructions

asoplata:
This is a major revamp of our Installation page. As per #950, this adds a new section at the top of the install page for recommending and describing how a user should install HNN-Core using our newly-published `conda` packages.

Observers will note that the conda package install path allows for MPI etc. usage, but is **significantly** easier of an install. I (and a few others) have tested the `conda` install method on all listed platforms, including MPI support on a Windows WSL install.

Because we still want to retain `pip` install path instructions and descriptions, I have left them in the Install page but tried to shorten them where I can. In particular:
- The description of and usage guide for both Joblib and MPI have been moved to our new Textbook site, using the [new tutorial here for Joblib](https://dylansdaniels.github.io/website_redesign/content/04_using_hnn/parallelism_joblib.html) and a [renamed tutorial here](https://dylansdaniels.github.io/website_redesign/content/04_using_hnn/parallelism_mpi.html) with slightly more content on MPI usage.
- Most of the other examples/guides suggested after a particular installation type now link against our Textbook website.

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -12,11 +12,16 @@ Note: If you are at Brown University and trying to install HNN-Core on the OSCAR
 
 # `conda` Installation
 
-We *strongly* recommend that you install HNN-Core from our Anaconda package instead of from `pip`. This `conda` package includes all features (GUI, Optimization, and both Parallelism features), a working installation of OpenMPI, and is significantly easier to install. Follow the below instructions for your operating system.
+We recommend that you install HNN-Core from our Anaconda package instead of from `pip`. This `conda` package includes all features (GUI, Optimization, and both Parallelism features), a working installation of OpenMPI, and is significantly easier to install.
 
-The only exception is if you are using a High-Performance Computing (HPC) environment (also called a computing cluster or supercomputer). The `conda` installation may work on your HPC, but if you want to use your own MPI libraries, you can try our [`pip` MPI method](#pip-mpi-macos-or-linux), and you can always ask us for help at [our Github Discussions page][].
+The only exceptions are if:
+
+1. You want to install HNN-Core alongside MNE. In that case, we recommend you use the one of the below [`pip` Installation](#pip-installation) methods.
+2. You are using a High-Performance Computing (HPC) environment (also called a computing cluster or supercomputer). The `conda` installation may work on your HPC, but if you want to use your own MPI libraries, you can try our [`pip` MPI method](#pip-mpi-macos-or-linux), and you can always ask us for help at [our Github Discussions page][].
 
 If you want a minimal install of only the HNN-Core API using `conda`, then follow the instructions below, but replace the package name `hnn-core-all` with `hnn-core`.
+
+Follow the below instructions for your operating system.
 
 ## `conda` MacOS or Linux
 
@@ -89,7 +94,7 @@ If you see output similar to `WARNING: Your machine hardware does not appear to 
 
 # `pip` Installation
 
-Alternatively, you can install HNN-Core through `pip` if you want to contribute to development, only want certain features, or have issues with the `conda` install method. If using the `pip` method, please follow the instructions in Steps 1, 2, and 3, in that order. Again, we recommend the above `conda` method over the below `pip` methods.
+Alternatively, you can install HNN-Core through `pip` if you want install it alongside MNE, to contribute to development, only want certain features, or have issues with the `conda` install method. If using the `pip` method, please follow the instructions in Steps 1, 2, and 3, in that order.
 
 ## Step 1. `pip` Python Environment
 
--- a/doc/install.md
+++ b/doc/install.md
@@ -6,6 +6,8 @@ ask for help on [our GitHub Discussions page][]!
 
 If you wish to try out HNN-Core in the cloud without installing it locally, see the [Installation page at the HNN Textbook](https://dylansdaniels.github.io/website_redesign/content/01_getting_started/installation.html).
 
+Note: If you are at Brown University and trying to install HNN-Core on the OSCAR system, you should **not** follow these instructions, and instead follow the [instructions here](https://github.com/jonescompneurolab/oscar-install).
+
 --------------
 
 # `conda` Installation
--- a/doc/install.md
+++ b/doc/install.md
@@ -194,7 +194,7 @@ Many of the examples on [our HNN Textbook website][] make use of Joblib support,
 
 If you want to use MPI, we recommend you first try to install our `conda` package detailed [at the top of the page](#conda-installation). Otherwise, we **strongly** recommend you use the [Anaconda Python Distribution](https://www.anaconda.com/download/success) specifically, since Anaconda is the easiest way to download [OpenMPI](https://anaconda.org/conda-forge/openmpi) binaries, which are *not* Python code. If you need to use your own specific MPI binaries/libraries (such as if you are installing on an HPC platform) and you need help, then let us know on [our Github Discussions page][].
 
-To install HNN-Core with its MPI dependendencies using the `pip` method, follow the below instructions for your operating system.
+To install HNN-Core with its MPI dependencies using the `pip` method, follow the below instructions for your operating system.
 
 #### `pip` MPI: Windows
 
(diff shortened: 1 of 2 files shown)


asoplata:
A few points:

1. This does not remove any existing install functionality or instructions: all of the same instructions on how to install all `pip` variants, including with MPI, are still present and valid. This only adds a new, recommended install pathway.

2. I can definitely add some text that suggests that users with MNE installed should use the traditional `pip` installation methods, since those have been the only install path for years and are much less likely to conflict with MNE. I will do that in a new commit.

3. I will make a separate issue about better MNE co-installation support, since it is nontrivial and important, but I don't think it should block this PR.
    - If we want to actively support users installing MNE and HNN in the same environment/workflow, then *there should be clear instructions* on how to do that, and at least one method that we frequently test and do our best to make sure it works. Currently, the only mention of a co-installation with MNE is in <https://jonescompneurolab.github.io/hnn-core/stable/auto_examples/workflows/plot_simulate_somato.html#sphx-glr-auto-examples-workflows-plot-simulate-somato-py>, but that makes a lot of assumptions, including that it only installs MNE with its "core dependencies" (<https://mne.tools/dev/install/manual_install.html#manual-install>), the user does not already have MNE installed, the user knows where to pip install, etc.
    - Ideally, any co-installation pathway should support both MNE "with all dependencies" (i.e. installer and conda) and also HNN with at least optimization and MPI. This is important because researchers who already have MNE or are installing it will also benefit more from the most advanced HNN usage. However, since Anaconda is the only cross-platform solution for distributing the MPI binaries we need, this will take some investigation and testing of `conda` compabilities.
    - I will mention that I support this very much: I think one of the most important development goals for HNN right now should be better MNE integration, especially when it comes to guidance and documentation about how researchers exchange data between the two programs. Co-installation is an important part of expanding our support for such workflows, and I think should be a priority for me this summer.

asoplata:
Yeah I've been testing MNE/HNN co-installs and like you suggested, installing the MNE `conda`, then `pip` HNN inside, is almost certainly the best path. I still need to test some edge cases like system-wide MNE installers vs conda, such as: if there are any incompatibilities with the `conda` MPI packages that HNN's parallel needs (probably not), etc. but this path is probably going to be the one that I eventually recommend. I will make a separate issue about that, however, where I document my investigation. I'll add that we also have more people (i.e. me ;) ) who can fully dedicate their time if a user reaches out and needs help with an MNE/HNN co-install.

Re: cautioning the virtual environment: definitely agree that caution is warranted, and that goes doubly for `conda` use specifically. I thought you were saying we should not push users to use conda but that was a misinterpretation on my part. Trust me, I also don't like having to rely on conda either, especially the fact that our MPI usage requires not just the `defaults` conda packages but also the less-stable `conda-forge`. However, 1. it's the the best way for our users to get MPI installed on their computers (which probably never be possible with `pip` alone), and 2. in the lab we've agreed that the speedup MPI offers is very important for making a good first impression on users (single-core sims are pretty slow, even for the longer sims we do in our Workshops), and therefore needs to be as simple as possible to install so that new users can take advantage.

doc/install.md
```diff
@@ -1,292 +1,283 @@
 (install)=
 # Installation Guide
 
-If you have any questions or problems while installing `hnn_core`, feel free to
-ask for help [on our GitHub Discussions
-page](https://github.com/jonescompneurolab/hnn-core/discussions)!
+If you have any questions or problems while installing HNN-Core, feel free to
+ask for help on [our GitHub Discussions page][]!
```
dylansdaniels, line 5: Is there a reason the link () was removed?

doc/install.md
```diff
@@ -1,292 +1,283 @@
...
-  will be installing NEURON using its PyPI package alongside `hnn_core`.
-- Before you install `hnn_core`, it is important that you install "Xcode
-  Command-Line Tools".  This can be done easily by opening a Terminal and
-  running the following command (followed by clicking through the prompts):
-    ```
-    $ xcode-select --install
-    ```
+## Step 2. `pip` Platform-specific requirements
+
+### `pip` on Windows
```
dylansdaniels, line 107: Above we do MacOS/Linux before Windows, might want to flip the ordering here for consistency

doc/install.md
```diff
@@ -1,292 +1,283 @@
...
+
+### `pip` on MacOS
 
 - If you already installed the [NEURON][] software system-wide using its
   traditional installer package, it is recommended to **remove it** first. We
-  will be installing NEURON using its PyPI package alongside `hnn_core`.
-
---------------
+  will be installing NEURON using its PyPI package.
+- Before you install HNN-Core, it is important that you install "Xcode Command-Line Tools". This can be done easily by opening a Terminal, running the following command, and then clicking through the prompts in the window that will pop up. Note that you must restart your computer after Xcode Command-Line Tools has finished installing!
```
dylansdaniels, line 120: Maybe add a check for whether they have Xcode CLI first?

doc/install.md
asoplata: It wasn't removed, but rather this uses the definition of the link at the bottom of the page (see here https://github.com/jonescompneurolab/hnn-core/pull/1051/files#diff-7ab5dff47ed5423c876901a792bf21a93da0dc8713f068bf828091824606579fR281-R283 ). In the new version of `install.md`, in the final lines of the file, there are reference definitions for this link content. Notice how the link here in line 5 ends in brackets, NOT parens e.g. `[asdf][]` as opposed to `[asdf]()`. This is using markdown's footnotes https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#footnotes since we re-use that same link frequently on the page.

doc/install.md
```diff
@@ -1,292 +1,283 @@
...
+wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
+bash Miniconda3-latest-Linux-x86_64.sh
+```
+
+If you see output similar to `WARNING: Your machine hardware does not appear to be x86_64`, then please contact us via [our Github Discussions page][]. You may be able to install by using <https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-aarch64.sh> instead.
+
+7. Close the Ubuntu window, then open a new one. Your prompt should now show `(base)` at the beginning.
+8. Inside this window, follow the steps above in the [`conda` MacOS or Linux](#conda-macos-or-linux) section, then return here.
+9. HNN-Core should now be installed!
+10. Note: On Windows, every time you start the GUI, you will need to navigate to <http://localhost:8866> in your browser (or refresh if you are already on the page).
```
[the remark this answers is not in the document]
asoplata: Yes that is correct; when I tested it, it did not seem to automatically open a browser tab. Probably due to that we're running the server in linux but there isn't XDG/etc. connections to Windows MIME types like browser application tab-opening, but I'm not sure.

doc/install.md
```diff
@@ -1,292 +1,283 @@
...
+include: [ipykernel](https://ipykernel.readthedocs.io/en/stable/), [ipympl](https://matplotlib.org/ipympl/), [ipywidgets](https://github.com/jupyter-widgets/ipywidgets) >=8.0.0, and [voila](https://github.com/voila-dashboards/voila).
 
-- [ipykernel](https://ipykernel.readthedocs.io/en/stable/)
-- [ipympl](https://matplotlib.org/ipympl/)
-- [ipywidgets](https://github.com/jupyter-widgets/ipywidgets) >=8.0.0
-- [voila](https://github.com/voila-dashboards/voila)
+To start the GUI, simply run:
 
-**GUI Usage**: To start the GUI, simply run:
+```
```
dylansdaniels, line 166: We do recommend they use an environment in Step 1, so maybe we wan to call out that *if* they follow that piece of the advice, they'll need to activate their environment before running `hnn-gui`

doc/install.md
dylansdaniels: nvm i see what you did below now

doc/install.md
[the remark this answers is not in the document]
dylansdaniels: Just wanted to make sure I understood correctly 👍

doc/install.md
dylansdaniels: All looks good to me

PR #1055, opened 2025-06-02
[MRG] doc: trivial update for textbook links

asoplata:


asoplata:
I am going to go ahead and merge this myself since 1. it's a trivial change, 2. it only affects docs of a single page, and 3. without it, all Linkcheck tests will fail.

PR #1057, opened 2025-06-05
Allow user-defined initial weights and average dipoles in obj func

carolinafernandezp:
Allow user-defined initial weights, update objective function to average 3 dipoles rather than simulate a single one, enhance optimization example.

### DIFF
--- a/examples/howto/optimize_evoked.py
+++ b/examples/howto/optimize_evoked.py
@@ -12,16 +12,24 @@ of the model simulation to match an experimental dipole waveform.
 #          Ryan Thorpe <ryan_thorpe@brown.edu>
 #          Mainak Jas <mjas@mgh.harvard.edu>
 
+from hnn_core.optimization import Optimizer
+from urllib.request import urlretrieve
 import os.path as op
 
 import matplotlib.pyplot as plt
+from matplotlib.lines import Line2D
 
 ###############################################################################
-# Let us import hnn_core
+# 1. Import libraries and set up environment
+# ------------------------------------------
+#
+# We'll start by importing the necessary `hnn_core` modules and other libraries
+# required for this tutorial.
 
 import hnn_core
 from hnn_core import (MPIBackend, jones_2009_model, simulate_dipole,
                       read_dipole)
+from hnn_core.viz import plot_dipole
 
 hnn_core_root = op.join(op.dirname(hnn_core.__file__))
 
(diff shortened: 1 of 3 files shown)


examples/howto/optimize_evoked.py
```diff
@@ -12,203 +12,278 @@
...
+                             sigma=1.,
+                             numspikes=1,
+                             location='proximal',
+                             weights_ampa=weights_ampa_p2,
+                             synaptic_delays=synaptic_delays_p,
+                             n_drive_cells=n_drive_cells,
+                             cell_specific=cell_specific)
+
+# Simulate initial dipole
+n_trials = 10
```
asoplata, line 121: Hey Carolina, I definitely like the new trials support! However, if my understanding is correct, this example now requires 10x the amount of time to run that it did before. Prior to this branch, this notebook tended in the past to take about 20 minutes to run, so if this will take 10x as long, that's too long for an example notebook. Would the example still make sense if you decreased the number of trials here, OR used a smaller `max_iter` argument later on in the `Optimizer` call?

hnn_core/optimization/general_optimization.py
```diff
@@ -288,8 +306,8 @@ def _run_opt_bayesian(
     tstop,
     constraints,
     set_params,
-    obj_fun,
     initial_params,
+    obj_fun,
     max_iter,
     obj_fun_kwargs,
```
asoplata, line 312: Can you please add a docstring entry for `obj_fun_kwargs` in both `_run_opt_bayesian` and `_run_opt_cobyla`? You could use the following:

```
    obj_fun_kwargs : dict
        Additional arguments, along with their associated values, to be passed to the
        final optimization function.

```

examples/howto/optimize_evoked.py
```diff
@@ -12,203 +12,278 @@
...
+            show=False, color='tab:green', average=True)
+
+# Legend
+legend_handles = [Line2D([0], [0], color='tab:blue', lw=1.0),
+                  Line2D([0], [0], color='tab:orange', lw=1.0),
+                  Line2D([0], [0], color='tab:green', lw=1.0)]
+ax.legend(legend_handles, ['experimental', 'initial', 'optimized'])
+
+# Convergence plot
+optim.plot_convergence()
```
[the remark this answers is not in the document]
asoplata: You may want to add this line to `optimize_rhythmic.py` as well.

hnn_core/optimization/objective_functions.py
```diff
@@ -126,22 +129,27 @@ def _maximize_psd(
...
 
     # for each f band
     f_bands_psds = list()
+    relative_bandpower = obj_fun_kwargs["relative_bandpower"]
+
+    # Convert tuple to list
+    if isinstance(relative_bandpower, tuple):
+        relative_bandpower = list(relative_bandpower)
+    # If a single number is intended to be applied to all bands
+    if not isinstance(relative_bandpower, (list, np.ndarray)):
```
asoplata, line 145: I would make 2 changes here:

1. Turn `if` into `elif` since tuple is not in this list
2. change the `isinstance` check to compare `relative_bandpower` to all `Iterable` types. You could do this via the following:

Add the following import to the top of the file
```
from collections.abc import Iterable
```

Then, change this line to 
`    if not isinstance(relative_bandpower, Iterable):`

asoplata: It may also work to do the inverse and use something like the following: `if (isinstance(rela..., float)) or (isinstance(rela..., int))`

asoplata: If it's already necessary to convert `relative_bandpower` from a tuple to a list, then why not changed the expected type of it to be a list in the first place? This would simplify its usage and the description. In this case, the docstring could look something like this (assuming I'm understanding everything correctly):

```
    relative_bandpower : list of float | float
        Weight for each frequency band. If a list is provided, the relative weights should correspond to each frequency band in `f_bands`. If a float is provided, each frequency band in `f_bands` has equal weight.
```

examples/howto/optimize_evoked.py
ntolley: Yeah I think `n_trials` could remain at 1 for the notebook, but perhaps @carolinafernandezp you could add a comment that real experimental ERPs are an average over trials and the simulations can be averaged to reflect that

examples/howto/optimize_evoked.py
```diff
@@ -12,203 +12,278 @@
...
 
 
 def set_params(net, params):
 
     # Proximal 1
     net.add_evoked_drive('evprox1',
-                         mu=5.6813,
-                         sigma=20.3969,
+                         mu=26.61,
+                         sigma=2.47,
```
ntolley, line 167: If `cell_specific=False` and `n_drive_cells=1` then I think the variance only comes into play when there are multiple trials I think

If you are using 1 trial based on the earlier feedback perhaps set this to 0.0

examples/howto/optimize_evoked.py
```diff
@@ -12,203 +12,278 @@
...
+# --------------------------------
+#
+# For optimal results, set initial_params to your hand-tuned values since they
+# already provide a good fit.
+#
+# If initial_params is not set, the optimizer will use the midpoint of the
+# constraints as the initial parameters.
 
-from hnn_core.optimization import Optimizer
+initial_params = dict({'evprox2_ampa_L5_pyramidal': 0.1,
```
ntolley, line 227: I'm a little confused how this gets used in the `run_opt()` functions. I see that they are passed as arguments to the objective functions, but I don't see any new changes that really use `initial_params`

Was `initial_params` a variable used internally and now this is just exposing the ability to set the value?

hnn_core/optimization/general_optimization.py
```diff
@@ -101,6 +108,11 @@ def __init__(
         else:
             self.obj_fun = obj_fun  # user-defined function
             self.obj_fun_name = None
+        # Initial weights for the objective function
+        if initial_params is not None:
+            self.initial_params = initial_params
```
ntolley, line 113: Are there any type checks on the format of initial params that get triggered later on?

One good unit test would be to check if there's a mismatch between the keys of `initial_params` and `constraints`

examples/howto/optimize_evoked.py
```diff
@@ -12,203 +12,278 @@
...
-                    'evprox2_ampa_L5_basket': (0.01, 1.),
-                    'evprox2_ampa_L5_pyramidal': (0.01, 1.),
-                    'evprox2_nmda_L2_basket': (0.01, 1.),
-                    'evprox2_nmda_L2_pyramidal': (0.01, 1.),
-                    'evprox2_nmda_L5_basket': (0.01, 1.),
-                    'evprox2_nmda_L5_pyramidal': (0.01, 1.),
-                    'evprox2_mu': (100., 120.),
-                    'evprox2_sigma': (2., 30.)})
+lower_bound = 0.1 * 10**-1
+upper_bound = 0.1 * 10
```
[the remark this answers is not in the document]
ntolley:
```suggestion
lower_bound = -2
upper_bound = 0
```
Alternatively it may be better that we recommend people use a log-scale for synaptic weights. I have found this reparameterization to work much better for optimizers since the relationship between synaptic conductance and membrane potential is non-linear

I'll comment above how to update the parameters appropriately for a log-scale prior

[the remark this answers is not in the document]
carolinafernandezp: The only reason I kept it that way was to make the point that it's better / more meaningful to think about setting constraints using a scaling factor approach instead of hard-coding arbitrary boundaries like 0.01 and 1.0 (given that the weights aren't direct biological values). What I could do is add a comment describing that reasoning and go with your 0.01 - 1.0 suggestion.

carolinafernandezp:
Thank you @asoplata and @ntolley for all the feedback! I incorporated your comments. 

Please see the example runs below for comparison (`n_procs` was set to 10). 

<table style="width:100%;">
  <caption><h4>n_trials=10, max_iter=200, ~59 minutes</h4></caption>
  <thead>
    <tr>
      <td style="width:50%; text-align:center;">
        optimization_dipoles
      </td>
      <td style="width:50%; text-align:center;">
        convergence_plot
      </td>
    </tr>
  </thead>
</table>

<table style="width:100%;">
  <caption><h4>n_trials=3, max_iter=50, ~5 minutes</h4></caption>
  <tbody>
    <tr>
      <td style="width:50%; text-align:center;">
        optimization_dipoles_50
      </td>
      <td style="width:50%; text-align:center;">
        convergence_plot_50
      </td>
    </tr>
  </tbody>
</table>

hnn_core/optimization/general_optimization.py
```diff
@@ -101,6 +108,21 @@ def __init__(
         else:
             self.obj_fun = obj_fun  # user-defined function
             self.obj_fun_name = None
+        # Initial weights for the objective function
+        if initial_params is not None:
+            # Check if initial_params is a dict
+            if not isinstance(initial_params, dict):
```
ntolley, line 114: My only other comment would be to add a unit test for this and then it's good to go!

ntolley:
```suggestion
            from .externals.mne import _validate_type  # this import would be moved to top
                    _validate_type(initial_params, types=(dict,), item_name="initial_params")
```
Also instead of using `isinstance()` you can use the `check_type()` function, this is so that there is consistency in the error messages

ntolley (commented):
This looks great @carolinafernandezp!

There's some edge cases that I think need unit tests, otherwise the code is good to go, here's the suggestions:

1) `initial_params` is a not a `dict` (I made a suggestion on a function to test this
2) `initial_params` keys do not match keys in `constraints`
3) `initial_params` values contain a  non `float` or `int`
4) `initial_params` values are outside the range of `constraints`

carolinafernandezp:
Thank you @ntolley, tests have been added!

asoplata:
Yes, please add mention of your work to the `Public API Changes` and `Changelog` entries in the latest change notes, using the file at `doc/whats_new.md`. Note that the file now uses Markdown formatting by default, but with support for ReST now. You can see more guidance for authoring here https://jonescompneurolab.github.io/hnn-core/dev/contributing.html#updating-the-developer-portal

asoplata:
@ntolley I think you should be the one to "Squash and merge" this PR when it's ready for merge

carolinafernandezp:
@asoplata some tests in `test_parallel_backends.py` are failing when `run_hnn_core_fixture` instantiates `MPIBackend()`  with `n_procs=3` (['mpiexec', '--use-hwthread-cpus', '-np', '3', 'nrniv', '-python', ...]). It looks like the number of available physical cores for ubuntu-latest (for public repos) is 4 so I'm not sure why this is failing but I also don't know much about how this works.

ntolley:
Thanks for implementing these changes @carolinafernandezp! All is looking clean, going to run the notebook locally for one last check.

@asoplata should I wait on anything for your hotfix before merging this?

ntolley:
Unfortunately we may need to rebase this PR to get those other tests to pass

@carolinafernandezp there was NEURON update that breaks a lot of our code, @asoplata added a hotfix to cap the NEURON version, but since this PR was opening before that hotfix the unit test runners are installing the new version of NEURON and failing

katduecker:
Hi all,

I just rebased my 1-yo local repo to master and encountered an issue with the optimisation, which I believe is based on a change made in this PR. Before, the user did not have to define n_trials, but now, an error is raised when the user doesn't.
Just at first glance I am confused where in my optimisation I would define n_trials, and I would also think it may be useful to set a default here to 1 so that this doesn't happen to other veteran users?

It's a great option but we might run into backwards compatibility issues if there is no default.

@carolinafernandezp @asoplata @ntolley 

Thanks!

PR #1067, opened 2025-06-13
doc: fix textbook usage links after url update

asoplata:
linkcheck, oh linkcheck
how thou frustrates me


asoplata:
Since this is a tiny documentation-only fix, I will merge it without review.

PR #1068, opened 2025-06-13
[MRG] Add explicit base net regen

asoplata:
This adds a new script, `dev_scripts/regenerate_base_network.py`, for doing what its name implies, and adds accompanying documentation in the Contributing Guide. This also makes some small improvements to the Contributing Guide in general.

As far as I could tell, this fixes an undocumented "missing documentation" issue: it was not explicitly clear how the default network file for the GUI (`hnn_core/param/jones2009_base.json`) was created, nor how it should be re-created in the future if changes were needed. Doing some historical research, it appears that the current `jones2009_base.json` file at the time it was last updated (commit af4b10588e5af12a48c1ffff809395f4d8376307 ) was created by using the "flat JSON" `hnn_core/param/default.json` as the input to the function `hnn_core/params::convert_to_json()`. You can test this by using the `regenerate_base_network.py` script included in this PR on that old commit, and verifying that it produces an output `jones2009_base.json` file that is the same as before, except for a few negligible floating-point differences. This is almost certainly how the file was created initially.

This PR adds a new script that "regenerates" `jones2009_base.json` using the original way, as part of any default model updates, and provides a convenient `Makefile` recipe for doing so.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,10 +44,8 @@ lint:
 modl:
 	cd hnn_core/mod/ && nrnivmodl
 
-regenerate-base-network:
+regenerate-networks:
 	python ./dev_scripts/regenerate_base_network.py
-
-regenerate-test-network:
 	python ./hnn_core/tests/regenerate_test_network.py
 
 spell:
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,9 @@ lint:
 modl:
 	cd hnn_core/mod/ && nrnivmodl
 
+regenerate-base-network:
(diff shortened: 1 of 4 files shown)


asoplata:
I'll note that this does **not** add a newly-regenerated `jones2009_base.json` file itself, since that is unnecessary. This only adds the means to do so. I have tested a newly-regenerated version locally, and the tests pass. This is mostly intended for historical reference and actual usage as part of #985.

PR #1058, opened 2025-06-05
[MRG] Bugfix: Update vecevent mod

asoplata:
This may resolve #1052, but needs to be comprehensively tested.

In testing for #1052, the issue seemed to be that, with NEURON 8.2.7 on Windows, `nrnivmodl` is no longer able to compile our `vecevent.mod` file, and complains about inconsistent types. Here is example output:

```
(test-nrn-827) C:\Users\austi\mod-test>nrnivmodl
x86_64-w64-mingw32-gcc.exe -DDLL_EXPORT -DPIC -I/cygdrive/c/nrn/bin//../include -I/cygdrive/c/nrn/bin//../src/scopmath -I/cygdrive/c/nrn/bin//../src/nrnoc -I/cygdrive/c/nrn/bin//../src/oc  -c mod_func.c
nocmodl vecevent
Translating vecevent.mod into vecevent.c
Notice: VERBATIM blocks are not thread safe
x86_64-w64-mingw32-gcc.exe -DDLL_EXPORT -DPIC -I/cygdrive/c/nrn/bin//../include -I/cygdrive/c/nrn/bin//../src/scopmath -I/cygdrive/c/nrn/bin//../src/nrnoc -I/cygdrive/c/nrn/bin//../src/oc  -c vecevent.c
vecevent.c:241:16: error: conflicting types for 'vector_vec'; have 'double *(void)'
  241 | extern double* vector_vec();
      |                ^~~~~~~~~~
In file included from C:/nrn/include/hocdec.h:280,
                 from C:/nrn/include/section.h:37,
                 from vecevent.c:11:
C:/nrn/include/oc_ansi.h:57:16: note: previous declaration of 'vector_vec' with type 'double *(void *)'
   57 | extern double* vector_vec(IvocVect*);
      |                ^~~~~~~~~~
vecevent.c:242:12: error: conflicting types for 'vector_capacity'; have 'int(void)'
  242 | extern int vector_capacity();
      |            ^~~~~~~~~~~~~~~
C:/nrn/include/oc_ansi.h:58:12: note: previous declaration of 'vector_capacity' with type 'int(void *)'
   58 | extern int vector_capacity(IvocVect*);
      |            ^~~~~~~~~~~~~~~
vecevent.c:243:14: error: conflicting types for 'vector_arg'; have 'void *(void)'
  243 | extern void* vector_arg();
      |              ^~~~~~~~~~
[...]

PR #1070, opened 2025-06-13
[MRG] MAINT: Remove Python 3.8 support, add Python 3.13 support

asoplata:
This removes explicit support for Python 3.8, and explicitly adds support for 3.13. This also increments the versions used in *some* of our unit tests: unit tests now run for every other supported Python version (3.9, 3.11, and 3.13). This also brings our Python version support to exactly match that of the latest release of NEURON, our most important dependency, see https://github.com/neuronsimulator/nrn/releases/tag/8.2.7

Note that this was built off of #1058, due to its reversion of the NEURON 8.2.6 version cap. This should only be merged after #1058 is merged.

### DIFF
--- a/.github/workflows/unix_unit_tests.yml
+++ b/.github/workflows/unix_unit_tests.yml
@@ -5,7 +5,7 @@
       branches: ['**']
     pull_request:
       branches: ['**']
-  
+
   concurrency:
     group: ${{ github.workflow }}-${{ github.ref }}
     cancel-in-progress: true
@@ -17,7 +17,7 @@
       strategy:
         matrix:
           os: [ubuntu-latest,macos-latest]
-          python-version: [3.8, 3.11, 3.12]
+          python-version: [3.9, 3.11, 3.13]
 
       steps:
         - uses: actions/checkout@v4
--- a/.github/workflows/windows_unit_tests.yml
+++ b/.github/workflows/windows_unit_tests.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Install Neuron on Windows separately
         shell: cmd
         run: |
-          powershell -command "gh release download 8.2.6 --repo https://github.com/neuronsimulator/nrn --pattern '*.exe' --output nrn-setup.exe"
+          powershell -command "gh release download --repo https://github.com/neuronsimulator/nrn --pattern '*.exe' --output nrn-setup.exe"
           start /b /wait .\nrn-setup.exe /S /D=C:\nrn_test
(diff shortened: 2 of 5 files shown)


asoplata:
Looking at what's available from NEURON, if you use a Python 3.8 environment and then install NEURON (NRN), it installs the prior version, 8.2.6, NOT the current version. In other words, NRN8.2.6 is still available from Pypi for those with Python 3.8. However, the latest version, NRN8.2.7, *has* raised the Python minimum version to 3.9.

To balance bugfixes with keeping up to date with our dependencies, I think we release one new version (v0.4.2), **before** we merge any Python version bumps.

The next new version (v0.4.2) including Katharina's LFP bugfix #1062 , the `vecevent.mod` fix #1058, and everything else currently on `master`, but NOT including the Python version bump #1070 . This will become HNN-Core v0.4.2, will still be compatible with Python 3.8, and will *not* have its metadata in `setup.py` enforce `python>=3.9`. This way, anyone still on 3.8 will still get this and other bugfixes, but we will still be able to build directly off of `master`. This version will support both NRN8.2.6 and NRN8.2.7. (I will note that the `vecevent.mod` fix #1058 appears to be backwards-compatible with 3.8, as illustrated [here](https://github.com/asoplata/hnn-core/actions/runs/15690704190), [here](https://github.com/asoplata/hnn-core/actions/runs/15690704188), [here](https://github.com/jonescompneurolab/hnn-core/actions/runs/15710068417) and [here](https://github.com/jonescompneurolab/hnn-core/actions/runs/15710068415).)

Then, after we release v0.4.2, #1070 (the Py versions update) should be merged. We don't need a second release now, but when we do, it should have a `minor` rather than `patch` version increment (i.e. to the `0.5.x` series) to indicate its version support changing.

I will also need to manually update and rebuild all the the Conda packages for the new NEURON version.

If we need to backport future bugfixes onto a version that supports Py3.8, such as if a researcher asks for it, then there is a very common way to do that. We would make a new, off-`master` branch for a hypothetical `v0.4.3`, which is built off of `v0.4.2`, but only includes the necessary backported bugfixes, then push that patch-release to Pypi. This is traditionally what is done using maintenance branches like our (for example) `maint/0.4.1` branch, etc.

Task list:

[ ] Get automatic Github->Pypi packaging working #1073
[ ] Update changelog
[ ] Publish v0.4.2 release, including manual documentation steps
[ ] Update NEURON depencies, rebuild, and push all Conda packages of https://github.com/jonescompneurolab/hnn-core-conda-packaging (maybe take a stab at automating with Github Actions)

PR #1071, opened 2025-06-13
[WIP]: Harmonize unit test installs

asoplata:
This changes our `unix_unit_tests.yml` to go back to using Conda's versions of MPI, rather than Ubuntu's `apt` package versions. This is based on the fact that we now always recommend using Conda for all MPI installations on every platform, see https://jonescompneurolab.github.io/hnn-core/stable/install.html

This also does a little cleanup, including removing miniconda Github Action keys that are either unnecessary or no longer supported, among other things.

This is not ready for merge yet, but will be used for testing.


asoplata:
@dylansdaniels This has been rebased so that it's no longer dependent on the Python version update. This should be safe to merge immediately when the last unit test finishes.

asoplata:
Sorry, was helping Chetan with some install issues. I've finished rebasing this onto master, so the conflicts should be solved. If this passes the unit tests, then it is ready for merge.

PR #1073, opened 2025-06-16
[MRG]: pkg: test initial Github building

asoplata:
This PR is for testing an attempt to change our build process to use Github Acitons -> Pypi, a la https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ .

All non-building workflows have been commented out.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,26 +1,27 @@
----
-name: Codespell
-
-on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
-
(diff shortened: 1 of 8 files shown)


dylansdaniels:
May the --force be with you

dylansdaniels:
Looks like a huge improvement. We'll have to discuss our new release process and document it, which will likely be a larger discussion. But this code is "non-destructive" in the sense that it's still safe to merge this in before we iron out all of those details. I'm comfortable merging this :)

PR #1074, opened 2025-06-17
[MRG] Fix Incorrect Raster Plot Title When Dipoles Not Overlaid  

dylansdaniels:
fix incorrect title on raster plot - the title will now change conditionally based on whether the dipole overlays are also plotted


asoplata:
I did a force-push which did NOT include code changes, but only changed the *commit message itself*. The commit message title was too long (>50 chars), so I formatted the commit message contents to fit the standard practice. No other comments on the code, checks out. Will auto-merge after tests pass. Thanks!

PR #1076, opened 2025-06-23
[MRG] Bugfix: raise minimum opt iters for upstream deps

asoplata:
This is at least a temporary (possibly permanent) fix for a new bug causing our MacOS unit tests to fail. This bug is probably due to a change in upstream dependency changes in `scipy.optimize`, where the COBYLA solver now seems to require that the number of maximum iterations (MAXFUN) be `at least num_vars + 2`, where num_vars is the dimension size of the variables being "optimized over", which for `optimized_evoked` is at least 10. I will write a longer description of the issue elsewhere. For now, this can probably be merged, and it should allow current tests to pass.


dylansdaniels:
This sounds good as a temporary fix to allow tests to pass

PR #1090, opened 2025-07-14
[MRG] doc: branch protection for publish workflow

asoplata:
Self-merging since this is only a documentation change.


asoplata:
Rebased so this PR now depends on #1091

PR #1091, opened 2025-07-15
[MRG] fix: update CircleCI image to Ubuntu 22.04

asoplata:
This is a possible fix to recent CircleCI build failures (example: https://app.circleci.com/pipelines/github/jonescompneurolab/hnn-core/4636/workflows/5d565736-9fc2-4ec9-b791-10f2e1b291c5/jobs/4921 ) and possibly Ubuntu 3.9 unit tests (example https://github.com/jonescompneurolab/hnn-core/actions/runs/16277024151/job/45959260921?pr=1088 ).

The CircleCI failure outputs the following:
```
ImportError: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /home/circleci/miniconda/envs/testenv/lib/python3.13/site-packages/numpy/_core/_multiarray_umath.cpython-313-x86_64-linux-gnu.so)
ImportError: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /home/circleci/miniconda/envs/testenv/lib/python3.13/site-packages/numpy/_core/_multiarray_umath.cpython-313-x86_64-linux-gnu.so)
```

which suggests, among other things, that either the OS is too old and/or our version of Numpy is too new. NEURON 8.2.6 enforced Numpy <2 but we recently updated to NEURON 8.2.7, which uses Numpy >2 and <=2.2.3 (see https://github.com/neuronsimulator/nrn/releases and for the explicit numpy version cap, see the `pyproject.toml` file included in the source release of https://github.com/neuronsimulator/nrn/releases/tag/8.2.7 ). Therefore, support for Numpy >2 is also new for *us* due to our strict NEURON dependency. 

For the Ubuntu 3.9 unit test failures, I am less confident that this is the cause, but it could still be a cause. To be clear we have been getting random/stochastic `MPI Timeout error code 143` unit test failures for as long as I have been working for HNN, and the cause is still unknown. Usually, re-running the unit test once or twice allows passing. This change may help with that problem.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -117,7 +117,7 @@ if __name__ == "__main__":
           ],
           platforms='any',
           install_requires=[
-              'numpy >=1.14',
+              'numpy <=2.2.3',
               'NEURON >=7.7; platform_system != "Windows"',
               'matplotlib>=3.5.3',
               'scipy',
--- a/setup.py
+++ b/setup.py
@@ -117,7 +117,7 @@ if __name__ == "__main__":
           ],
           platforms='any',
           install_requires=[
-              'numpy <=2.2.3',
+              'numpy >=1.14',
               'NEURON >=7.7; platform_system != "Windows"',
               'matplotlib>=3.5.3',
               'scipy',


ntolley:
When you say we updated to NEURON 8.2.7, is this just stating that hnn-core is compatible with this version? I see that we're only requiring NEURON > 7.7

asoplata:
You're right, I should have been clearer. It's more accurate to say that *we added support* for NEURON 8.2.7 ; we do *not* require it. However, NEURON 8.2.7 is now the default (as opposed to 8.2.6) for anyone doing a standard `pip` install of NEURON using Python 3.9 or higher. Therefore, most users doing a fresh install of HNN will probably be installing NEURON 8.2.7 automatically, and therefore will use its dependency caps as well.

After thinking abou tit, your comment made me realize that raising the Numpy version floor *would* de facto force NEURON 8.2.7 usage by users as of this PR, because NEURON 8.2.6 actually requires `"numpy>=1.9.3,<2",` (copied from `setup.py` of the source-distribution of NEURON 8.2.6). If we aspire to support as wide a range of dependencies as possible (including NEURON 8.2.6), then raising the Numpy floor here is not ideal.

asoplata:
I've just pushed a new commit that I believe fixes the issue for the CircleCI build, but in a completely different way: it updates the host image used in the CCI building from Ubuntu 20.04 (which it was still set to) to Ubuntu 22.04. This seems to also fix the issue, probably because it has newer version of GLIB library objects than 20.04 did.

asoplata:
I'll also note that, while this new change to update the CCI host image version only affects CircleCI (therefore will not have an effect on our Ubuntu 3.9 unit tests), in the most recent tests, Ubuntu 3.9 appears to now be passing, whereas it was not a few days ago. 🤷

PR #1095, opened 2025-07-30
[WIP] Introducing layer_dict for cell coordinates

Chetank99:
Hi!!

so this pull request is the first step towards refactoring the way cell positions and network structures are created and managed.

In this PR, the main change was the introduction of a `layer_dict` to provide a more structured and anatomically-based way of defining cell coordinates, moving away from the previous `pos_dict` implementation

-  In the function `_create_cell_coords` in `hnn_core/network.py`, main changes can be seen:
	- **In the current method**....It creates a `pos_dict` with keys like 'L2_pyramidal' and 'L5_pyramidal' and contained functions like `_calc_pyramidal_coord` and `_calc_basket_coord`
    
	- **What the PR proposes** .....It now returns a `layer_dict` with more descriptive keys representing anatomical layers, such as 'L5_bottom', 'L2_bottom', 'L5_mid', and 'L2_mid'.
		- this step is essential as it forms the basis of our future PRs which will build upon this to create metaData and custom position logic on top of these changes
    

The `Network` class constructor in `hnn_core/network.py` was also changed 

- It accepts`pos_dict` and `cell_types` as arguments during initialization. This allows for the future PR's for the creation of networks with custom cell arrangements and types, rather than being hard-coded.
    
- If no custom positions are provided, it falls back to a default mechanism that uses the new `layer_dict` to map cell types to their respective anatomical positionss.

some methods are being shown as green but there are exactly the same, just re-arranged in position they are placed at (`add_cell_type`,`_add_cell_type`,`add_spike_train_drive`)

The `jones_2009_model` in `hnn_core/network_models.py` was updated to use the new stuff

 - Defines the necessary `cell_types` and their templates.
    
- it uses `create_cell_coords` to get the anatomical `layer_dict`
	- maps these  to the appropriate cell types to create the `pos_dict`
    
- finally passes both `cell_types` and `pos_dict` to the `Network` constructor to build the model

for some context....
the work done till now was this:
Chunk 1: Centralizes all cell creation into a single factory that packages cell objects with their metadata.

Chunk 2: Enforces the consistent use of short cell names (like L5Pyr) across the entire codebase...it is contested for now

Chunk 3: Standardizes cell positioning to be generated from a layer_dict, allowing for optional custom overrides.

Chunk 4 (Metadata): Refactors all logic for connections, drives, and plotting to be driven by cell metadata instead of hardcoded type checks.

The PRs will be submitted in the order 3 → 1 → 2 → 4.
(depending on decision to keep chunk 2 or not)

This order is based on dependencies: we start with the most independent chunk (Chunk 3) and finish with the most dependent chunk (Chunk 4), which relies on all the others being complete.

### DIFF
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -63,16 +63,16 @@ def _create_cell_coords(n_pyr_x, n_pyr_y, zdiff, inplane_distance):
     Sort of a hack because of redundancy.
     """
 
-    def _calc_pyramidal_coord(xxrange, yyrange, zdiff):
-        list_coords = [pos for pos in it.product(xxrange, yyrange, [zdiff])]
+    def _calc_pyramidal_coord(xxrange, yyrange, z_coord):
+        list_coords = [pos for pos in it.product(xxrange, yyrange, [z_coord])]
         return list_coords
 
-    def _calc_basket_coord(n_pyr_x, n_pyr_y, zdiff, inplane_distance, weight):
-        xzero = np.arange(0, n_pyr_x, 3) * inplane_distance
-        xone = np.arange(1, n_pyr_x, 3) * inplane_distance
+    def _calc_basket_coord(n_x, n_y, z_coord, inplane_distance, weight):
+        xzero = np.arange(0, n_x, 3) * inplane_distance
+        xone = np.arange(1, n_x, 3) * inplane_distance
         # split even and odd y vals
-        yeven = np.arange(0, n_pyr_y, 2) * inplane_distance
-        yodd = np.arange(1, n_pyr_y, 2) * inplane_distance
+        yeven = np.arange(0, n_y, 2) * inplane_distance
+        yodd = np.arange(1, n_y, 2) * inplane_distance
         # create general list of x,y coords and sort it
         coords = [pos for pos in it.product(xzero, yeven)] + [
             pos for pos in it.product(xone, yodd)
@@ -83,7 +83,6 @@ def _create_cell_coords(n_pyr_x, n_pyr_y, zdiff, inplane_distance):
         list_coords = [
             (pos_xy[0], pos_xy[1], weight * zdiff) for pos_xy in coords_sorted
         ]
-
         return list_coords
 
     def _calc_origin(xxrange, yyrange, zdiff):
@@ -99,19 +98,20 @@ def _create_cell_coords(n_pyr_x, n_pyr_y, zdiff, inplane_distance):
     xxrange = np.arange(n_pyr_x) * inplane_distance
     yyrange = np.arange(n_pyr_y) * inplane_distance
 
-    pos_dict = {
-        "L5_pyramidal": _calc_pyramidal_coord(xxrange, yyrange, zdiff=0),
-        "L2_pyramidal": _calc_pyramidal_coord(xxrange, yyrange, zdiff=zdiff),
-        "L5_basket": _calc_basket_coord(
-            n_pyr_x, n_pyr_y, zdiff, inplane_distance, weight=0.2
+    # Create layer dictionary with anatomical layer positions
+    layer_dict = {
+        "L5_bottom": _calc_pyramidal_coord(xxrange, yyrange, z_coord=0),
+        "L2_bottom": _calc_pyramidal_coord(xxrange, yyrange, z_coord=zdiff),
+        "L5_mid": _calc_basket_coord(
+            n_pyr_x, n_pyr_y, z_coord=0, inplane_distance=inplane_distance, weight=0.2
         ),
-        "L2_basket": _calc_basket_coord(
-            n_pyr_x, n_pyr_y, zdiff, inplane_distance, weight=0.8
+        "L2_mid": _calc_basket_coord(
+            n_pyr_x, n_pyr_y, z_coord=0, inplane_distance=inplane_distance, weight=0.8
         ),
         "origin": _calc_origin(xxrange, yyrange, zdiff),
     }
 
-    return pos_dict
+    return layer_dict
 
 
 def _compare_lists(s, t):
@@ -376,6 +376,8 @@ class Network:
         add_drives_from_params=False,
         legacy_mode=False,
         mesh_shape=(10, 10),
+        pos_dict=None,
+        cell_types=None,
     ):
         # Save the parameters used to create the Network
         _validate_type(params, dict, "params")
@@ -401,14 +403,6 @@ class Network:
                 DeprecationWarning,
                 stacklevel=1,
             )
-
-        # Source dict of names, first real ones only!
-        cell_types = {
-            "L2_basket": basket(cell_name=_short_name("L2_basket")),
-            "L2_pyramidal": pyramidal(cell_name=_short_name("L2_pyramidal")),
-            "L5_basket": basket(cell_name=_short_name("L5_basket")),
-            "L5_pyramidal": pyramidal(cell_name=_short_name("L5_pyramidal")),
-        }
 
         self.cell_response = None
         # external drives and biases
@@ -444,17 +438,42 @@ class Network:
 
         self._inplane_distance = 1.0  # XXX hard-coded default
         self._layer_separation = 1307.4  # XXX hard-coded default
-        self.set_cell_positions(
-            inplane_distance=self._inplane_distance,
-            layer_separation=self._layer_separation,
-        )
 
-        # populates self.gid_ranges for the 1st time: order matters for
-        # NetworkBuilder!
-        for cell_name in cell_types:
-            self._add_cell_type(
-                cell_name, self.pos_dict[cell_name], cell_template=cell_types[cell_name]
+        # Handle positions and cell types
+        if pos_dict is not None and cell_types is not None:
+            # Use provided positions and cell types
+            _validate_type(pos_dict, dict, "pos_dict")
+            _validate_type(cell_types, dict, "cell_types")
+            self.pos_dict = deepcopy(pos_dict)
+
+            # Add cell types from provided dictionary
+            for cell_name, cell_template in cell_types.items():
+                if cell_name in self.pos_dict:
+                    self._add_cell_type(
+                        cell_name, self.pos_dict[cell_name], cell_template=cell_template
+                    )
+        else:
+            # Default behavior - create standard network
+            cell_types_default = {
+                "L2_basket": basket(cell_name=_short_name("L2_basket")),
+                "L2_pyramidal": pyramidal(cell_name=_short_name("L2_pyramidal")),
+                "L5_basket": basket(cell_name=_short_name("L5_basket")),
+                "L5_pyramidal": pyramidal(cell_name=_short_name("L5_pyramidal")),
+            }
+
+            self.set_cell_positions(
+                inplane_distance=self._inplane_distance,
+                layer_separation=self._layer_separation,
             )
+
+            # populates self.gid_ranges for the 1st time: order matters for
+            # NetworkBuilder!
+            for cell_name in cell_types_default:
+                self._add_cell_type(
+                    cell_name,
+                    self.pos_dict[cell_name],
+                    cell_template=cell_types_default[cell_name],
+                )
 
         if add_drives_from_params:
             _add_drives_from_params(self)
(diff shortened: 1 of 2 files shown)


hnn_core/externals/mne.py
```diff
@@ -420,7 +420,9 @@ def _check_option(parameter, value, allowed_values, extra=""):
     )
     allowed_values = list(allowed_values)  # e.g., if a dict was given
     if len(allowed_values) == 1:
-        options = f"The only allowed value is {repr(allowed_values[0])}"
+        pass
```
ntolley, line 423: why was this necessary? There are very few instances where a `pass` is the right choice

Also we really should not modify `_check_option` as it is used all over the code base. If you have a function that needs this sort of behavior then I would suggest manually raising these messages and writing a test to check for the correct behavior

hnn_core/network.py
```diff
@@ -31,87 +31,75 @@
...
-        "L2_basket": _calc_basket_coord(
-            n_pyr_x, n_pyr_y, zdiff, inplane_distance, weight=0.8
+        "L2_mid": _calc_basket_positions(
+            n_pyr_x, n_pyr_y, z_coord=0.8 * zdiff, inplane_distance=inplane_distance
         ),
         "origin": _calc_origin(xxrange, yyrange, zdiff),
     }
 
-    return pos_dict
+    return layer_dict
```
ntolley, line 102: It seems like there should be some simple unit tests to verify the correct creation of this object? Especially since it's a new format to store this information

hnn_core/network.py
```diff
@@ -376,60 +364,38 @@ def __init__(
...
         self.threshold = self._params["threshold"]
         self.delay = 1.0
-
-        # extracellular recordings (if applicable)
         self.rec_arrays = dict()
-
-        # contents of pos_dict determines all downstream inferences of
-        # cell counts, real and artificial
-        self._n_cells = 0  # used in tests and MPIBackend checks
+        self._n_cells = 0
```
ntolley, line 383: Are you sure this and `_n_gids` are being updated correctly? Can you add a comment explaining specifically what is included in this count? (i.e. real vs artificial cells)

hnn_core/externals/mne.py
asoplata: I def agree with Nick here, but if I recall correctly this was something done as part of early debugging efforts. If this is needed for a particular function's options, then that particular function should almost certainly be changed instead of changing this globally.

hnn_core/network.py
```diff
@@ -31,87 +31,75 @@
...
-        list_coords = [pos for pos in it.product(xxrange, yyrange, [zdiff])]
+    def _calc_grid_positions(xxrange, yyrange, z_coord):
+        """Calculate positions on a regular grid at given z-coordinate."""
+        list_coords = [pos for pos in it.product(xxrange, yyrange, [z_coord])]
         return list_coords
 
-    def _calc_basket_coord(n_pyr_x, n_pyr_y, zdiff, inplane_distance, weight):
-        xzero = np.arange(0, n_pyr_x, 3) * inplane_distance
-        xone = np.arange(1, n_pyr_x, 3) * inplane_distance
+    def _calc_basket_positions(n_x, n_y, z_coord, inplane_distance):
```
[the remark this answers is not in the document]
asoplata: `weight` should be re-added here, instead of being hard-coded later on inside the `layer_dict` creation.

hnn_core/network.py
```diff
@@ -31,87 +31,75 @@
...
-            (pos_xy[0], pos_xy[1], weight * zdiff) for pos_xy in coords_sorted
-        ]
-
+        list_coords = [(pos_xy[0], pos_xy[1], z_coord) for pos_xy in coords_sorted]
         return list_coords
 
     def _calc_origin(xxrange, yyrange, zdiff):
-        # origin's z component isn't used in calculating distance functions.
-        # will be used for adding external drives.
+        """Calculate the origin position."""
```
asoplata, line 78: I appreciate adding a docstring to a function that didn't have one before, but your new docstring should still retain the old, useful information that was originally a comment.

hnn_core/network.py
```diff
@@ -376,60 +364,38 @@ def __init__(
         add_drives_from_params=False,
         legacy_mode=False,
         mesh_shape=(10, 10),
+        pos_dict=None,
+        cell_types=None,
     ):
-        # Save the parameters used to create the Network
         _validate_type(params, dict, "params")
         self._params = params
-        # Initialise a dictionary of cell ID's, which get used when the
```
asoplata, line 383: Why were all these comments deleted? Unless you are making changes that **require** these comments to be re-written or deleted, they should be left alone. Even if not all of them are necessary, sometimes they reveal extremely helpful information to someone who is new (or old!) to the codebase. Comments are part of documentation too, and sometimes they're the most important part of documentation.

hnn_core/network.py
```diff
@@ -376,60 +364,38 @@ def __init__(
...
-
-        # contents of pos_dict determines all downstream inferences of
-        # cell counts, real and artificial
-        self._n_cells = 0  # used in tests and MPIBackend checks
+        self._n_cells = 0
         self.pos_dict = dict()
         self.cell_types = dict()
 
-        # set the mesh shape
+        # Legacy mode handling
```
asoplata, line 387: As far as I can tell, this legacy mode handling section hasn't actually been changed at all, other than its position has been moved. Is there a reason for moving this to a different location? It is best to not move around the order of code that isn't being changed since 1. it appears like it *is* being changed in the diff, and 2. if the order doesn't matter, then it should stay where it is.

hnn_core/network.py
```diff
@@ -1492,56 +1538,19 @@ def add_tonic_bias(
...
-        self._n_gids += len(pos)
-        self.gid_ranges[cell_name] = range(ll, self._n_gids)
-        self.pos_dict[cell_name] = pos
-        if cell_template is not None:
-            self.cell_types.update({cell_name: cell_template})
-            self._n_cells += len(pos)
-
-    def _rename_cell_types(self, name_mapping: Dict[str, str]):
-        """Renames cell types in the network.
+    def add_cell_type(self, cell_name, cell_template, positions):
```
asoplata, line 1541: As far as I can tell, this function is never actually used, and is mostly just a wrapper around the almost identically named `_add_cell_type`. If there needs to be a public version of this function, then we should instead add the checks here to inside `_add_cell_type` and "promote" `_add_cell_type` to a public method of `Network`.

hnn_core/network.py
```diff
@@ -985,130 +1155,6 @@ def add_bursty_drive(
             probability,
         )
 
-    def add_spike_train_drive(
```
asoplata, line 988: As far as I can tell, no code has changed inside this function, only the location of the function has been changed. Please revert its location

hnn_core/network.py
```diff
@@ -63,16 +63,16 @@ def _create_cell_coords(n_pyr_x, n_pyr_y, zdiff, inplane_distance):
...
-    def _calc_pyramidal_coord(xxrange, yyrange, zdiff):
-        list_coords = [pos for pos in it.product(xxrange, yyrange, [zdiff])]
+    def _calc_pyramidal_coord(xxrange, yyrange, z_coord):
+        list_coords = [pos for pos in it.product(xxrange, yyrange, [z_coord])]
         return list_coords
 
-    def _calc_basket_coord(n_pyr_x, n_pyr_y, zdiff, inplane_distance, weight):
-        xzero = np.arange(0, n_pyr_x, 3) * inplane_distance
-        xone = np.arange(1, n_pyr_x, 3) * inplane_distance
+    def _calc_basket_coord(n_x, n_y, z_coord, inplane_distance, weight):
```
katduecker, line 70: Am I seeing correctly that you changed the name to z_coord here but you're still using input zdiff in network_models?

hnn_core/network.py
```diff
@@ -444,18 +438,43 @@ def __init__(
...
+
+            # Add cell types from provided dictionary
+            for cell_name, cell_template in cell_types.items():
+                if cell_name in self.pos_dict:
+                    self._add_cell_type(
+                        cell_name, self.pos_dict[cell_name], cell_template=cell_template
+                    )
+        else:
+            # Default behavior - create standard network
+            cell_types_default = {
```
katduecker, line 457: Why not just call this cell_types instead of cell_types_default?

hnn_core/network.py
```diff
@@ -531,15 +550,22 @@ def set_cell_positions(self, *, inplane_distance=None, layer_separation=None):
...
             n_pyr_y=self._N_pyr_y,
             zdiff=layer_separation,
             inplane_distance=inplane_distance,
         )
-        # update positions of the real cells
-        for key in pos.keys():
-            self.pos_dict[key] = pos[key]
+
+        # Map layers to cell types, for default mapping
+        self.pos_dict = {
```
katduecker, line 562: This should be defined in network_models and not network (we don't want hard-coded cell names in network). Or are you going to address this in your next PR?

hnn_core/network_models.py
```diff
@@ -68,11 +72,39 @@ def jones_2009_model(
...
+    # Create layer positions
+    layer_dict = _create_cell_coords(
+        n_pyr_x=mesh_shape[0],
+        n_pyr_y=mesh_shape[1],
+        zdiff=1307.4,  # Default layer separation
+        inplane_distance=1.0,  # Default in-plane distance
+    )
+
+    # Map cell types to layer positions
+    pos_dict = {
```
katduecker, line 92: If you define pos_dict here why define it in network?

hnn_core/network.py
Chetank99: i kept it as cell_types_default to avoid shadowing the cell_types parameter in the function signature. Since we have cell_types as an optional parameter (in __init__), using the same name for the local variable might create confusion and potentially mask the parameter?? idk that was my thinking..

the _default makes it explicit that these are the fallback/default cell types used when no custom cell types are provided?

i can also not have it, it is more of a explanation based naming, i can work around it too

hnn_core/network.py
asoplata: One thing Chetan and I discussed was Chetan leaving this in for backwards compatibility issues (e.g. some expert users still use the base Network), both for the new __init__ arguments and overall behavior

hnn_core/network.py
asoplata: After going over this with Chetan, I think these lines were only highlighted due to comment changes that Chetan originally made. He since rebased all of the comment-only changes out. Looking at the much cleaner status of the PR, I don't believe he makes any changes to `_n_cells` or `_n_gids`, or any GID logic for that matter. @ntolley Does it appear the same way to you now? (In the most recent PR version)

hnn_core/tests/test_network.py
```diff
@@ -92,6 +92,77 @@ def base_network():
...
+        'origin': custom_layer_dict['origin']
+    }
+    custom_net = Network(params, pos_dict=custom_pos_dict, cell_types=custom_cell_types)
+    assert 'L2_pyramidal' in custom_net.cell_types
+    assert 'L5_pyramidal' in custom_net.cell_types
+    assert len(custom_net.pos_dict['L2_pyramidal']) == 4
+    assert custom_net.gid_ranges['L2_pyramidal'] == range(0, 4)
+    assert custom_net.gid_ranges['L5_pyramidal'] == range(4, 8)
+
+    # seeing if i add a spike train drive
```
asoplata, line 139: Just for good measure, I would recommend you also add simple versions of all other types of drives to the custom network. You want to test that there are no interactions which break other parts of the network object.

hnn_core/tests/test_network.py
```diff
@@ -92,6 +92,77 @@ def base_network():
...
+        'L2_pyramidal': pyramidal(cell_name=_short_name('L2_pyramidal')),
+        'L5_pyramidal': pyramidal(cell_name=_short_name('L5_pyramidal'))
+    }
+    custom_layer_dict = _create_cell_coords(n_pyr_x=2, n_pyr_y=2, z_coord=1307.4, inplane_distance=1.0)
+    custom_pos_dict = {
+        'L2_pyramidal': custom_layer_dict['L2_bottom'],
+        'L5_pyramidal': custom_layer_dict['L5_bottom'],
+        'origin': custom_layer_dict['origin']
+    }
+    custom_net = Network(params, pos_dict=custom_pos_dict, cell_types=custom_cell_types)
```
asoplata, line 132: It's good to test that your new Network arguments work with an irregular input, as you've done here. However, you should also test that the new Network arguments work when you manually provide the normal, expected values, such as those used in the Jones 2009 model codel

hnn_core/tests/test_network.py
```diff
@@ -92,6 +92,77 @@ def base_network():
...
     return net, params
 
+def test_network_models_mod():
+    from hnn_core.network import _create_cell_coords
+    from hnn_core.cells_default import pyramidal, basket
+    from hnn_core.params import _short_name
+
+    params = read_params(params_fname)
+
+    # layer_dict change
```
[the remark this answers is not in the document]
Chetank99: I agree!
test of _create_cell_coords moved to a separate func..

hnn_core/tests/test_network.py
Chetank99: agreed and addressed !
testing constructor with standard Jones 2009 values to ensure it works with normal/expected parameters, not just irregular ones.
line 265

hnn_core/network.py
asoplata: oh you bet we're adding tests :)

hnn_core/network.py
asoplata: AFAIK, all zdiffs have been converted to z_coords

PR #1096, opened 2025-07-30
[WIP]: Add GID-shifting support for future dual-network-simulation

asoplata:
Turning this into a PR just to keep track of the conversation, even if we end up shifting some of the work into another branch.


hnn_core/network.py
```diff
@@ -1492,10 +1495,14 @@ def add_tonic_bias(
                     t_stop=tstop,
                 )
 
-    def _add_cell_type(self, cell_name, pos, cell_template=None):
+    def _add_cell_type(self, cell_name, pos, cell_template=None, gid_start=None):
         """Add cell type by updating pos_dict and gid_ranges."""
-        ll = self._n_gids
-        self._n_gids += len(pos)
+        if gid_start is None:
```
asoplata, line 1500: This could be a one-liner since the logic is simple, like
```
ll = self._n_gids if (gid_start is None) else gid_start
```

hnn_core/network.py
```diff
@@ -1535,10 +1547,11 @@ def add_tonic_bias(
                     t_stop=tstop,
                 )
 
-    def _add_cell_type(self, cell_name, pos, cell_template=None):
+    def _add_cell_type(self, cell_name, pos, cell_template=None, gid_start=None):
```
asoplata, line 1550: Please add tests inside `tests/test_network.py` for most (ideally all) your changes that you have made to both `network.py` AND `network_builder.py`. Specifically, `network_builder.py` does not have its own test file currently, and instead its tests should go in `test_network.py`.

hnn_core/tests/test_minimal_dual_network.py
```diff
@@ -0,0 +1,271 @@
+import os
```
asoplata, line 1: The `minimal_dual_network_simulation.py` script does not need tests. Since it is an example or tutorial of *how* to do the dual network simulations, it doesn't need tests, but should be converted into an "example/tutorial" kind of format. I wouldn't worry about adding too much example documentation to `minimal_dual_network_simulation.py` for now, however, since the example script itself is a WIP based on the future "true" combined network feature.

examples/howto/plot_dual_network_simulation.py
```diff
@@ -0,0 +1,172 @@
...
+    gid_start=drive_gid_2
+)
+
+print("Net1 gid ranges:", net1.gid_ranges)
+print("Net2 gid ranges:", net2.gid_ranges)
+
+print("Next GIDs after drives:", net1._get_next_available_gid(), net2._get_next_available_gid())
+
+###############################################################################
+# Step 7: (Optional) Export Configurations
```
dylansdaniels, line 102: It seems like we don't really need this optional section? There's no explanation to why we're doing this and the written outputs are never used

examples/howto/plot_dual_network_simulation.py
```diff
@@ -0,0 +1,172 @@
...
+print("Configuration files written (net1_config.json, net2_config.json).")
+
+###############################################################################
+# Step 8: Simulate Networks Independently
+# ---------------------------------------
+
+dpl1 = simulate_dipole(net1, tstop=20.0, dt=0.025, n_trials=1)
+dpl2 = simulate_dipole(net2, tstop=20.0, dt=0.025, n_trials=1)
+
+print("Net1 dipole peak:", max(abs(dpl1[0].data['agg'])))
```
dylansdaniels, line 116: Some comments on why we're showing these things would be good

examples/howto/plot_dual_network_simulation.py
asoplata: Deleted in latest commit.

asoplata:
Ready for round 2 of review

examples/howto/plot_dual_network_simulation.py
```diff
@@ -0,0 +1,171 @@
...
+# Step 8: Plot Dipoles
+# --------------------
+dpls = [dpl1[0], dpl2[0]]
+plot_dipole(dpls, show=True)
+
+###############################################################################
+# Step 9: Combined Spike Raster
+# ------------------------------
+
+def plot_combined_spike_raster(net_a, net_b, title="Combined Spike Raster"):
```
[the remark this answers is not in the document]
asoplata: This is kind of a validation plot, since this is still in beta: this allows us to confirm that everything is working correctly, even for a GID-shifted network versus a non-GID-shifted one (same inputs leads to same outputs). This plot specifically was what Maira was using for much of the debugging, especially including cell-type-specific discrepancies between the two networks as they arose. I wanted to leave this in here especially because it will be a useful validation function when we return to developing this functionality.

examples/howto/plot_dual_network_simulation.py
```diff
@@ -0,0 +1,171 @@
+"""
+==============================================
+10. Simulating Two Minimal Independent Networks
```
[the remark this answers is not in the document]
asoplata: You raise a good point. Just to make things more complicated, we are in the process of moving all of our "execution of example notebooks" over to the Textbook repo here: https://github.com/jonescompneurolab/textbook . Once we do that, we are planning to *remove* the Examples from the Developer Portal website here https://jonescompneurolab.github.io/hnn-core/stable/auto_examples/index.html since the Textbook will have replaced the examples. When that happens, we will still be able to keep this example script as a notebook on the Textbook site for future use, but *without* actually publishing or showing it on the webpage unless we want to. 

In other words, after we finish updating the Textbook execution stuff, we can choose whether or not to display this example script/notebook, while still keeping it around.

hnn_core/network.py
```diff
@@ -1019,6 +1032,10 @@ def add_bursty_drive(
         conn_seed : int
             Optional initial seed for random number generator (default: 3).
             Used to randomly remove connections when probability < 1.0.
+        gid_start : int, default: None
```
ntolley: Maybe add the default values for the synaptic delays and space constants like it was done in the previous modifications? No need to overhaul all the docstrings, but just to keep consistent with what is changed in the other drives

hnn_core/network.py
```diff
@@ -1583,10 +1615,25 @@ def add_tonic_bias(
...
+            Coordinates of cell soma in xyz-space.
+        cell_template : Instance of Cell object, default: None
+            Previously-instantiated `Cell` object to add for corresponding `cell_name`.
+        gid_start : int, default: None
+            Optional number to use as the starting index for the drive's GID range,
+            instead of the default. Default case (None) builds the GID range of the
+            drive starting from the Network's max GID + 1.
+
+        """
+        ll = self._n_gids if (gid_start is None) else gid_start
```
ntolley: what is `ll`?, is there a more explicit variable name we could use?

hnn_core/network.py
```diff
@@ -2175,6 +2222,80 @@ def _standardize_spike_data(self, spike_data):
 
         return standardized_data, n_drive_cells, source_to_gid_map
 
+    def _get_next_available_gid(self):
+        """Return the next unused GID number, equal to current max GID + 1."""
+        max_gid = -1
+        for rng in self.gid_ranges.values():
+            if len(rng) > 0:
+                max_gid = max(max_gid, max(rng))
```
```suggestion
        for gid_range in self.gid_ranges.values():
            if len(gid_range) > 0:
                max_gid = max(max_gid, max(gid_range))
```
Nitpick but `rng` is pretty exclusively used to mean random number generator in a lot of code

hnn_core/network.py
```diff
@@ -2175,6 +2222,80 @@ def _standardize_spike_data(self, spike_data):
...
+                gid_range = list(self.gid_ranges[ct])
+                n_targets = len(conn["target_gids"])
+                # Replace with correct GIDs from gid_ranges
+                conn["target_gids"] = gid_range[:n_targets]
+            # Shift src_gids if src_type is in mapping
+            if "src_type" in conn and conn["src_type"] in self.cell_types.keys():
+                ct = conn["src_type"]
+                gid_range = list(self.gid_ranges[ct])
+                n_srcs = len(conn["src_gids"])
+                conn["src_gids"] = gid_range[:n_srcs]
```
ntolley: this code could be condensed into a loop over `['src_type', 'target_type']`

hnn_core/network.py
```diff
@@ -2175,6 +2222,80 @@ def _standardize_spike_data(self, spike_data):
...
+
+            if "gid_pairs" in conn and conn["gid_pairs"]:
```
ntolley: The code looks correct to me so I think this would pass, but is there an explicit test ensuring that the same cells are connected when `conn['probability'] < 1.0`?

I think this is actually one of the most critical sections of code as it ensures that the same cells are connected even after their ids change

hnn_core/network_builder.py
```diff
@@ -408,7 +408,13 @@ def _gid_assign(self, rank=None, n_hosts=None):
             n_hosts = _get_nhosts()
 
         # round robin assignment of cell gids
-        for gid in range(self._rank, self.net._n_cells, n_hosts):
+        gids = list()
+        for cell_type, gid_range in self.net.gid_ranges.items():
+            # Only assign real cell types (not drives) here
```
ntolley: Does MPI break if drives are added before shifting?

hnn_core/tests/test_network.py
```diff
@@ -1911,3 +1911,98 @@ def test_update_weights_metadata():
...
+    # testing the main function of this test, but is also heavily dependent on the
+    # defaults of the drives themselves and Network construction.
+    assert net_base._get_next_available_gid() == 540
+
+
+def test_shift_gid_ranges(base_network):
+    """Test that `Network._shift_gid_ranges` works and simulates as expected."""
+    # "Base" has many connections and drives, but for now we're only interested in the
+    # non-drive params config
+    net_base, params = base_network
```
ntolley:
```suggestion
    _, params = base_network
```

ntolley (changes requested):
@asoplata I gave this a full read through! I think this is close to merge, the most pressing comment is the last one on the test, I think I am misunderstanding `get_next_available_gid()`

hnn_core/network.py
```diff
@@ -2175,6 +2222,80 @@ def _standardize_spike_data(self, spike_data):
...
+
+        Notes
+        -----
+        - This function is intended for future work in simulating multiple `Network`
+          objects within the same simulation. Specifically, this is intended for usage
+          with the "second" or any additional `Network` objects, such that the first
+          `Network` and additional `Networks` use a disjoint set of GIDs.
+        - This successfully shifts GIDs for individual cells and for non-drive synapses,
+          but currently does NOT shift GIDs for external drives. You should shift your
+          Network's GIDs only BEFORE adding drives to the Network.
```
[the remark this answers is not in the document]
asoplata: Resolved in a new commit.

hnn_core/network.py
asoplata: Added some in the latest commit, including greatly expanding that the connectivity (using equal `conn_seeds`) uses the same "shape", but shifted, between the base and shifted networks: https://github.com/Myrausman/hnn-core/commit/8445b45f5f0b5810ffaf5a0dd16e0454fe1fe514

examples/howto/plot_dual_network_simulation.py
```diff
@@ -0,0 +1,171 @@
...
+
+from hnn_core import jones_2009_model, simulate_dipole
+from hnn_core.viz import plot_dipole
+
+
+###############################################################################
+# Step 2: Helper to Build a Minimal Network
+# -----------------------------------------
+#
+# Creates a network with one cell per canonical cell type. Their GIDs reassigned to
```
ntolley:
```suggestion
# Creates a network with one cell per canonical cell type. Their GIDs are then reassigned to
```

examples/howto/plot_dual_network_simulation.py
```diff
@@ -0,0 +1,171 @@
...
+# Step 4: Add An Evoked Drive to net1
+# -----------------------------------
+# Each network receives a single distal evoked drive with matching timing
+# parameters but distinct names.
+
+drive_gid_1 = net1._get_next_available_gid()
+net1.add_evoked_drive(
+    'evdist1_net1', mu=5.0, sigma=1.0, numspikes=1, location='distal',
+    weights_ampa={'L2_pyramidal': 0.1, 'L5_pyramidal': 0.1},
+    gid_start=drive_gid_1
```
ntolley: Does it really make sense to expose this to the public API? If you set this value smaller than the next available gid, then there's an error (?)

But if it is larger, then there's a weird gap in the gid range

hnn_core/tests/test_network.py
```diff
@@ -2073,3 +2079,212 @@ def test_check_global_synaptic_gains_uniformity():
...
+    net_shifted._shift_gid_ranges(gid_start=gid_offset)
+    assert net_shifted._get_next_available_gid() == (270 + gid_offset)
+
+    gid_offset = 37
+    net_shifted._shift_gid_ranges(gid_start=gid_offset)
+    assert net_shifted._get_next_available_gid() == (270 + gid_offset)
+    _ = simulate_dipole(net_shifted, tstop=20.0, n_trials=1)
+
+
+@pytest.mark.parametrize("probability", [1.0, 0.67])  # SIX SEVEN /dance
```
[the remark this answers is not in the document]
asoplata: are you frame-mogging my comment???

asoplata:
In review, Nick and I have determined that, because we are in the process of migrating the `Connectivity` objects to `pandas` dataframes in #1322, it is best if we wait until that PR is merged, and then update the relevant functions here such as `Network._shift_gid_ranges` to use the new dataframes. The dataframes should make it much easier to make the relevant edits.

PR #1102, opened 2025-08-14
[WIP] parallel: double non-encoding timeout lengths

asoplata:
This is potentially and quick-and-easy bandaid for making MPI Timeout happen a *little* less frequently (see #774). I have intentionally not touched the timeout used in `run_subprocess` via `timeout_cycles` because I haven't had time to investigate its consequences. However, raising the other timeouts *may* make MPI Timeouts less likely to happen, *especially* in our Unit test runs (where they happen a significant fraction of the time, like more 10% of the time, IMHO).

@ntolley @dylansdaniels Is there any reason to think that increasing the length of these timeouts would DAMAGE simulation quality or reliability in any way? I still need to investigate that myself.

### DIFF
--- a/hnn_core/parallel_backends.py
+++ b/hnn_core/parallel_backends.py
@@ -119,7 +119,6 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
     timeout_cycles = timeout / 0.02
 
     pickled_obj = base64.b64encode(pickle.dumps(obj))
-    
 
     # non-blocking adapted from https://stackoverflow.com/questions/375427/non-blocking-read-on-a-subprocess-pipe-in-python#4896288  # noqa: E501
     out_q = Queue()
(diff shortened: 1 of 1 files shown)


asoplata:
Note: first round of CI tests all ran without having to restart *any* unit tests.

asoplata:
Recent example of where this band-aid may be helpful is that I had to restart the Ubuntu 3.9 Unit tests for another PR here **four times**, 3 of which were due to an "error code 143 MPI Timeout" at the very end of our tests https://github.com/jonescompneurolab/hnn-core/actions/runs/16974963673/job/48127191028?pr=1100 . The frequent need to re-run our MPI-failed tests due to this issue is not just annoying, but also ends wasting compute time, since I often need to re-start jobs 2 or 3 times just to get around this issue.

asoplata:
copy-pasting comment I made to Dylan over slack:

Most of the timeouts were increased from like, 1 second to two seconds. The biggest one was 30 seconds to 60, but that may be useful and necessary depending on the speed of the pickling/encoding. It's definitely a bandaid, but while timeouts mean the process takes longer both for users and CI jobs, I don't think that means more *processing*. We're just giving it more time to do the different steps, since we're clearly already up against (or, in the case of failure, past) the "edge" of time tolerances for managing everything with the MPI processes.

PR #1106, opened 2025-09-02
[WIP]: Refactor `cells_default.py`for simplicity and incoming new models

asoplata:
cc: @Chetank99 

This branch and PR should be the main place where we perform our refactoring of `cells_default.py`, in order to both better organize it, and, possibly, actually include Katharina's new celltypes. @Chetank99 and @katduecker are more than welcome to push commits *directly* to my branch for this PR, if desired (instead of having to do a PR on my personal fork to the branch).

I've added a working example of changing the L5 Pyramidal cell's creation functions into a self-contained "child class" of `Cell`, at the bottom of `cells_default.py`. This includes an example script at the top-level directory to show that the results output the same. Just to make it unambiguous that the script is *not* using the old `pyramidal` function of creating the L5 Pyramidal (or `_cell_L5pyr`), I have commented out some of its code.

Note: this will NOT currently pass testing for a number of reasons, including removing some functions, and changing an object's type.

### DIFF
--- a/hnn_core/cells_default.py
+++ b/hnn_core/cells_default.py
@@ -52,6 +52,7 @@ def _get_pyr_soma(p_all, cell_type):
         Ra=p_all[f"{cell_type}_soma_Ra"],
     )
 
+
 '''
 def _cell_L2Pyr(override_params, pos=(0.0, 0.0, 0), gid=0.0):
     """The geometry of the default sections in L2Pyr neuron."""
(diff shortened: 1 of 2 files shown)


asoplata:
@katduecker @Chetank99 Are we still planning on doing this? Do we want to proceed after Katharina's model is merged, or did we decide to drop this effort?

Chetank99:
@asoplata 
We never really had a follow-up discussion on this post-GSoC. I think it would be good to sync and decide next steps, happy to discuss how (or if, depending on y'all) we want to proceed.

PR #1107, opened 2025-09-02
[MRG] fix: add a DOI to linkcheck ignore

asoplata:
For some reason, recent linkchecks with this DOI have been failing, even though (I'm not 100% sure) it was working fine before. This change simply adds this particular publication DOI link to both linkcheck ignore lists.

Example failing can be found here: https://github.com/jonescompneurolab/hnn-core/actions/runs/17408169238/job/49423224430#step:5:486


dylansdaniels:
Good catch, looks good and will merge now

PR #1108, opened 2025-09-12
[MRG] doc: add hnn_io module to public API

asoplata:
Even if there are still design decisions and refactoring to do in the future for our file format versions and input/output, the existing IO in `hnn_io` SHOULD still be documented in the public API. This is important especially because `hnn_io.write_network_configuration` I believe (I could be wrong) is the ONLY way to export your `Network` to a file, which can then be loaded successfully into the GUI.

### DIFF
--- a/hnn_core/network_models.py
+++ b/hnn_core/network_models.py
@@ -19,7 +19,7 @@ def jones_2009_model(
     legacy_mode=False,
     mesh_shape=(10, 10),
 ):
-    """Instantiate the network model described in Jones et al. 2009 [1]_
+    """Instantiate the network model described in Jones et al. 2009
 
     Parameters
     ----------


doc/api.rst
```diff
@@ -100,8 +100,20 @@ Parallel backends (:py:mod:`hnn_core.parallel_backends`):
...
+-----------------------------------------------------------------
+.. currentmodule:: hnn_core.hnn_io
+
+.. autosummary::
+   :toctree: generated/
+
+   dict_to_network
+   network_to_dict
+   read_network_configuration
+   write_network_configuration
```
jasmainak, line 113: I haven't closely followed the developments but out of curiosity, what is the difference between `network_to_dict` and `write_network_configuration`? Could the user just write a json from the dictionary? You potentially want to avoid too many ways to do the same thing. Zen of Python:

> There should be one-- and preferably only one --obvious way to do it.

asoplata:
Hey Nick, I have added a commit with what could be one solution. The issues are twofold:

1. At https://jonescompneurolab.github.io/hnn-core/dev/api.html , like you point out first, there is a link to the citation in the docstring of the function, but it's formatted weirdly, as if it's the function's hash ending with `-1`. It does appear to work and take you to the citation, but it does look weird, I agree. If we remove the citation from the top-level line of the docstring, it will no longer show up here, but anyone who is interested in the citation will still be able to see the citation in the page for the function itself. In comparison, there was no link in earlier versions of the docs, shown here:

    image

2. The second this is multiple citations to a reference at the page for the function itself: https://jonescompneurolab.github.io/hnn-core/dev/generated/hnn_core.jones_2009_model.html#hnn_core.jones_2009_model . The citations to the references are "technically correct", but not user-friendly. This is another case where deleting the first citation (in the first, main line of the docstring) will also make things easier to read for users. Showing the full page here: 

    image

    Note that there are two citations to reference `[1]`, and only a single citation to reference `[2]`. The `(1,2)` links in the bibliography at the end are independent references to *each* of the multiple citations for the *same* reference. Instead, if we only have a single citation, then the `[1]` in the bibliography at the bottom will look just like the second reference link `[2]`. This is made more confusing by the fact that we only have 2 citations for ref 1, but we also have only 2 refs, so which goes to which is not immediately obvious to a user.

Therefore, for the reasons you brought this up, I think the best thing is to delete the first citation to reference 1 in this function, and only leave in the single citation inside the `Notes` section. This is what I've done in the latest commit.

PR #1126, opened 2025-09-19
[MRG] doc: add small tip&tricks to contrib

asoplata:


ntolley (approved):
Super useful! I find myself looking up these commands all the time as well

PR #1153, opened 2025-10-06
install: force nrn install < 9.0

asoplata:


asoplata:
The deployed TestPyPI version (`v0.4.3rc0`) *does* install correctly, while the current stable (`v0.4.2`) does not. I think we're good to go.

Issue #1141, opened 2025-09-19
ENH/INSTALL/DOC: Upon pip install, print out text that includes a link to the user survey


asoplata:
We should add output for the terminal that, when a user does a successful `pip install` of hnn-core, it prints out text thanking them for installing it, and more importantly displays the URL of our demographic survey that is on our frontpage, requesting that they fill it out.

asoplata:
Resolved by #1158

PR #1158, opened 2025-10-13
[MRG] Print survey link upon first launch of GUI or first import

sketch123456:
This PR is in regard to issue [#1141 ](https://github.com/jonescompneurolab/hnn-core/issues/1141). 

There is no reliable way to print a message after the user runs  `pip install` that works across all installation methods. It may be done if the user runs the verbose install (`pip install -v`), but this may not be entirely user-friendly.

As such, my approach to this problem was to print a link to the survey upon the first successful import or the first successful launch of the GUI. 

1. First Import: When a user runs `import hnn_core`, the message is displayed via `hnn_core/__init__.py`
2. First GUI Launch: When a user runs `hnn-gui` without importing the package directly, the message is displayed via `hnn_core/gui/__init__.py`

A persistent config file `~/.hnn_core_config.json` ensures the message is only shown once, regardless of which interaction path the user takes first.

Changes Made:
1. Created `hnn_core/_first_run.py` containing the `check_first_run()` function that:
- Checks if `~/.hnn_core_config.json` exists
- Prints the survey message if this is the user's first run
- Creates the config file to mark that the message has been shown
2. Updated `hnn_core/__init__.py` to call `check_first_run()` to catch first import scenarios
3. Updated `hnn_core/gui/__init__.py` to call `check_first_run()` to catch first GUI launch scenarios

Both implementations wrap the check in try-except blocks to ensure that any errors in displaying the message don't break core functionality.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ if __name__ == "__main__":
         'parallel': ['joblib', 'psutil'],
         'test': ['codespell', 'pytest', 'pytest-cov', 'pytest-xdist', 'ruff'],
         'docs': ['mne', 'myst-parser', 'nibabel', 'numpydoc', 'pillow',
-                 'pooch', 'pydata-sphinx-theme', 'sphinx', 'sphinx-gallery',
+                 'pooch<1.9.0', 'pydata-sphinx-theme', 'sphinx', 'sphinx-gallery',
                  'sphinx-copybutton', 'tdqm'],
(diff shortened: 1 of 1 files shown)


asoplata:
@ntolley @dylansdaniels This is now ready for review by another member of the team (since I edited a bunch of the code). The "flag" file for the survey is now saved to, essentially, `<environment>/<python-version>/site-packages/hnn_core/survey_seen.json`.

asoplata:
This will resolve #1141 .

PR #1114, opened 2025-09-16
doc: tiny: update NEURON URLs

asoplata:
This makes two very tiny changes: it changes (or appends to) existing links to the old URL of the NEURON website, https://www.neuron.yale.edu , and swaps/adds links to the new main NEURON site at https://nrn.readthedocs.io . The latter is the new location of the main NEURON website according to
https://github.com/neuronsimulator/nrn/issues/3602#issuecomment-3299864426

PR #1160, opened 2025-10-21
[MRG] Have linkcheck ignore NSG for SSL errors

asoplata:
Recently, we've been getting a single error across all of our Linkcheck runs, and these all have to do with some failed verification of the SSL certificate from https://www.nsgportal.org . See here for an example https://github.com/jonescompneurolab/hnn-core/actions/runs/18693857814/job/53306075052#step:5:675

>  (           index: line   65) broken    https://www.nsgportal.org/ - HTTPSConnectionPool(host='www.nsgportal.org', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1016)')))

However, when I browse to the site and inspect the cert, or use a third-party "SSL inspection" site, everything is fine. I first tried using `pip` install of Mozilla's Root Certificates ( https://pypi.org/project/certifi/ ) to resolve the issue in case it was related to a local cert problem, but that didn't fix it.

Finally, I ended up simply adding it to the Linkcheck ignore sets, both for our main linkcheck check and for the extended manual check. This is probably a good idea anyways, since we have had linkcheck fail against their URL in the past for other reasons in a random manner, and we don't want to hammer their servers anymore than necessary.

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -2,10 +2,10 @@ name: Linkcheck
 
 on:
   push:
-  #   branches:
-  #     - master
-  # schedule:
-  #   - cron: "0 0 1,15 * *"
+    branches:
+      - master
+  schedule:
+    - cron: "0 0 1,15 * *"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}


doc/conf.py
```diff
@@ -248,6 +248,7 @@
 
 linkcheck_ignore = [
     'https://neuron.yale.edu/neuron',
+    'https://www.nsgportal.org',
```
ntolley: Is there any way we would know if these links should be removed from their ignore status? Or is it just a permanent demotion :smiling_face_with_tear:

asoplata: I think permanent is best, in part because these sites tend to have random reliability problems as well: recall that our own old HNN webpage hosted by Brown would constantly fail, etc. If anything, we're reducing unnecessary traffic to potentially-brittle servers. Also, eventually we could probably drop the NEURON url from the ignore since that's their old URL which they're trying to move away from (IIRC), and I already tried to update the existing links to use the newer URL in #1114

PR #1168, opened 2025-10-23
[WIP] Experimental duecker_ET model branch

katduecker:
This is the latest version of the duecker_2025_ET_model, rebased to the current state of the branch. The model is open to all HNN developers at their own risk and discretion. This model is still under development, not peer-reviewed, and regular code changes are expected.

I will add guidelines on how to use the model to this PR. Please reach out to me if you have any questions!


ntolley:
Looking forward to giving this a spin!

hnn_core/cells_default.py
```diff
@@ -319,7 +385,7 @@ def _exp_g_at_dist(x, zero_val, exp_term, offset):
 
     """
 
-    return zero_val * np.exp(exp_term * x) + offset
+    return zero_val * (slope * np.exp(exp_term * x)  + offset)
```
asoplata: There may be a bug here: `offset` is now multiplied by `zero_val`, but it wasn't before. In the current `master` code, `offset` is always 0, so this change does not break current tests. However, that doesn't mean it's not a bug; @katduecker are you sure that `offset` should be multiplied by `zero_val`? I can understand that that may be your very reason for introducing `slope`, but just wanted to double-check.

PR #1169, opened 2025-10-23
[MRG] fix: opt: use safe default n_trials for rmse opt

asoplata:
@carolinafernandezp @ntolley  This addresses the optimization parameter regression-bug accidentally introduced by #1057 , where n_trials was changed to being explicitly required to be set by the user if using `dipole_rmse` optimization. Instead, n_trials is set by 1 if not provided.

Nick or Carolina, can you please review this ASAP? It is a tiny fix

### DIFF
--- a/hnn_core/optimization/general_optimization.py
+++ b/hnn_core/optimization/general_optimization.py
@@ -156,13 +156,13 @@ class Optimizer:
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


dylansdaniels:
@asoplata this looks fine i can hit merge if you're ok with it

Issue #962, opened 2025-01-10
MAINT: Upgrade ALL network configuration file formats (except for tests)


asoplata:
Currently, the newest version of the network configuration files are the “hierarchical JSON” format, which are required for any networks loaded using the GUI. We should complete the “migration” of all network configuration files we actually expect users to use into this format, and either remove or archive all “legacy” config files, including the “flat JSON” format and the original “param” format (except for use in testing).

Part of why this is important is that if a user is using the GUI, then all JSON network config files they potentially interact with, such as https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/param/default.json , EXCEPT for https://github.com/jonescompneurolab/hnn-core/tree/master/hnn_core/tests/assets , should be using the latest format.

Once this is done, we should also include a `config_version` attribute into the format in order to track version differences of the config files themselves.

dikshant182004:
hey @asoplata ,i have understood the task there is one question in my mind that should i archive all that legacy files in another directory for historical use or delete them .i am thinking to transfer them in another directory and update the .gitignore or documentation that they are not for regular use

dikshant182004:
one more thing as we are migrating into hierarchial json format ,does the gui script is able to handle the nested structure or we need to update that also

asoplata:
Hello Dikshant! Welcome, and thank you very much for your interest in contributing to HNN-Core's development! We're very happy to have you here :)

Firstly, our Contribution Guide can be found here: https://jonescompneurolab.github.io/hnn-core/stable/contributing.html . Judging from your Python experience, you should already be pretty familiar with how we do things. We do rebase+squash instead of merges.

Secondly, our Code of Conduct can be found here: https://github.com/jonescompneurolab/hnn-core/blob/master/CODE_OF_CONDUCT.md .

Lastly, if you are interested in contributing to development, we strongly recommend that you start with an existing Issue that has the `good first issue` tag. Some examples are #954, #917, #966, and #967. Unfortunately, this current issue is not a good fit for a new developer, since there's certain complexities for how to do this that I didn't include in the initial issue description.

dikshant182004:
hii @asoplata thanks for guiding me i will start with the good first issue first then will move forward .

dikshant182004:
hii @asoplata ,can we move forward in this issue now ,as the `currentscape` needs to be on hold.

asoplata:
[...]

PR #1180, opened 2025-11-03
[WIP] Fix bugs in read_params workflow

katduecker:
I have been experiencing issues when loading in network parameters from a json file, somehow the simulation always looked different. There were a couple of small bugs due to hard-coded synaptic delays in params.py. There is also an update of the event_seeds that is only applied when reading params from a file, but not when defining an event seed. I have commented it out for now but we'd have to decide if it's needed.


asoplata:
I have made a 95% complete PR that applies one possible solution at https://github.com/katduecker/hnn-core/pull/2 .

asoplata:
This comment first documents how the implicit `event_seed`'s (called "seeds") of various external drives work currently (currently as `master` https://github.com/jonescompneurolab/hnn-core/commit/90f4d636b13647869f11825daf40eb985b6965cb ), in the case of passing `add_drives_from_params=True` to `jones_2009_model()`, both with and without `legacy_mode=True`. This comment then later describes what (few) changes we are making to enable `legacy_mode` to work as it did before, while enabling non-`legacy_mode` to use user-provided seeds *without modification*, enabling reproducibility. This comment is also intended to act as documentation for these changes, and to be referred to by comments in the code. You'll want to sit down for this one.

It is easiest if we begin by explaining how things currently work on `master`, as of https://github.com/jonescompneurolab/hnn-core/commit/90f4d636b13647869f11825daf40eb985b6965cb :

## Current add-drives, no legacy

Let's assume you are doing the following code example:
```python
net = jones_2009_model(add_drives_from_params=True)  # NO legacy_mode
dpl = simulate_dipole(net, tstop=170.0)
```

As far as I can tell, all code for loading, changing, and setting the unique `event_seed` of each drive is contained entirely in the appropriately-named `drives.py::_add_drives_from_params()` [here](https://github.com/jonescompneurolab/hnn-core/blob/90f4d636b13647869f11825daf40eb985b6965cb/hnn_core/drives.py#L156). Let's go through this function.

1. First, we call `params.py::_extract_drive_specs_from_hnn_params`. This loads our parameters from `param/default.json`, where the initial seed for all drives is 2:
https://github.com/jonescompneurolab/hnn-core/blob/90f4d636b13647869f11825daf40eb985b6965cb/hnn_core/param/default.json#L18-L20
Because `legacy_mode` is off, we only load the three typical drives, along with their initial seeds:
    - evdist1: `2`
    - evprox1: `2`
    - evprox2: `2`

2. Next, each drive is then added to the Network in the main for-loop. I will note, however, that they are added with the above `event_seed`, but these seeds will be overwritten later.

3. Finally, we get to the bottom drive-related block. This adds `net.gid_ranges[drive_name][0]` to the seeds of each of these three drives. In the default case, this changes the seeds such that they become the following:
    - evdist1: `272` = 2 (initial event_seed) + 270 (initial gid)
    - evprox1: `507` = 2 (initial event_seed) + 505 (initial gid)
    - evprox2: `777` = 2 (initial event_seed) + 775 (initial gid)

4. As far as I can tell, there are no more changes to the drives' seeds, and these become the final seed values used in this case. These are probably the seeds that many people are actually using, unless they add their drives manually or specify their own `params` dictionary or "flat JSON" file.

## Current add-drives, yes legacy

`legacy_mode=True` makes things much more complicated. Let's assume you are doing the following code example, and go through `drives.py::_add_drives_from_params()` again.

```python
net = jones_2009_model(add_drives_from_params=True, legacy_mode=True)
dpl = simulate_dipole(net, tstop=170.0)
```

1. First, we call `params.py::_extract_drive_specs_from_hnn_params`. This loads our parameters from `param/default.json`, where the initial seed for all drives is 2:
https://github.com/jonescompneurolab/hnn-core/blob/90f4d636b13647869f11825daf40eb985b6965cb/hnn_core/param/default.json#L18-L20
However, `legacy_mode` makes *further* changes to our three "typical" drives and [decrements them by 18 here](https://github.com/jonescompneurolab/hnn-core/blob/90f4d636b13647869f11825daf40eb985b6965cb/hnn_core/params.py#L243). `legacy_mode` also adds more drives than just the typical three (this will be important later). The drives' seeds at this point are as follows:
    - bursty1:  `2`
    - bursty2:  `2`
    - evdist1:  `-16` = 2 (initial event seed) - 18 (legacy offset)
    - evprox1:  `-16` = 2 (initial event seed) - 18 (legacy offset)
    - evprox2:  `-16` = 2 (initial event seed) - 18 (legacy offset)
    - extgauss: `2`
    - extpois:  `2`

2. Next, each drive is then added to the Network in the main for-loop like before. Again, its seeds at this point will be overwritten later.

3. Finally, we get to the bottom drive-related block. This adds `net.gid_ranges[drive_name][0]` to the seeds of every drives. However, since `bursty1` and `bursty2` are drives specific to `legacy_mode`, they are created first. This has a downstream effect where it changes the `gid_range`-related addition to the seeds of ALL other drives, including the typical three. Yet another difference is that in the `legacy_mode=True` case, `evdist1` has a gid_range length of 270 instead of the typical 235, which will cause `evprox1` and `evprox2`'s initial gid ranges to begin at a higher number than in the non-`legacy_mode` case. In summary, what this means is that the seeds of the typical three drives are very different from the `legacy_mode=False` case. The drives' seeds at this point are as follows:
    - bursty1:   `272` = 2 (initial event seed) +  270 (initial gid)
    - bursty2:   `282` = 2 (initial event seed) +  280 (initial gid)
    - evdist1:   `274` = 2 (initial event seed) +  290 (initial gid) - 18 (legacy offset)
    - evprox1:   `544` = 2 (initial event seed) +  560 (initial gid) - 18 (legacy offset)
    - evprox2:   `814` = 2 (initial event seed) +  830 (initial gid) - 18 (legacy offset)
    - extgauss: `1102` = 2 (initial event seed) + 1100 (initial gid)
    - extpois:  `1372` = 2 (initial event seed) + 1370 (initial gid)

4. As far as I can tell, there are no more changes to the drives' seeds, and these become the final seed values used in this case.

## New `event_seed` code

In the new commit https://github.com/katduecker/hnn-core/commit/1931f5814304a41227b185b84353dcffc6e4c958 , we make the following changes, which will allow for per-drive seeds to be faithfully used, except in the case when both `legacy_mode` and `add_drives_from_params` are True. We:

1. Update the `param/default.json` file to use the *true* seeds in the non-`legacy_mode` case, i.e.:
    ```
    "prng_seedcore_evprox_1": 507,
    "prng_seedcore_evdist_1": 272,
    "prng_seedcore_evprox_2": 777,
    ```
2. Change the gid_range addition at https://github.com/jonescompneurolab/hnn-core/blob/90f4d636b13647869f11825daf40eb985b6965cb/hnn_core/drives.py#L241 so that only uses the old "gid_range" addition method when `legacy_mode=True`, AND so it uses the original values of the seeds of the three typical drives *at that point in the code*. To do this, we change the code from
    ```python
    for drive_name, drive in net.external_drives.items():
        drive["event_seed"] += net.gid_ranges[drive_name][0]
    ```
    to:
    ```python
    if net._legacy_mode:
        for drive_name, drive in net.external_drives.items():
            if drive_name in ("evdist1", "evprox1", "evprox2"):
                drive["event_seed"] = -16
            drive["event_seed"] += net.gid_ranges[drive_name][0]
    ```

3. Regenerate our default hierarchical-JSON networks accordingly.

asoplata:
For the synaptic delays...we may want to split that into its own PR/issue. As of the most recent commit b94978a , the only remaining test that fails is one where we test that the dipole output is identical. Specifically, that test is `test_parallel_backends.py::TestParallelBackends::test_compare_hnn_core` https://github.com/jonescompneurolab/hnn-core/blob/90f4d636b13647869f11825daf40eb985b6965cb/hnn_core/tests/test_parallel_backends.py#L371 

In that test, we compare the dipole output of the typical sim to existing dipole output data here, on the `test_data` branch https://github.com/jonescompneurolab/hnn-core/blob/test_data/dpl.txt . Unsurprisingly, since the network is ever-so-slightly different, the dipoles are not an exact match. However, when I plotted them, they were more different than I was expecting, as shown in the below figure. In this figure, `dpl` is the existing dipole test data (i.e. excluding the synaptic delay changes), while `dpl2` is the new dipole (i.e. including the synaptic delay changes). 
image

To my untrained eye, these are different enough for concern, but I'm not an expert at interpretation of this. Possibilities for what to do next:

1. We don't make this synaptic delay change at all: probably a bad idea.
2. We decide this is close enough, and upload new dipole data to be the new default that we test against. In this case, we would add the output `dpl2.txt` data to a new file named something like `dpl-updated-synaptic-delays.txt` or whatever, on the `test_data` branch.
3. We "kick the can down the road": we separate the synaptic delays out from this issue/PR into its own, and investigate the differences more thoroughly. The first thing I would do in this case is reproduce everything in the ERP tutorials and check that they are qualitatively similar. Such an investigation could be as detailed as the investigator decides it should be.

asoplata:
cc @ntolley @katduecker @dylansdaniels 

Discussion of this issue spawned discussion on at least 3 separate points:

1. Removing the "hardcoding" / overwriting of values in `params.py::create_pext`: I believe this is the only task that should be done in this PR specifically. This should remove all (or as many as possible) the over-writing values that are used in the code, where the code has read in the typical "flat JSON" file `param/default.json`, and then overrides the variable read in. As Katharina found, most or all synaptic delays appear to be over-written, meaning that their values in `param/default.json` are not used. The solution is to remove the hard-overwriting code, and instead replace it with code that reads the parameter from flat JSON file. Simultaneously, we should make changes to the flat JSON file such that it contains the values that the code was *over-writing to*. This will allow the Network that is made from the flat JSON file to be more accurate to the file itself.

2. Updating incorrect synaptic delay parameters: As we discussed, whether or not a parameter is over-written in the code is a different problem than that value's scientific accuracy. We should make a separate issue that:
    1. Investigates, at minimum, what the synaptic delays were (in which sources), and what they should be. Especially including sources of the original code we're coming from (including any applicable Jones 2007/2009 code and HNN-Original).
    2. For values where the current `hnn_core` value is wrong, we THEN need to investigate just how much of a difference these make for all uses of our "default" simulation dipole output. (E.g. "Are there qualititative differences for the simulations currently on our Tutorials?") etc.

3. Eventually deprecating the "flat JSON" file and its related `params.py`.  For that, see this issue, starting from here: https://github.com/jonescompneurolab/hnn-core/issues/962#issuecomment-3486857828

asoplata:
I think we should add new keys for the synaptic delay variables in order to put them into `param/default.json`. For example,  they could resemble a new key like other keys, such as `syndelay_evprox_1_L2Basket_ampa` in `param/default.json`; in this case, in the code you'd replace `0.1` with something like `p["syndelay_" + skey + "_L2Basket_ampa"]` like its nearby code.

Additionally, just to be clear, I think we should keep the synaptic delay value themselves *the same as they are in `master`* and *not* change the delay values themselves in this PR. This PR should only be for the code in `params.py` to not hardcode as many variables. Any code that changes the synaptic delay values themselves (and thus changes Dipole output) should go in a new, as-yet-uncreated PR, tied to #1194

asoplata:
Just for reference, as of https://github.com/katduecker/hnn-core/commit/247e10a63aaed4dee3e96f6e67b6d135da8f5b12

Differences that KD has found (not exhaustive) include:
- the three synaptic delays previously discussed for the drives
- in `default.json`,  `"gbar_L2Pyr_L5Basket": 0.00025,` should be changed to `"gbar_L2Pyr_L5Basket": 0.02,`

This is related the fixes we're applying to #1227 in preparation for getting it ready for merge

PR #1181, opened 2025-11-04
[MRG] doc: minor addition of google colab to release doc

asoplata:


asoplata:
Turns out our API GColab notebook was broken too... (albeit in an easily fixable way)

PR #1183, opened 2025-11-05
[MRG] GUI add and test "smart" loading of global gains

asoplata:
This resolves the undocumented issue we discussed in the Dev meeting on 2025-10-27,
Notion meeting notes here:
https://www.notion.so/jonescompneurolab/HNN-Core-dev-meeting-2025-10-27-29644dbbcce6800c8275de7b12260b8f where we discussed how, if a user makes a change in the GUI to a Global synaptic gain value, then downloads that Network to a file, and then uploads the same Network to continue where they left off, we needed better handling of the loaded synaptic weights.

Before, since every connection has only one gain value according to the API/NEURON, each single gain entry in the Network Connectivity "accordion" was loaded with whatever the value is in the Network file. In the above case, this would have been whatever they changed the Global gain to. However, in the GUI, after loading the Network file, the Global synaptic gains would have been reset to their default values, which could confuse users, and would make it very inconvenient to make further Global synaptic gain modifications.

Instead, what this commit does is, when a user uploads a Network file (or when the GUI first initializes), the global gains of that Network are checked. If they pass the
`_check_global_synaptic_gains_uniformity` check (discussed in #918), meaning that all global gains are uniform within their class, AND if those global gains appear to not be the default values, then the GUI stores the Network's custom global gains for display later, but then overwrites the *single* gains used in the Network object. This is because, when the user later runs a sim or downloads the Network anew, the GUI will create a *new* Network object using the various gain value states stored in the GUI itself, *not* in the loaded Network object. This will then re-apply the global gains all over again, but make it easier for the user to see and understand the current state.

If the above check *fails*, then we should *not* do "smart" loading of the global gains, because the uniformity assumption is violated. Instead, in this case, the GUI global gains will simply display their default values, the user will receive a warning in the log window that the assumption is violated, and the single gain entries in the Network Connectivity accordion *will* display their individual values.

### DIFF
--- a/hnn_core/gui/gui.py
+++ b/hnn_core/gui/gui.py
@@ -4384,7 +4384,7 @@ def add_network_connectivity_tab(
     ### Global synaptic gains
     # ---------------------------------------------------------------
     global_gain_out.clear_output()
-    gain_types = ("e_e", "e_i", "i_e", "i_i")
+    gui_global_gain_types = ("e_e", "e_i", "i_e", "i_i")
 
     # Do "smart" loading of the synaptic gains, where custom global gains are
     # "extracted" from the network and displayed in the "global gains" parts of the GUI,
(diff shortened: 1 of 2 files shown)


asoplata:
This needs to be rebased after #1190 is merged

asoplata:
This has been rebased and is ready for merge

asoplata:
- [x] TODO self: expand tests

PR #1185, opened 2025-11-06
[MRG] feat: Extracting Katharina's voltage-initialization changes

asoplata:
This is the first commit of many to come where I'm going to extract the parts of #1168 relevant to the new, proper voltage initialization.

The plan is:
1. We'll separate the changes to the voltage initialization stuff (and *nothing else*) and make it its own PR onto `master` (this PR),
2. Update the tests that become broken as we go along,
3. After all the updates are applied, then everyone reviews the changes (which will likely end up in a slightly different form from how they currently exist in #1168),
4. Then we merge these voltage initialization changes into `master`,
5. Then finally, we merge/rebase the voltage initialization changes (which are now on `master`) *back* into the code at #1168 .
6. Repeat the process for all other code features included in the lengthy #1168 work, until everything gets merged, piece-by-piece.

For starters, this first commit consists of *only* the extracted changes to `cell.py` and nothing else. Fortunately, in this case, the tests seem to pass without having to change anything.

### DIFF
--- a/hnn_core/cells_default.py
+++ b/hnn_core/cells_default.py
@@ -384,7 +384,7 @@ def _cell_L5Pyr(override_params, pos=(0.0, 0.0, 0), gid=0):
     sections["soma"] = _get_pyr_soma(
         p_all,
         "L5Pyr",
-        v_init=-72,
+        v_init=v_init["soma"],
     )
 
     end_pts = {
--- a/hnn_core/tests/test_cells_default.py
+++ b/hnn_core/tests/test_cells_default.py
@@ -245,7 +245,7 @@ def test_basket_voltage_init():
     # Current section initial voltage, for soma (only section) of both basket types. The
     # original location of this value can be found at
     # https://github.com/jonescompneurolab/hnn-core/blob/8a0fffef8d8803e2404d7237f9adeabecd1285ed/hnn_core/network_builder.py#L681
-    expected_basket_v0 = -64.9737
+    expected_basket_v0 = {"soma": -64.9737}
 
     # Initial setup
     load_custom_mechanisms()
@@ -256,10 +256,13 @@ def test_basket_voltage_init():
 
     # Test initial voltages (v0) for basket cells after HNN-Core Cell class creation,
     # but before NEURON cell building
-    for cell_name, cell in cells.items():
-        assert np.isclose(cell.sections["soma"].v0, expected_basket_v0), (
-            f"{cell_name} soma v0={cell.sections['soma'].v0}, expected {expected_basket_v0}"
-        )
+    for _, cell in cells.items():
+        for sec_name, sec in cell.sections.items():
+            v0 = sec.v0
+            expected_v0 = expected_basket_v0[sec_name]
+            assert np.isclose(v0, expected_v0), (
+                f"HNN-Core Basket cell {sec_name} v0={v0}, expected {expected_v0}"
+            )
 
     # Build NEURON cells and sections, then initialize
     for cell in cells.values():
@@ -268,11 +271,13 @@ def test_basket_voltage_init():
 
     # Test that the initial voltages of the built NEURON sections (at the midpoint)
     # still match our expected values
-    for cell_name, cell in cells.items():
-        soma_v = cell._nrn_sections["soma"](0.5).v
-        assert np.isclose(soma_v, expected_basket_v0, atol=0.1), (
-            f"{cell_name} soma initial voltage is {soma_v}, expected {expected_basket_v0}"
-        )
+    for _, cell in cells.items():
+        for sec_name in cell._nrn_sections:
+            v = cell._nrn_sections[sec_name](0.5).v
+            expected_v = expected_basket_v0[sec_name]
+            assert np.isclose(v, expected_v, atol=0.1), (
+                f"NEURON-built Basket cell {sec_name} initial voltage is {v}, expected {expected_v}"
+            )
 
 
 def test_l2pyr_voltage_init():
@@ -290,7 +295,7 @@ def test_l2pyr_voltage_init():
     # but before NEURON cell building
     for sec_name, sec in l2pyr.sections.items():
         assert np.isclose(sec.v0, expected_l2pyr_v0), (
-            f"L2Pyr {sec_name} v0={sec.v0}, expected {expected_l2pyr_v0}"
+            f"HNN-Core L2Pyr {sec_name} v0={sec.v0}, expected {expected_l2pyr_v0}"
         )
 
     # Build NEURON cells and sections, then initialize
@@ -302,7 +307,7 @@ def test_l2pyr_voltage_init():
     for sec_name in l2pyr._nrn_sections:
         v = l2pyr._nrn_sections[sec_name](0.5).v
         assert np.isclose(v, expected_l2pyr_v0, atol=0.1), (
-            f"L2Pyr {sec_name} initial voltage is {v}, expected {expected_l2pyr_v0}"
+            f"NEURON-built L2Pyr {sec_name} initial voltage is {v}, expected {expected_l2pyr_v0}"
         )
 
 
@@ -330,9 +335,10 @@ def test_l5pyr_voltage_init():
     # Test initial voltages (v0) for L5Pyr cells after HNN-Core Cell class creation,
     # but before NEURON cell building
     for sec_name, sec in l5pyr.sections.items():
+        v0 = sec.v0
         expected_v0 = expected_l5pyr_v0[sec_name]
-        assert np.isclose(sec.v0, expected_v0), (
-            f"L5Pyr {sec_name} v0={sec.v0}, expected {expected_v0}"
+        assert np.isclose(v0, expected_v0), (
+            f"HNN-Core L5Pyr {sec_name} v0={v0}, expected {expected_v0}"
         )
 
     # Build NEURON cells and sections, then initialize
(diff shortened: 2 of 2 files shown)


asoplata:
I think this is ready for review and merge as-is. Note that as far as I can tell, this does *NOT* change the initial conditions of any simulation, and output appears to be identical to how it was before. (You can check this with one of the final tests, which is `pytest hnn_core/tests/test_parallel_backends.py::TestParallelBackends::test_compare_hnn_core`, and directly compares simulation output to saved dipole data.)

I was planning on writing tests for the new voltage-initialization, but as it turns out we have very, very little testing of `cells_default.py` in total. Many of the docstrings in `cells_default.py` are also extremely out of date, limited, or incorrect.

*However*, since #1168 includes a *large* number of changes to `cells_default.py` which are independent of voltage-initialization, I think it's better to defer expansion of testing and documenting for that file to a subsequent PR. There will be soon/eventually be a subsequent PR that introduces the main Duecker_ET model itself, and that will include all the changes in `cells_default.py` relevant to that model. I think it makes more sense to re-evaluate the testing of the file at that time, since there may be some opportunities for refactoring that make writing the tests easier (or not).

asoplata:
Force-pushed in order to remove an incorrect statement in an earlier commit message. (I previously thought that this would change the initial membrane voltages, but I see Katharina went to pains to *not* do that, and it's correct that the initial membrane voltages for each section are now still the same as they were under the old code).

asoplata:
Another force-push in order to label @katduecker as the author of most of the commits, since she was the one who wrote the code originally (and this way she gets explicit credit). Is this okay with you Katharina?

No more force-pushes should be necessary.

hnn_core/cells_default.py
```diff
@@ -34,22 +36,64 @@ def _get_dends(params, cell_type, section_names):
...
             diam=dend_prop["diam"],
             Ra=dend_prop["Ra"],
             cm=dend_prop["cm"],
+            v=v,
         )
     return sections
 
 
-def _get_pyr_soma(p_all, cell_type):
+# KD: In the new model, the basal dendrites are differently tuned from the apical
```
katduecker: My fault , but should it read "Tuning of the basal dendrites in duecker_ET_model is different from apical dendrites"?

hnn_core/cells_default.py
```diff
@@ -230,9 +308,16 @@ def _cell_L5Pyr(override_params, pos=(0.0, 0.0, 0), gid=0.0):
     )
 
 
-def _get_basket_soma(cell_name):
+def _get_basket_soma(cell_name, v_init=-64.9737):
```
ntolley:
```suggestion
def _get_basket_soma(cell_name, v_init=-64.9737):
```
Nitpick but is there any reason to have 4 decimal points here rather than 2 like most of the other voltages?

hnn_core/cells_default.py
asoplata: What about

```
# KD: In the new `duecker_ET_model` model, tuning of the basal dendrites is different
# from the tuning of the apical dendrites.
```

hnn_core/cells_default.py
```diff
@@ -15,51 +15,203 @@
...
+        '{cell_type}_{section}_{property}' (e.g., 'L2Pyr_apicaltrunk_L'), where
+        key-value pairs are only provided for those values where the user wants to use
+        custom, non-default parameters. The default parameters can be found in
+        `params_default.py::get_L2Pyr_params_default`. If no overrides are desired, then
+        this argument should be None.
+    pos : tuple of (float, float, int), default=(0.0, 0.0, 0)
+        Coordinates of cell soma in xyz-space.
+    gid : int, default=0
+        The unique, "global ID" (GID) of the cell. Once the GID is set, it cannot be
+        changed.
```
ntolley:
```suggestion
        The unique, "global ID" (GID) of the cell.
```
Wouldn't #1096 make this statement incorrect? Perhaps just leave it out...

hnn_core/cells_default.py
```diff
@@ -168,6 +401,7 @@ def _cell_L5Pyr(override_params, pos=(0.0, 0.0, 0), gid=0.0):
         "basal_3": [[0, 0, -50], [106, 0, -156]],
     }
 
+    # AES TODO: what's up with this units comment?
```
ntolley: it's definitely referring to the way conductance is encoded in the mechanisms in the corresponding mod files, but there is a mismatch in the number of units listed in the comment and the variable below

hnn_core/cells_default.py
```diff
@@ -230,12 +464,76 @@ def _cell_L5Pyr(override_params, pos=(0.0, 0.0, 0), gid=0.0):
     )
 
 
-def _get_basket_soma(cell_name):
+def _get_basket_soma(cell_name, v_init=-64.9737):
```
ntolley:
```suggestion
def _get_basket_soma(v_init=-64.9737):
```
We should definitely get rid of this right? If it's not used as indicated below

hnn_core/cells_default.py
```diff
@@ -368,20 +700,29 @@ def basket(cell_name, pos=(0, 0, 0), gid=None):
...
+        custom, non-default parameters. The default parameters can be found in either
+        `params_default.py::get_L2Pyr_params_default` or
+        `params_default.py::get_L5Pyr_params_default`. If no overrides are desired, then
+        this argument should be None.
     gid : int or None (optional)
-        Each cell in a network is uniquely identified by it's "global ID": GID.
-        The GID is an integer from 0 to n_cells, or None if the cell is not
-        yet attached to a network. Once the GID is set, it cannot be changed.
+        The unique, "global ID" (GID) of the cell. Once the GID is set, it cannot be
+        changed.
```
ntolley:
```suggestion
        The unique, "global ID" (GID) of the cell.
```

hnn_core/tests/test_cells_default.py
```diff
@@ -238,3 +238,128 @@ def test_linear_g_at_dist():
...
+        expected_v0 = expected_l5pyr_v0[sec_name]
+        assert np.isclose(sec.v0, expected_v0), (
+            f"L5Pyr {sec_name} v0={sec.v0}, expected {expected_v0}"
+        )
+
+    # Test that v0 values are correctly applied to built NEURON sections
+    l5pyr.build(sec_name_apical="apical_trunk")
+
+    # After building, NEURON sections should have v initialized to v0
+    # Check a few key sections with different v0 values
```
ntolley: why not do a for loop like above? seems like it would take even less code for a more thorough test

hnn_core/tests/test_cells_default.py
```diff
@@ -238,3 +238,128 @@ def test_linear_g_at_dist():
...
+    apical_2_nrn_sec = l5pyr._nrn_sections["apical_2"]
+    apical_tuft_nrn_sec = l5pyr._nrn_sections["apical_tuft"]
+    soma_nrn_sec = l5pyr._nrn_sections["soma"]
+
+    # Initialize with finitialize() which should use the v0 values
+    h.finitialize()
```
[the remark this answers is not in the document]
ntolley: actually in the function above this is placed where I would expect to see it

hnn_core/cells_default.py
```diff
@@ -134,8 +297,56 @@ def _cell_L2Pyr(override_params, pos=(0.0, 0.0, 0), gid=0.0):
...
+        '{cell_type}_{section}_{property}' (e.g., 'L5Pyr_apicaltrunk_L'), where
+        key-value pairs are only provided for those values where the user wants to use
+        custom, non-default parameters. The default parameters can be found in
+        `params_default.py::get_L5Pyr_params_default`. If no overrides are desired, then
+        this argument should be None.
+    pos : tuple of (float, float, int), default=(0.0, 0.0, 0)
+        Coordinates of cell soma in xyz-space.
+    gid : int, default=0
+        The unique, "global ID" (GID) of the cell. Once the GID is set, it cannot be
+        changed.
```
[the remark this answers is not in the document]
asoplata: Change applied or resolved in a new commit.

hnn_core/cells_default.py
asoplata: Change applied or resolved in a new commit.

hnn_core/tests/test_cells_default.py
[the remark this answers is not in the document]
asoplata: Yeah, I probably made these with Claude, but didn't notice that the initxn was inconsistent. I did the refactors you mentioned in your other comments on these tests, and made the initxn consistent, and greatly improved the comments documenting. All three test functions are now very similar and clean.  Change applied or resolved in a new commit.

hnn_core/tests/test_cells_default.py
```diff
@@ -238,3 +238,111 @@ def test_linear_g_at_dist():
...
+
+    # Test that the initial voltages of the built NEURON sections (at the midpoint)
+    # still match our expected values
+    for cell_name, cell in cells.items():
+        soma_v = cell._nrn_sections["soma"](0.5).v
+        assert np.isclose(soma_v, expected_basket_v0, atol=0.1), (
+            f"{cell_name} soma initial voltage is {soma_v}, expected {expected_basket_v0}"
+        )
+
```
ntolley:
```suggestion
    for cell_name, cell in cells.items():
        for sec_name in cell._nrn_sections:
            v = cell._nrn_sections[sec_name](0.5).v
            assert np.isclose(v, expected_basket_v0, atol=0.1), (
                f"{cell_name} {sec_name} initial voltage is {v}, expected {expected_basket_v0}"
            )
```
This is equivalent since basket cells just have one section, but this is better in the future if we ever extend this test to include other types of basket cells with more compartments (i.e. you just need to add to the list)

hnn_core/cells_default.py
```diff
@@ -153,8 +362,30 @@ def _cell_L5Pyr(override_params, pos=(0.0, 0.0, 0), gid=0.0):
...
+    sections = _get_dends(
+        p_all,
+        cell_type="L5Pyr",
+        section_names=section_names,
+        v_init=v_init,
+    )
+    sections["soma"] = _get_pyr_soma(
+        p_all,
+        "L5Pyr",
+        v_init=-72,
```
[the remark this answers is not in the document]
asoplata: done

PR #1188, opened 2025-11-07
[MRG]: feat: KD add remaining extracted non-model code

asoplata:
I believe this adds the last of the code that is extracted from #1168 that is *not* the actual Duecker model itself, but instead code changes that are used by the default model. These functions are not currently tested at all, and I have not added tests for them for similar reasons to https://github.com/jonescompneurolab/hnn-core/pull/1185#issuecomment-3502924088

After this, we should be able to make a PR that adds the actual Duecker model itself. However, there will be additional work for that step: we need to update `hnn_io.read_network_configuration` and the equivalent `...write...` such that they support the Duecker model.

Continuing the pattern, this is based on top of #1187.

### DIFF
--- a/hnn_core/tests/test_cells_default.py
+++ b/hnn_core/tests/test_cells_default.py
@@ -3,7 +3,7 @@ import pytest
 from neuron import h
 import numpy as np
 
-from hnn_core.cells_default import pyramidal, basket
+from hnn_core.cells_default import pyramidal, basket, _exp_g_at_dist, _linear_g_at_dist
 from hnn_core.network_builder import load_custom_mechanisms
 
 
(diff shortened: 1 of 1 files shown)


hnn_core/cells_default.py
```diff
@@ -314,12 +399,16 @@ def _exp_g_at_dist(x, zero_val, exp_term, offset):
...
-    offset: float |int
+    offset : float | int
         Offset value added to output
-
+    slope : int | float, default=1
+        Slope of the exponential component
     """
-
-    return zero_val * np.exp(exp_term * x) + offset
+    # AES: TODO Should the offset be multiplied by zero_val? This was also changed in
```
katduecker: Yes, I double-checked that this is indeed what I wanted it to do. The distribution of the Ih that I need this function for in the new model is:

gbar = zero_val * (-0.8696 + 2.087 * np.exp(x/323))

Taken from Hay et al., 2011.

Actually, maybe we shouldn't call it zero_val, but gbar_proximal or something like that.

hnn_core/cells_default.py
```diff
@@ -303,23 +303,24 @@ def _get_mechanisms(p_all, cell_type, section_names, mechanisms):
...
-def _exp_g_at_dist(x, zero_val, exp_term, offset):
+def _exp_g_at_dist(x, gbar_at_zero, exp_term, offset, slope=1):
     """Compute exponential distance-dependent ionic conductance.
 
     Parameters
     ----------
     x : float | int
         Distance from soma
-    zero_val : float | int
+    gbar_at_zero : float | int
```
katduecker: Yep this variable name makes a lot more sense

Issue #1203, opened 2025-12-16
Flip the order of `set_params()` arguments for `BatchSimulate`


ntolley:
Currently, `BatchSimulate` and `Optimizer` require a `set_params()` function which accepts a `Network` object, and  a dictionary called `param_values` which is used to update different parameter values in the network during batch simulation or optimization.

The `BatchSimulate` class expects the function to be:
```python
def set_params(param_values, net):
```

Whereas the `Optmizer` class expects a different order of the arguments:
```python
def set_params(net, param_values):
```

I propose changing the `BatchSimulate` class to expect `set_params(net, param_values)`. I think this makes the most sense, and I believe was the original intention since the documentation incorrectly describes the argument in this way:
https://github.com/jonescompneurolab/hnn-core/blob/e27afa9269e4a6fe221e23517b689da7f54f26d8/hnn_core/batch_simulate.py#L26

This will require changing the internals of `BatchSimulate`, as well as updating the tests and associated [code tutorial](https://jonescompneurolab.github.io/hnn-core/stable/auto_examples/howto/plot_batch_simulate.html#sphx-glr-auto-examples-howto-plot-batch-simulate-py)

PR #1208, opened 2025-12-21
[MRG] Align BatchSimulate set_params signature with Optimizer fixes issue #1203

Rahul-2k4:
### **Summary**

This PR addresses issue #1203 by standardizing the set_params callback signature used by BatchSimulate. The callback is updated to accept arguments in the order (net, param_values), aligning it with the Optimizer API and the existing documentation.

### **⚠️ Breaking Change**

- This is a breaking public API change for users who provide a custom set_params callback.
- Old signature: set_params(param_values, net)
- New signature: set_params(net, param_values)

### **Changes Included**

- Updated the internal call to set_params in BatchSimulate (_run_single_sim)
- Updated test fixtures to reflect the new callback signature
- Updated the batch simulation example to prevent user confusion
- Added a note to doc/whats_new.md under in-progress development changes

### **Rationale**

This change removes an inconsistency between BatchSimulate and Optimizer and ensures that the implementation matches the documented API. A clean break was chosen, as proposed in the issue, to avoid long-term technical debt from deprecation shims.

Please let me know if you’d prefer a different approach or any additional adjustments.


.gitignore
```diff
@@ -39,3 +39,5 @@ cover
 __pycache__
 
 *.log
+
+GEMINI.md
```
ntolley: @Rahul-2k4 it's ok to use LLM's for code assistance but you certainly need to read through all changes before submitting PRs!

This isn't really relevant to the PR scope and should be removed

asoplata:
Unfortunately yes, I believe this warrants a deprecation warning, just to be safe. However I think it can be short, like publish the warning on our next `0.5.x` and then apply the deprecation at `0.6.0`.

Also, the failing test is *not* due to this PR specifically - this same test is currently failing on `master` itself. Investigating.

Rahul-2k4:
Hi @ntolley,

Thanks for the review! I've addressed both comments:

**Changes made:**
- ✅ Added PR reference and author attribution to the changelog entry in `doc/whats_new.md`
- ✅ Removed the unrelated `GEMINI.md` entry from `.gitignore` 

**Summary of the feature implementation:**

This PR aligns the `BatchSimulate.set_params` callback signature with the `Optimizer` class by swapping the argument order from `(param_values, net)` to `(net, param_values)`.

**Files updated:**
- `hnn_core/batch_simulate.py`: Updated the call to `self.set_params(net, param_values)`
- `hnn_core/tests/test_batch_simulate.py`: Updated test fixture signature
- `examples/howto/plot_batch_simulate.py`: Updated example function signature and docstring
- `doc/whats_new.md`: Added API change note with PR reference

This ensures consistency across the API and matches the documented signature. Could you please review when you have a chance?

Thanks!

asoplata:
Force-merging since the only breaking check is CircleCI, which is also only breaking because of an unrelated issue with using upstream MNE sample data in our example scripts.

PR #1212, opened 2026-01-09
[WIP] PR for recording transmembrane currents and reproducing the axial dipole

dylansdaniels:
*This is considered "Version 1" of the TM current code

### DIFF
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -6,12 +6,11 @@
 from copy import deepcopy
 
 import numpy as np
+from neuron import h, nrn
 from numpy.linalg import norm
 
-from neuron import h, nrn
-
+from .externals.mne import _check_option, _validate_type
 from .viz import plot_cell_morphology
-from .externals.mne import _validate_type, _check_option
 
 # Units for e: mV
 # Units for gbar: S/cm^2
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -394,6 +394,21 @@ class Cell:
         self.vsec = dict()
         self.isec = dict()
         self.ca = dict()
+        # [new]
+        # initialize dicts to store transmenbrane (tm) current recordings
+        self.agg_i_mem = dict()  # aggregate tm currents
+        self.agg_ina = dict()    # aggregate tm sodium
+        self.agg_ik = dict()     # aggregate tm potassium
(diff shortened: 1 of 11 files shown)


ntolley:
Thanks @dylansdaniels this is a lot of work! we might need a dedicated code review meeting to get started on this, Definitely excited to see this used in some of the analyses we've been brainstorming (i.e. cell type specific contributions to LFP)

dylansdaniels:
The biggest challenge in validating the implementation of the dipole reconstruction from the individual transmembrane currents (as opposed to from i_mem, the total transmembrane current) was that we would need synaptic currents at the segment level in order to validate that everything works as expected. But because we model model the synaptic currents as being at the midpoint for the section, this isn't readily available.

To get around this, I did the reconstruction for *just* the soma, which has 1 segment, and was able to show that we can recreate i_mem closely enough that the difference is almost certainly noise (and by that I mean, accumulated rounding errors due to adding over multiple arrays, or possibly small differences in how/when in the integration the different channels are recorded by NEURON)

Here is the reconstruction / comparison:
NRMSE is normalized RMSE :)
image

@ntolley

ntolley:
Just finished my first review @dylansdaniels ! Really great work, the implementation is super clean and easy to understand

I want to get your opinion on an alternative organization for the recording functionality. The naming convention of `isec` and `vsec` are a bit opaque and not a good fit with this addition. I would propose the following:

Rename `vsec -> vmem` and `isec -> isyn`, as the voltages are really the "membrane potential", and the currents in `isec` are exclusively synaptic currents.

I also want you to consider lumping all of the membrane currents into a single attribute called `imem`, having the channels be accessible by key names in the dictionary (just like the current `isec`). My main concern is that dramatically expanding the arguments to the `simulate_dipole` API both 1) adds a ton of code that needs to be tested, and 2) makes our lives much more complicated if we want to simplify this feature and deprecate 10+ new arguments. I feel like turning on/off individual currents is a good feature, but there is likely a cleaner way to do it by passing special strings/dictionaries via `simulate_dipole(record_imem=<>)`

asoplata:
Closing this, since:

1. It is not planned to be merged, since @dylansdaniels is working on a Version 2 of the same feature
2. @annacatt already has work on her own branch built on top of this for the LFP/CSD analysis work

PR #1213, opened 2026-01-09
[MRG] Auto rename duplicate simulations

### LINKED ISSUE #1137: ENH/GUI: Auto-append new simulation name in GUI
In the GUI, when the user tries to run a second simulation, but they have not explicitly changed the name of the simulation, we should change the behavior. Currently, there is a hard stop and the user is forced to change the name. It would be more user-friendly if, instead, in this situation we simply changed the name automatically and appended some text to it, such as `default` becoming `default-002`, etc. This way, the user can always explicitly change the name if they want to, but they don't have to, especially if they're iterating between values in one of the parameter tabs.

Tusharjamdade:
Fixes #1137 
Improves the GUI simulation workflow by removing the hard stop when a user reruns a simulation without changing its name. Instead of forcing a manual rename, the GUI now automatically generates a unique simulation name by appending an incremented numeric suffix.
```
default → default-2 → default-3 → default-4 → …
```

### DIFF
--- a/doc/whats_new.md
+++ b/doc/whats_new.md
@@ -35,6 +35,13 @@ merged into `master`! Use `git log` instead and cross-reference instead. -->
 
 ### Changelog
 
+- In the GUI, add auto-renaming of new simulations. This fixes an annoying issue with
+  the GUI where every time a user wanted to run a new simulation, they had to go back to
+  the Simulation tab and manually change the simulation name. Now, whenever they run a
+  new simulation without changing the name manually, the new simulation will use the old
+  name, but with a reasonable number appended.
+  By [Tushar Jamdade][] in {gh}`1213`.
+
 - Add support for printing a welcome message that asks the user to fill out the HNN
   survey, which is printed to the standard output after (and only after) the first time
   the `hnn_core` module is imported. Whether or not the message is displayed is


hnn_core/gui/gui.py
```diff
@@ -2686,9 +2686,13 @@ def run_button_clicked(
...
         if simulation_data[_sim_name]["net"] is not None:
-            print("Simulation with the same name exists!")
-            simulation_status_bar.value = simulation_status_contents["failed"]
-            return
+            base = _sim_name.split("-")[0]
+            idx = 2
+            while f"{base}-{idx:03d}" in simulation_data:
+                idx += 1
+
+            _sim_name = f"{base}-{idx:03d}"
```
ntolley:
```suggestion
            _sim_name = f"{base}-{idx}"
```
I think it's probably fine to not use fixed with number formatting

Tusharjamdade:
Hi @ntolley, thank you for the review! I’ve made the suggested changes, please take a look when you get a chance.

Tusharjamdade:
Hi maintainers, this PR is for GSoC 2026. Could you please add the `gsoc-2026` label? I would also appreciate a review whenever you have time. Thank you!

Tusharjamdade:
Thank you for the detailed review! I've updated the implementation to use `rsplit("-", 1)` so that hyphenated names are preserved and existing numeric suffixes are incremented correctly (e.g., sim-34 → sim-35 instead of sim-34-2).

The logic now handles all the edge cases you pointed out, and the new tests are passing. Please let me know if you'd like any further refinements.

PR #1219, opened 2026-01-18
[WIP] fixed the parameters of set_params

Anniieee13:
Fixed the parameters of set_params function from (net, params) to (params, net) everywhere.

### DIFF
--- a/hnn_core/tests/test_general_optimization.py
+++ b/hnn_core/tests/test_general_optimization.py
@@ -47,7 +47,7 @@ def test_optimize_evoked(solver):
     # define set_params function and constraints
     net_offset = jones_2009_model(mesh_shape=(3, 3))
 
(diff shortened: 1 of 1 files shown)


asoplata:
Hello Annie, are you still interested in pursing this PR? If not, or if we don't hear back in a month or two, we may close this PR as stale. Thank you for your time.

Anniieee13:
Yes hello, sorry for the late reply. I'm interested!

asoplata:
Here's the next steps from the prior comment:

- Please update ALL uses of `BatchSimulate.set_params` to use the new argument ordering, including other test code and internal code. Please add a test for `set_params` that verifies that at least `net` is the correct type.

asoplata:
Hey Annie, we're going to go with completing this issue in #1208 instead of your PR. Thank you for your time.

Issue #1196, opened 2025-12-11
ENH: Add CMA-ES algorithm as an optimization solver


ntolley:
I think one of the biggest shortcomings of our current optimization algorithms is that they are entirely sequential (i.e. 1 simulation per epoch). Alternative algorithms draw several samples tin an epoch to build a more accurate update of where the algorithm "believes"

Personally I am quite fond of CMA-ES (covariance matrix adaption evolution strategy). This approach shares mathematical similarities with Bayesian optimization, and in my own testing on toy optimization problems seems to be very simulation efficient.

https://en.wikipedia.org/wiki/CMA-ES

The math is not too difficult to translate into Python, but it will definitely be a project to get a proof of concept working. There are certainly existing packages that implement this, but the challenge will be translating this to work with our current Optimization backend code.

asoplata:
Resolved via #1221

PR #1221, opened 2026-01-25
[MRG] Optimization enhancements: CMA-ES algorithm and correlation loss function

ntolley:
This PR makes a lot of changes to implement new optimization functions, namely:
- Covariance matrix adaption evolutionary strategy (#1196)
- Correlation coefficient as an objective function (already started here: #1206)

Rather than trying to merge this PR in one go, I think it should be broken into smaller tasks as this required a decent amount of changes all over the code base. However there is value to have a fully functioning branch to refer back to and play around with.

### DIFF
--- a/hnn_core/tests/test_general_optimization.py
+++ b/hnn_core/tests/test_general_optimization.py
@@ -7,7 +7,6 @@ import pytest
 
 from hnn_core import jones_2009_model, simulate_dipole
 from hnn_core.optimization import Optimizer
-from hnn_core.optimization.objective_functions import _anticorr_evoked
 import numpy as np
 
 


hnn_core/optimization/general_optimization.py
```diff
@@ -77,13 +78,13 @@ def __init__(
...
-                + "drives, provide a Network object with no "
-                + "external drives."
-            )
+        # Nick: I don't think this should be a hard requirement
+        # if initial_net.external_drives:
+        #     raise ValueError(
+        #         "The current Network instance has external "
+        #         + "drives, provide a Network object with no "
+        #         + "external drives."
+        #     )
```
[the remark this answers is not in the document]
katduecker: Just digging into this because I am having an issue with the final value in optim.obj_ not matching with the correlation/rmse/rmse-correlation (tried all of them) between experimental and simulation dipole I get when running simulate_dipole with optim.net_

One thing I noticed with the method you use here is that the weights in optim.net_.external_drives are all zero and they are also exported that way when using write_network_configuration

I think this is super confusing when sharing the network, or checking if the solutions the CMA-ES came up with are reasonable. It's not straightforward to check the final solution because it's not stored in the place that we instruct users to find the information on the drives... I am digging into whether this is the underlying cause of the issue (it probably isn't), but wanted to document while the thought is still fresh.

hnn_core/optimization/general_optimization.py
```diff
@@ -409,6 +438,92 @@ def _obj_func(predicted_params):
...
+        solutions = es.ask()
+        es.tell(solutions, _obj_func(solutions))
+        es.disp()
+    es.result_pretty()
+
+    # get optimized params
+    opt_params = solutions
+
+    # get objective values
+    obj = [np.min(obj_values[:idx]) for idx in range(1, max_iter + 1)]
```
katduecker: This is tracking the best solution over all iterations, but you take the best params of the last iterations. See Carolina's PR #1240 

I suggest 

```
# get best params
best_params = es.result.xbest

# get objective values
obj = [np.min(obj_values[:idx]) for idx in range(1, max_iter + 1)]

# get optimized net
params = _update_params(initial_params, best_params)
```

hnn_core/optimization/general_optimization.py
```diff
@@ -409,6 +438,92 @@ def _obj_func(predicted_params):
...
+            initial_params=initial_params,
+            set_params=set_params,
+            predicted_params=predicted_params,
+            update_params=_update_params,
+            obj_values=obj_values,
+            tstop=tstop,
+            obj_fun_kwargs=obj_fun_kwargs,
+        )
+
+    _b_obj_func = cma.BoundDomainTransform(_obj_func, constraints)  # evaluates fun only in the bounded domain
```
katduecker: I have been getting a lot (most) solutions outside the bounds. I am not sure what this does, but the solution seems to be to add 
'bounds': constraints,
to cma.CMAEvolutionStrategy

hnn_core/parallel_backends.py
```diff
@@ -560,10 +560,11 @@ def simulate(self, net, tstop, dt, n_trials, postproc=False):
             The Dipole results from each simulation trial
         """
 
-        print(
-            f"Joblib will run {n_trials} trial(s) in parallel by "
-            f"distributing trials over {self.n_jobs} jobs."
-        )
+        if net._verbose:
```
ntolley: A batch simulate test is failing that I'm at a loss to explain, it indicates that simulating serially is faster in parallel

The only real changes to batch simulate or parallel backends have to do with the verbose setting

Is it possible that these if blocks are slowing down parallel execution in a non-trivial way? But it doesn't make sense to me why serial would be faster...

hnn_core/optimization/objective_functions.py
```diff
@@ -5,8 +5,61 @@
...
+    elif len(np.array(predicted_params).shape) == 2:
+        is_batch = True
+    else:
+        raise ValueError(
+            f"Incorrect shape for predicted params. Got {np.array(predicted_params).shape}"
+        )
+    return is_batch
+
+
+def _preprocess_dipole(dpls, obj_fun_kwargs):
```
asoplata: Future refactor idea: promote this to a public `Dipole` method

hnn_core/optimization/objective_functions.py
```diff
@@ -39,31 +92,79 @@ def _rmse_evoked(
...
-    # smooth & scale
-    if "scale_factor" in obj_fun_kwargs:
-        [dpl.scale(obj_fun_kwargs["scale_factor"]) for dpl in dpls]
-    if "smooth_window_len" in obj_fun_kwargs:
-        [dpl.smooth(obj_fun_kwargs["smooth_window_len"]) for dpl in dpls]
+        batch_simulation = BatchSimulate(
+            net=new_net,
+            set_params=set_params_batch,
+            save_outputs=False,
+            save_dpl=True,
```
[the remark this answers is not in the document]
ntolley: I'm not actually sure I understand this interpretation of `save_dpl`, the intended function is `save_outputs` determines if any simulation data is written to a file, and the other `save_xyz` determines what is included/excluded

Let's discuss the batch analysis functionality separately, at the moment all analyses are rather cheap to execute so I haven't seen people stumble upon a need for parallelization. Truthfully we don't really have much analysis intrinsic to HNN other than some alternative visualizations.

hnn_core/optimization/general_optimization.py
```diff
@@ -76,14 +77,12 @@ def __init__(
...
         opt_params_ : list
             The list of optimized parameter values.
-        """
 
-        if initial_net.external_drives:
-            raise ValueError(
-                "The current Network instance has external "
-                + "drives, provide a Network object with no "
-                + "external drives."
-            )
```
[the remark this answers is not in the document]
katduecker: It was convenient to implement the set_params the way you did but it does require really good knowledge of the code base to understand what's going on. We'd have to really educate users on it. I worry that users run into issues or the optimization doesn't run at all because someone may accidentally not define any drives (not knowing). In the tutorials, I'd show both methods and be very explicit about how each of them works.

katduecker:
I was having a quick look but don't have a ton of time unfortunately! Did you remove the set_params_opt_drives function because I had a comment on that!

ntolley:
> I was having a quick look but don't have a ton of time unfortunately! Did you remove the set_params_opt_drives function because I had a comment on that!

Apologies yes it was removed! @asoplata and I decided that it didn't really make sense to keep those functions in the source code, but would like to add them back later in an example on the textbook (or perhaps some "utility" functions file?)

PR #1226, opened 2026-01-27
[MRG] maint: add AI config files to gitignore

asoplata:
Also moves around some pre-existing ignored files. Also adds `uv.lock` for upcoming uv usage.


dylansdaniels:
discussed during code review meeting 2/5/2026

PR #1228, opened 2026-01-28
[MRG] Added first verbose checks

annacatt:
Add a verbose argument to simulate_dipole, with True as the default. This allows users to silence printed output by explicitly setting verbose=False. In the current implementation, warnings are still always printed. It would be good to discuss whether that behavior is desirable.

### DIFF
--- a/hnn_core/tests/test_network.py
+++ b/hnn_core/tests/test_network.py
@@ -2074,6 +2074,7 @@ def test_check_global_synaptic_gains_uniformity():
     assert "WARNING" in stdout
     assert "custom synaptic gain values" in stdout
 
+
 def test_verbose():
     """Test that verbose flag controls print statements."""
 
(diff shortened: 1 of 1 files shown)


hnn_core/network.py
```diff
@@ -457,6 +458,7 @@ def __init__(
         # artificial drive cells
         self.gid_ranges = OrderedDict()
         self._n_gids = 0  # utility: keep track of last GID
+        self._verbose = bool(verbose)
```
```suggestion
        self._verbose = True
```

I actually think that this shouldn't be an argument to the network instantiation.

Instead we just use the underscore indicates that this is a "private" attribute, as in users shouldn't freely modify it

hnn_core/network_models.py
```diff
@@ -37,7 +38,12 @@ def jones_2009_model(
     mesh_shape : tuple of int (default: (10, 10))
         Defines the (n_x, n_y) shape of the grid of pyramidal cells.
 
+    verbose : bool
+        If True, set the hidden attribute ``net._verbose`` to True so
+        downstream builders print verbose output. Default: False
+
```
ntolley:
```suggestion
```

hnn_core/tests/test_network.py
```diff
@@ -2073,3 +2073,19 @@ def test_check_global_synaptic_gains_uniformity():
...
+    with io.StringIO() as buf, redirect_stdout(buf):
+        simulate_dipole(net, dt=0.5, tstop=20.0, verbose=True)
+        stdout_verbose_true = buf.getvalue()
+
+    with io.StringIO() as buf, redirect_stdout(buf):
+        simulate_dipole(net, dt=0.5, tstop=20.0, verbose=False)
+        stdout_verbose_false = buf.getvalue()
+
+    assert stdout_verbose_true != ""
+    assert stdout_verbose_false == ""
```
ntolley:
```suggestion
```

PR #1227, opened 2026-01-27
[MRG] Introduce Duecker ET model (revised history)

asoplata:
This is the new PR that will merge in @katduecker 's new model, and this branch (`katduecker/duecker_ET_model_rev_history`) is where all future development on the model should go.

This supercedes #1168, which will be closed to reflect this. The branch for #1168 (`katduecker/duecker_ET_model`) is no longer expected to be developed, but will still be kept around as an historical "backup". This current branch includes a huge, conflict-resolving commit in 6a85cfc that is a "squashed" commit consistent of over 100 commits from the `katduecker/duecker_ET_model` branch. 

This new branch was created in order for the model code to have a cleaner ability to diff against and merge into `master` instead of overwriting the individual commit history (which can occasionally be useful).

PR #1229, opened 2026-01-29
[MRG] feat: Add support for new, custom synapse types

asoplata:
This change is in anticipation of @katduecker's new model in #1227. This makes significant changes to
`cells_default.py::_get_syn_props` (renamed and refactored from its former name `cells_default.py::_get_pyr_syn_props`), `cells_default.py::_get_basket_syn_props`, and especially `Cell.syn_create`. These changes include a few things:

1. The "synaptic property" dictionaries (e.g. the `syn_props` as returned by `_get_baskey_syn_props`) can, but are not required to, have a new key-value pair where the key is "type" and the value is the name of the synapse type to use.

2. Later, when the `Cell` is constructing the actual synapses in `Cell.syn_create`, if there is no synapse "type" provided via the synaptic properties, or if the "type" is `Exp2Syn`, then everything proceeds as it did before. An `Exp2Syn` synapse is created at the provided segment, and its `e`, `tau1`, and `tau2` parameters are set for that synapse. Instead of these synaptic parameters being hardcoded into the arguments for `syn_create`, they are now included as part of `kwargs`.

3. If the "type" key exists and its value is *not* `Exp2Syn`, then things are different (this is the main point of this branch). For example, on the branch for #1227 , on the following line, there is a new synapse "type" used called `NMDA_gao`:
https://github.com/katduecker/hnn-core/blob/d736459b6df2f005784d644b316528cb935fa7f1/hnn_core/params_default.py#L358 This is a new synapse type that corresponds to the new `NMDA_gao.mod` file in that branch. In the current branch (here), in `Cell.syn_create`, this new, non-`Exp2Syn` synapse type will instead lead to the following:

    3.1. A new synapse of this class will be created (see this line on
    the current branch):

    ```
    syn = synapse_class(secloc)
    ```

    3.2. Then, the remaining `kwargs` arguments will be assumed to be
    key-value pairs that are expected to correspond to parameters that
    can be set in this particular synapse type. These will be used to
    set the attribute values of this instantiation of this synapse
    type.

All together, this is slightly different from Katharina's original implementation of the custom synapses used in here https://github.com/katduecker/hnn-core/blob/duecker_ET_model_rev_history/hnn_core/cell.py#L876-L877 but this approach has the advantage of supporting runtime setting of custom synapse type parameters that are unique to that synapse type (e.g. in `params_default.py`) which are different from the traditional `{e, tau1, tau2}` set. (@katduecker, in other words, this will allow you to provide custom `gabab.mod` or `NMDA_gao.mod` synaptic parameters in your `params_default.py` file, if you so choose).

A few random notes:
- If you look at `get_L2Pyr_params_default`, you will see that I did *not* universally add the new "..._type" key to all synapses, but only some. This was done to illustrate that this is backwards-compatible with synaptic-property-inputting data structures that do not have the "..._type" key.
- I did have to regenerate the default network JSON files, but this is not because this change is backwards-compatible (it is). This was necessary because one of our tests tests full in-memory equivalency between a newly generated `jones2009` model object versus a read-from-file version of the same object.

Final note: I have not yet tested this against the custom `gabab.mod` or `NMDA_gao.mod` synapse types from #1227 yet. I'm not sure, but perhaps the best way to test this implementation against the custom MODs in 1227 is to do testing on a separate PR into #1227.

### DIFF
--- a/hnn_core/cell.py
+++ b/hnn_core/cell.py
@@ -315,10 +315,8 @@ class Cell:
     sections : dict of Section
         Dictionary with keys as section name.
     synapses : dict of dict
-        Keys are name of synaptic mechanism. Each synaptic mechanism
-        has keys for parameters of the mechanism, e.g., 'e', 'tau1',
-        'tau2'.
-        sections.
+        Keys are name of synaptic mechanism. Each synaptic mechanism dict has keys for
+        parameters of the mechanism, e.g., 'e', 'tau1', 'tau2'.  sections.
     sect_loc : dict of list
         Can have keys 'proximal' or 'distal' each containing
         names of section locations that are proximal or distal.
@@ -843,33 +841,93 @@ class Cell:
                     self.ca[sec_name] = h.Vector()
                     self.ca[sec_name].record(self._nrn_sections[sec_name](0.5)._ref_cai)
 
-    def syn_create(self, secloc, e, tau1, tau2):
-        """Create an h.Exp2Syn synapse.
+    def syn_create(self, secloc, **kwargs):
+        """Create a NEURON synapse at the given section location.
+
+        By default (or when ``mechname='Exp2Syn'``), creates an ``h.Exp2Syn`` two-state
+        kinetic synapse. A custom NEURON synapse mechanism can be used instead by
+        passing ``mechname=<synapse_class_name>`` along with any parameters that
+        synapse accepts.
 
         Parameters
         ----------
         secloc : instance of nrn.Segment
-            The section location. E.g., soma(0.5).
-        e: float
-            Reverse potential (in mV)
-        tau1: float
-            Rise time (in ms)
-        tau2: float
-            Decay time (in ms)
+            The section location. E.g., ``soma(0.5)``.
+        **kwargs : dict
+            Synapse parameters. The following keys are recognised:
+
+            mechname : str, optional
+                Name of the NEURON synapse class to instantiate (must be available in
+                NEURON's hoc interpreter). Custom synapse mechanisms must be compiled
+                from a MOD file. Defaults to ``'Exp2Syn'``.
+            e : float
+                Reversal potential (mV). Required for ``Exp2Syn``.
+            tau1 : float
+                Rise time constant (ms). Required for ``Exp2Syn``.
+            tau2 : float
+                Decay time constant (ms). Required for ``Exp2Syn``.
+
+            For custom synapse mechanisms, pass any parameters accepted by that synapse
+            as additional keyword arguments. Unknown or invalid parameter names will
+            raise a ``ValueError``.
 
         Returns
         -------
-        syn : instance of h.Exp2Syn
-            A two state kinetic scheme synapse.
+        syn : NEURON synapse object
+            The instantiated synapse object (e.g. ``h.Exp2Syn`` or a custom mechanism).
         """
         if not isinstance(secloc, nrn.Segment):
             raise TypeError(
-                f"secloc must be instance ofnrn.Segment. Got {type(secloc)}"
+                f"secloc must be instance of nrn.Segment. Got {type(secloc)}"
             )
-        syn = h.Exp2Syn(secloc)
-        syn.e = e
-        syn.tau1 = tau1
-        syn.tau2 = tau2
+
+        # Check for no "mechname" (including backwards compatibility), or case of mechname being
+        # explicitly "Exp2Syn"
+        if ("mechname" not in kwargs.keys()) or (kwargs["mechname"] == "Exp2Syn"):
+            syn = h.Exp2Syn(secloc)
+            try:
+                for param_name in ["e", "tau1", "tau2"]:
+                    setattr(syn, param_name, kwargs[param_name])
+            except KeyError:
+                raise KeyError(
+                    f"The Exp2Syn synapse in segment {secloc} is missing a required parameter."
+                )
+        else:
+            # Otherwise, there is a custom synapse mechanism to be used. Create a
+            # synapse of that mechanism, then use all remaining kwargs to set its
+            # attributes.
+            # -------------------------------------------------------------------------
+            try:
+                # This tries to obtain the synapse object as if it is an attribute of
+                # the HOC interpreter as a whole, which all valid synapse mechanisms
+                # should be. In other words, if you have compiled a synapse mechanism
+                # called e.g. 'NMDA_gao', then `h.NMDA_gao` should exist.
+                synapse_class = getattr(h, kwargs["mechname"])
+            except AttributeError:
+                raise ValueError(
+                    f"Synapse mechanism '{kwargs['mechname']}' not found in NEURON's "
+                    "hoc interpreter. You probably either need to recompile your MOD "
+                    "mechanism files or you are missing a required MOD file for this "
+                    "synapse mechanism."
+                )
+            # Create the synapse
+            syn = synapse_class(secloc)
+
+            # Set attributes of this synapse from their entry in kwargs, if present
+            for param_name, param_value in kwargs.items():
+                if hasattr(syn, param_name):
+                    setattr(syn, param_name, param_value)
+                elif param_name == "mechname":
+                    # Ignore "mechname" since it's only used in the control flow of
+                    # this function, but is never an actual attribute of the NEURON
+                    # Synapse object.
+                    continue
+                else:
+                    raise ValueError(
+                        f"Synapse mechanism '{kwargs['mechname']}' does not have a parameter "
+                        f"named '{param_name}'."
+                    )
+
         return syn
 
     def setup_source_netcon(self, threshold):
--- a/hnn_core/cells_default.py
+++ b/hnn_core/cells_default.py
@@ -284,7 +284,7 @@ def _cell_L2Pyr(override_params, pos=(0.0, 0.0, 0), gid=0):
         "distal": ["apical_tuft"],
     }
 
-    synapses = _get_pyr_syn_props(p_all, "L2Pyr")
+    synapses = _get_syn_props(p_all, "L2Pyr")
     return Cell(
         "L2Pyr",
         pos,
@@ -450,7 +450,7 @@ def _cell_L5Pyr(override_params, pos=(0.0, 0.0, 0), gid=0):
         "distal": ["apical_tuft"],
     }
 
-    synapses = _get_pyr_syn_props(p_all, "L5Pyr")
+    synapses = _get_syn_props(p_all, "L5Pyr")
     return Cell(
         "L5Pyr",
         pos,
@@ -507,11 +507,18 @@ def _get_basket_soma(v_init=-64.9737):
     )
 
 
-def _get_pyr_syn_props(p_all, cell_type):
-    """Return the default synaptic parameters for a given Pyramidal cell_type.
+def _get_syn_props(p_all, cell_type, syn_types=["ampa", "nmda", "gabaa", "gabab"]):
+    """Get synaptic properties for a specific cell type.
 
-    Extracts the 'e', 'tau1', and 'tau2' parameters for all synapse types for a
-    particular cell_type.
+    When parameters in `p_all` are provided for a "syn_type" (of the set {ampa, nmda,
+    gabaa, gabab}), and no "syn_type_mechname" key is present in p_all, then this will
+    *always* return the default synapse mechanism of "Exp2Syn". In order to use a custom
+    "mechanism" (such as "NMDA_gao" or "gabab_destexhe"), the incoming parameters in
+    `p_all` should include a key of the form `f"{cell_type}_{syn_type}_mechname"` with
+    the value being the name of the custom mechanism.
+
+    This was formerly called `_get_pyr_syn_props`, but `_pyr` was dropped from name due
+    to this function's future usage by interneurons in the upcoming Duecker model.
 
     Parameters
     ----------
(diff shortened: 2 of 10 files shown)


asoplata:
Small note I forgot to add: this implementation means that you should *ONLY* provide synaptic property key-values for *valid* parameters of that synapse type. In other words, here, for example

https://github.com/katduecker/hnn-core/blob/d736459b6df2f005784d644b316528cb935fa7f1/hnn_core/params_default.py#L355-L358

this implementation is incomaptible with the `tau1` and `tau2` that are defined above, but which are not actually valid parameters of the `NMDA_gao` MOD file. Since `e` is a valid parameter, that one is okay. In other words, `L2Pyr_nmda_tau1` and `L2Pyr_nmda_tau2` would need to be removed from that file to work with this implementation. However, you would be able to provide values for parameters that are custom to `NMDA_gao` in the same file, such as  "L2Pyr_nmda_Alpha" or whatever.

asoplata:
@katduecker Fortunately, we're getting the point where it would be useful for me to have either some kind of script and/or data (e.g. dipole output from a successful and reproducible run of your model), so that I can begin testing code changes to `master` against your model specifically. Do you mind sending me a script that creates your `Network` and runs a simulation, *exactly like you want it to*? This way I can generate data, try to make some code changes, then check with a second simulation to ensure that the output is identical and bit-perfect before and after. Feel free to email me such a script instead of putting on Github (but however you do it is fine)

asoplata:
This PR is now ready for review and (hopefully!!!) merge @katduecker @ntolley @dylansdaniels 

Key Changes:

1. Important: New `NMDA_gao` MOD file and synapse type (`hnn_core/mod/NMDA_gao.mod`)

    This adds the single new synaptic mechanism file needed for the upcoming Duecker model, `NMDA_gao.mod`, directly from #1227. This PR does not create any Networks that use this mechanism, but it does run new tests that test if the Synapse is properly created in `Cell` objects (mostly in `test_cell.py`). This includes tests of when the user (specifically, Katharina) uses `params_default.py`-like dictionaries to include and customize the parameters of the new Synapse type.

2. Synapse `type` parameter:

    All synapses can now, optionally, use a new `type` parameter that can be specified identically to their existing `Exp2Syn` variable parameters. For example, just as you can specify `L2Pyr_nmda_tau1` in `params_default.py`, you can now use a new key `L2Pyr_nmda_type` whose value can be any valid NEURON Synapse object (including `Exp2Syn`, `ExpSyn` which we don't use currently, and the new `NMDA_gao`), including new types that you compile locally from MOD files. This is backwards-compatible, and you can specify custom types at the level of `params_default.py` (and probably elsewhere in the "code chain"). However, with this PR, the JSON networks need to be regenerated because `type: Exp2Syn` is now present in all the default synapse configurations.

3. `Cell.syn_create()` has been refactored:

    This is where the Synapse objects are actually created in NEURON. This now takes dynamic `**kwargs` instead of fixed arguments for `e, tau1, tau2`, since this now supports Synapse types other than `Exp2Syn`. In your `params_default.py` etc., for custom Synapse types, you can provide not only the `type` but also parameters that the NEURON interpreter allows you to set for that Synapse object after it's been created (e.g. `Beta` parameter in `NMDA_gao.mod` using `L2Pyr_nmda_Beta` in `params_default.py`). There is error handling (and tests of it) if you try to pass invalid parameters for that synapse type (e.g. if you try to pass `tau2` for `NMDA_gao.mod`, which isn't valid). There is also error handling (and tests of it) if you try to pass an invalid custom Synapse type. If there is no `type` parameter then that's fine, and `Exp2Syn` is used, so everything is backwards-compatible.

4. (small) In `cells_default.py`, `_get_pyr_syn_props` has been renamed to `_get_syn_props` anticipation of new interneuron use in the upcoming Duecker. It has been extended to read the new `type` parameter, and this new usage is tested.

5. All of the above changes have pretty comprehensive testing IMHO in `test_cell.py`, `test_cells_default.py`, and a new file `test_params_default.py`. The tests **include testing for the new `NMDA_gao` synapse**, so compilation of this new MOD file is required. (If you are testing this locally and you have installed from source, then you will need to run `make` or `nrnivmodl` etc. to compile the file.)

(Random notes: The initial tests were created with Claude. However, as others have found, I personally found Claude's test output to be a bit over-verbose and occasionally kind of redundant. A good way to make sure that you inspect all of your LLM's code output is to see if it benefits from some refactoring to make it cleaner. It especially likes to hard-code values at multiple locations instead of setting them once and using a variable, etc. Claude is great but now that I've spent some time with it, I find something to improve with almost every test it writes.)

hnn_core/mod/NMDA_gao.mod
```diff
@@ -0,0 +1,170 @@
...
+	(mV) = (millivolt)
+	(uS) = (microsiemens)
+	(mM) = (milli/liter)
+}
+
+PARAMETER {
+	Cmax	= 1	 (mM)           : max transmitter concentration
+	Cdur	= 1	 (ms)		: transmitter duration (rising phase)
+	Alpha	= 4 (/ms /mM)	: forward (binding) rate (4)
+	Beta 	=0.01   (/ms)   : backward (unbinding) rate
```
ntolley:
```suggestion
	Beta 	= 0.01   (/ms)   : backward (unbinding) rate
```
the spacing is all over the place in this file 😄 

thank god we have auto-formatting now

hnn_core/param/jones2009_base.json
```diff
@@ -46,17 +46,20 @@
                     "ampa": {
                         "e": 0,
                         "tau1": 0.5,
-                        "tau2": 5.0
+                        "tau2": 5.0,
+                        "type": "Exp2Syn"
```
ntolley:
```suggestion
                        "syn_type": "Exp2Syn"
```
Do we want to make this `syn_type` to avoid base python keyword issues? I think this also came up in #1227

hnn_core/tests/test_cell.py
```diff
@@ -42,7 +42,7 @@ def test_cell():
 
     # test that ExpSyn always takes nrn.Segment, not float
     with pytest.raises(TypeError, match="secloc must be instance of"):
-        cell.syn_create(0.5, e=0.0, tau1=0.5, tau2=5.0)
+        cell.syn_create(0.5, e=0.0, tau1=0.5, tau2=5.0, type="Exp2Syn")
```
ntolley, line 45:
```suggestion
        cell.syn_create(0.5, e=0.0, tau1=0.5, tau2=5.0, syn_type="Exp2Syn")
```
yeah this was changed to syn_type by @katduecker

hnn_core/tests/test_cell.py
```diff
@@ -182,3 +182,178 @@ def test_artificial_cell():
         match="gid must be an integer",
     ):
         cell = _ArtificialCell(event_times, threshold, gid="one")
+
+
+def _create_cell_and_run_synapse_test(input_synapse_config):
```
ntolley:
```suggestion
def _create_cell_and_run_synapse_test(input_synapse_config, syn_type='Exp2Syn'):
```
It seems awkward to have to pop a dictionary element below, can the syn type information just be passed as another argument?

hnn_core/tests/test_cell.py
```diff
@@ -182,3 +182,178 @@ def test_artificial_cell():
...
+    """
+    custom_synapse_config = dict(type="NonExistentSynapseXYZ", e=0.0)
+    with pytest.raises(ValueError, match="not found in NEURON's hoc interpreter"):
+        _create_cell_and_run_synapse_test(custom_synapse_config)
+
+
+def test_create_synapses_valid_custom_type():
+    """syn_create with valid custom 'type' kwarg should create the specified synapse.
+
+    NOTE: This usese `ExpSyn` as an example of a "custom" synapse type, and it looks
```
ntolley:
```suggestion
    NOTE: This uses `ExpSyn` as an example of a "custom" synapse type, and it looks
```

hnn_core/tests/test_cell.py
```diff
@@ -182,3 +182,178 @@ def test_artificial_cell():
...
+    type.
+
+    This uses `Cell.build` to test both `Cell._create_synapses` and `Cell.syn_create`.
+    """
+    custom_synapse_config = dict(tau1=1.0, tau2=20.0, type="ExpSyn", unknown_param=9)
+    with pytest.raises(ValueError, match="does not have a parameter named"):
+        _create_cell_and_run_synapse_test(custom_synapse_config)
+
+
+def test_create_synapses_valid_NMDA_gao_type_default():
```
ntolley:
```suggestion
def test_create_synapses_valid_NMDA_gao_type():
```
I know that we've been trying to make the tests more granular but this feels a bit extreme since copy-pasted docstrings make up most of the diff

I suppose since this is running a function inside these tests actually do a lot more than it first appears...

hnn_core/tests/test_cell.py
```diff
@@ -182,3 +182,178 @@ def test_artificial_cell():
...
+def test_create_synapses_missing_param_empty_type():
+    """syn_create with no 'type' kwarg but missing a required parameter should raise error.
+
+    This is because the default 'type' is 'Exp2Syn', which requires the parameters
+    'e', 'tau1', and 'tau2'.
+
+    This uses `Cell.build` to test both `Cell._create_synapses` and `Cell.syn_create`.
+    """
+    custom_synapse_config = dict(tau1=1.0, tau2=20.0)
+    with pytest.raises(KeyError, match="missing a required parameter"):
```
ntolley, line 256: Perhaps this is a design decision that should be documented?

Specifically, if any synapse parameter is customized they must _all_ be specified

This is in contrast to using the defaults specified in the `.mod` file when some parameters are missing

asoplata:
Okay, it looks like we need to make a lot of changes to this PR. We should meet in person if we're going to proceed to getting 1227 up to date

hnn_core/mod/NMDA_gao.mod
asoplata: I don't believe `ruff` autoformatting works for `mod` files

hnn_core/tests/test_cells_default.py
```diff
@@ -238,3 +245,136 @@ def test_linear_g_at_dist():
...
+    # Simulate a p_all dict without any _type keys (legacy / backwards-compat case)
+    cell_type = "L2Pyr"
+    syn_type = "ampa"
+    p_all = {
+        f"{cell_type}_{syn_type}_e": 0.0,
+        f"{cell_type}_{syn_type}_tau1": 0.5,
+        f"{cell_type}_{syn_type}_tau2": 5.0,
+    }
+    result1 = _get_syn_props(p_all, cell_type, syn_types=[syn_type])
+    for syn_type, syn_param in zip([syn_type], ["e", "tau1", "tau2"]):
```
ntolley:
```suggestion
    for syn_param in ["e", "tau1", "tau2"]:
```
This seems unnecessary

hnn_core/tests/test_cells_default.py
```diff
@@ -238,3 +245,136 @@ def test_linear_g_at_dist():
...
+
+    # Add the type key to test that "Exp2Syn" is correctly read when present
+    p_all.update(
+        {
+            f"{cell_type}_{syn_type}_type": "Exp2Syn",
+        }
+    )
+    result2 = _get_syn_props(p_all, cell_type, syn_types=[syn_type])
+    # Test that type key and value are correctly present
+    for syn_type, syn_param in zip([syn_type], ["e", "tau1", "tau2", "type"]):
```
ntolley:
```suggestion
    for syn_param in ["e", "tau1", "tau2", "type"]:
```

hnn_core/tests/test_cells_default.py
```diff
@@ -238,3 +245,136 @@ def test_linear_g_at_dist():
...
+        "L2Pyr_nmda_Beta": 5.0,
+        "L2Pyr_nmda_Cdur": 1.2,
+    }
+    result1 = _get_syn_props(p_all, "L2Pyr", syn_types=["ampa"])
+    assert "ampa" in result1
+    assert "gabaa" not in result1
+
+    result2 = _get_syn_props(p_all, "L2Pyr", syn_types=["ampa", "gabaa"])
+    assert set(result2.keys()) == {"ampa", "gabaa"}
+    assert result2["ampa"]["type"] == "Exp2Syn"
```
ntolley:
```suggestion
    assert result2["ampa"]["type"] == "Exp2Syn"
    assert result2["ampa"]["e"] == 0.0
```
I think it's important to check for values that differ between AMPA and GABAa

hnn_core/tests/test_params_default.py
```diff
@@ -0,0 +1,31 @@
...
+
+def test_get_syn_props_l2pyr_default_params_backwards_compat():
+    """L2Pyr default params have mixed type/no-type keys; all synapses should resolve."""
+    p_all = get_L2Pyr_params_default()
+    result = _get_syn_props(p_all, "L2Pyr")
+    # All four default synapse types should be present
+    for syn in ["ampa", "nmda", "gabaa", "gabab"]:
+        assert syn in result
+        # All should fall back to or explicitly declare Exp2Syn
+        assert result[syn]["type"] == "Exp2Syn"
```
ntolley: I don't really understand why `nmda` has type 'NMDA_gao' in some situations and 'Exp2Syn' in others

what is it exactly the triggers use of one or the other?

ntolley: The reason I am confused is because in the test above this line
```
result3 = _get_syn_props(p_all, "L2Pyr", syn_types=["ampa", "nmda"])
```
Actually produces a synapse with type 'NMDA_gao'

ntolley:
All done with my review @asoplata!

hnn_core/tests/test_cell.py
katduecker: We might even want to flag which one is missing, but I do like not silently reverting to the mod file values and instead requiring explicit definitions of the values!

hnn_core/tests/test_cell.py
asoplata: I've removed most of the redundant docstring content. 

Also, the primary reason why these tests are separate instead of a single test was for parallelization and speed. However, if what you're saying is that the tests are split up too much, then I agree after giving it a second overview. It's technically slower (not meaningfully since all this does is build a cell in NEURON) but if we combined all these tests into a single test, it would be far easier for a human to understand. I'll do that in a new commit

hnn_core/tests/test_params_default.py
katduecker: All the tests should use the Jones/Neymotin model right? So no NMDA_gao synapse!

hnn_core/mod/NMDA_gao.mod
katduecker: I think it's the classical NEURON vibes and should be kept :)

hnn_core/param/jones2009_base.json
katduecker: Agreed

hnn_core/cell.py
```diff
@@ -830,33 +828,86 @@ def record(self, record_vsec=False, record_isec=False, record_ca=False):
                     self.ca[sec_name] = h.Vector()
                     self.ca[sec_name].record(self._nrn_sections[sec_name](0.5)._ref_cai)
 
-    def syn_create(self, secloc, e, tau1, tau2):
-        """Create an h.Exp2Syn synapse.
+    def syn_create(self, secloc, **kwargs):
+        """Create a NEURON synapse at the given section location.
+
+        By default (or when ``type='Exp2Syn'``), creates an ``h.Exp2Syn`` two-state
```
katduecker: syn_type

hnn_core/cell.py
```diff
@@ -830,33 +828,86 @@ def record(self, record_vsec=False, record_isec=False, record_ca=False):
...
-            Reverse potential (in mV)
-        tau1: float
-            Rise time (in ms)
-        tau2: float
-            Decay time (in ms)
+            The section location. E.g., ``soma(0.5)``.
+        **kwargs : dict
+            Synapse parameters. The following keys are recognised:
+
+            type : str, optional
```
katduecker: syn_type

hnn_core/cell.py
```diff
@@ -830,33 +828,86 @@ def record(self, record_vsec=False, record_isec=False, record_ca=False):
...
+                f"secloc must be instance of nrn.Segment. Got {type(secloc)}"
             )
-        syn = h.Exp2Syn(secloc)
-        syn.e = e
-        syn.tau1 = tau1
-        syn.tau2 = tau2
+
+        # Check for no "type" (including backwards compatibility), or case of type being
+        # explicitly "Exp2Syn"
+        if ("type" not in kwargs.keys()) or (kwargs["type"] == "Exp2Syn"):
```
katduecker: syn_type

hnn_core/cells_default.py
```diff
@@ -235,36 +235,65 @@ def _get_basket_soma(cell_name):
...
+            }
+    return syn_props
 
 
 def _get_basket_syn_props():
     return {
-        "ampa": {"e": 0, "tau1": 0.5, "tau2": 5.0},
-        "gabaa": {"e": -80, "tau1": 0.5, "tau2": 5.0},
-        "nmda": {"e": 0, "tau1": 1.0, "tau2": 20.0},
+        "ampa": {"e": 0, "tau1": 0.5, "tau2": 5.0, "type": "Exp2Syn"},
```
katduecker: syn_type ... argh sorry this will be annoying

hnn_core/cell.py
```diff
@@ -830,33 +828,86 @@ def record(self, record_vsec=False, record_isec=False, record_ca=False):
...
+            except KeyError:
+                raise KeyError(
+                    f"The Exp2Syn synapse in segment {secloc} is missing a required parameter."
+                )
+        else:
+            # Otherwise, there is a custom synapse type to be used. Create a synapse of
+            # that type, then use all remaining kwargs to set its attributes.
+            # -------------------------------------------------------------------------
+            try:
+                synapse_class = getattr(h, kwargs["type"])
```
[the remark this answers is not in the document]
katduecker: Yes I don't understand

hnn_core/tests/test_params_default.py
ntolley: hmm something we may need to decide on in person, `syn_type` has two possible meanings:
1) The underlying `.mod` file or equation that defines the synapse ('Exp2Syn' or 'NMDA_gao' for example)

2) A specific set of parameters for a given `.mod` file ('ampa' and 'gabaa' for example both use 'Exp2Syn' with different rate constants)

hnn_core/tests/test_cell.py
asoplata: All the synapse tests have been combined into a single one which is MUCH easier to read. Marking this as resolved.

hnn_core/tests/test_cell.py
asoplata: I haven't looked at this since Feb, but IIRC the "all must be specified" constraint only applies to `Exp2Syn`, which is the same design it had before (you have to pass all params). I could be misremembering

hnn_core/tests/test_params_default.py
asoplata: This was resolved: In the other case of `test_get_syn_props_multiple_syn_types` that is outside this file, that test was using a DIFFERENT, hardcoded `p_all` that is not using the default parameters. In the case of this file, these tests are using the default parameters, and in this PR, I have not added use of `NMDA_gao` to ANY default model usage. `NMDA_gao` usage is currently nowhere in our default models, and ONLY used explicitly in the new tests.

hnn_core/tests/test_cells_default.py
asoplata: Agreed, applied, and resolved

hnn_core/tests/test_cells_default.py
asoplata: Agreed, applied, and resolved

hnn_core/tests/test_cell.py
asoplata: sounds like we need to discuss this

katduecker: Actually scratch! I never defined the NMDA measures I think we should go with the defaults for those more complicated synapses as they were fitted to data

hnn_core/param/jones2009_base.json
asoplata: I've finished changing all PR-related instances of `"type"` to use `"mechname"`. I left out the underscore due to how params parses its JSON files.

hnn_core/tests/test_cell.py
asoplata: Discussed this and concluded 2026-05-15 that this approach is fine: `Exp2Syn`(the default) always requires the user to pass `e, tau1, tau2`, while any custom mechanisms accept any amount of parameters (including none) , as long as the parameters are valid

PR #1230, opened 2026-01-30
[WIP] Maint: NEURON 9.0 transition

asoplata:
This are in-progress commits to test how we could enable usage of NEURON >9.0.

### 1. No MOD file changes needed

So far, the good news is that despite what I initially thought, we do *not* actually have to update any MOD files.

### 2. However, default simulation output changes slightly

The slightly bad news, however, is that even though all of our MOD/mechanism source code is the same, the default simulations produces very slightly different results when run using NEURON >9.0 (9.0.1) instead of <9.0 (8.2.7). These differences are on the order of single float precision such as `##.XXXX` within the timespan of the default sim. This also means that the recorded counts of spikes from the test data are also slightly different (`L5_basket` spikes 85 instead of 86 times).

I've made a plot below of the "original" dipole (top), the new NEURON 9.0 dipole (middle), and the difference (bottom):

neuron9-diff-dpl

I'm not sure exactly why there is a small difference. It could be many things due to NEURON 9.0 changes. For example, it could be the fact that MOD files are now compiled using C++ instead of C, or some other unknown internal calculation changes. Honestly, it could take a long time (or not, if lucky) to really investigate why the difference is present at all. 

In the meantime, if the "new" NEURON 9.0 dipole data matches the test-simulation on every platform here, then I propose that we augment our tests to handle both NEURON <9.0 cases and >9.0 cases separately. This could easily be done by pushing the "new" dipole data to its own test-data storage file (e.g. see `master`'s current https://github.com/jonescompneurolab/hnn-core/blob/ffd57c96926b8ffc72ddfbe1482f98e803ad44af/hnn_core/tests/test_parallel_backends.py#L376-L382 ), then adding a simple NEURON version check in the relevant tests that it impacts (maybe only that one test).

### 3. Only remaining change: compiled shared libraries.

Besides the above, it seems there's only one other change needed. This has to do with handling our compiled shared libraries at simulation start time, which you can see in the change in `network_builder.py`. For whatever reason, on our current `master` using NEURON 8.2.7, on MacOS, our initial mechanism compilation (done by `setup.py`) builds two different files that are completely identical except for their filename. These two files are built at installation time into:

- `hnn_core/mod/.libs/libnrnmech.so`
- `hnn_core/mod/libnrnmech.dylib`

If you `diff` these two files (on MacOS), then you'll see that they are binary-identical.

The thing that has changed for NEURON 9.0 in this case is that the `.lib/libnrnmech.so` shared object is no longer built, however `libnrnmech.dylib` still is. I've tested using either, and both files seem to work fine for allowing simulations to run and tests to succeed. The only part that needs to be changed is line 190 on `network_builder.py` such that our code to detect this shared library includes the possibility to detect `libnrnmech.dylib`. In the case of MacOS using NEURON 9.0, this change enables that file to be detected, which is required if there is no longer a `.lib/libnrnmech.so` being compiled anymore. This particular change should be a safe and backwards-compatible change.


asoplata:
@katduecker I've added some AI-slop scripts that we can use to probe where the exact differences are. There are two directories of new test scripts: `evo-drive` which is a basic sim with an evoked drive, and `no-drive` where the drive is commented out, but there's a tonic bias present. The way to use this is:

1. First, in a conda environment that is built off of `master` and uses the older version of NEURON (something like 8.27), run `make clean` (in the top-level directory!) to remove built mechanism files, then `make` (in the top-level directory!) to rebuild them, then `python script_dpl_old.py` of which directory you want to run. This will produce the files `dpl_old.txt` (old dipoles) and `vsec_old.npz` which is the soma vsecs of the first cell of each cell type.

2. Next, in a SEPARATE conda environment that has been constructed to use THIS BRANCH and which has NEURON 9.0.2, again run `make clean` (in the top-level directory!) to remove built mechanism files in C, then `make` (in the top-level directory!) to rebuild them using C++ for NRN9, , then run `python script_dpl_new.py`. This will produce new versions of the corresponding data files, but more importantly will output plots that show the dipoles and any differences.

When I do the above, I see that simulations appear to be different if there is a drive (scripts in `evo-drive/`) but identical if there is not a drive (scripts in `no-drive/`), even if there is a tonic bias. This does suggest that `vecevent.mod` may be to blame. HOWEVER, when I ask Claude to identify differences between the NRN8 generated .C file and the NRN9 generated .Cpp file for vecevent.mod, it's insistent that *" These two files would produce identical simulation output given the same inputs "*. 

So, the source of our difference could very well be `vecevent.mod`, but I'm not 100% convinced even with this testing. Fortunately the differences probably are limited to our drives, however. @katduecker I don't have time to dig into this for at least several days, do you want to continue where I left off?

One idea for what to do next would be to change our data saving such that our dipole data is saved at double floating point precision, i.e. much higher precision than our usual single floating point precision. Since in all my examples, any differences are at around the decimal cutoff for our filetype, it's possible that the actual differences are smaller than our filetype decimal cut off.

PR #1240, opened 2026-02-06
[MRG] COBYLA solver: Track best-performing parameters during optimization

carolinafernandezp:
The COBYLA algorithm implemented by SciPy returns the set of parameters that attempts to minimize the objective function **at termination**, instead of the argmin over all evaluated parameter sets. This means that, currently, `optim.net_` when `optim = Optimizer(net, solver='cobyla')` does not return the optimized network, but the network that was last explored.

The fix I'm implementing here is to track the best parameter set and corresponding objective value over all evaluated parameter sets. In other words, the proposed fix is to store and return the parameter set corresponding to the lowest observed objective value rather than the final iterate for the COBYLA solver.

We already do this with the Bayesian optimization function we implemented in externals, which is why we don't run into the same issue with the bayesian solver.


asoplata:
Hey @carolinafernandezp I've made a pull request to your branch on your fork here https://github.com/carolinafernandezp/hnn-core/pull/1 that should fix the failing test, see that for details

asoplata:
I have done a rebase onto `master` in order to include Nick's conflicting changes with the rest of the optimization code. Note that anyone else working on Carolina's branch here will need to delete their existing branch and re-download the branch from her fork. For help with this, see [here in our Contributing Guide](https://jonescompneurolab.github.io/hnn-core/stable/contributing.html#keeping-your-code-up-to-date-by-rebasing).

asoplata (approved):
I've fixed the remaining issues (passing tests, adding `best` support to `_anticorr_evoked`, and made a lot of documentation changes. I also fixed (?) a possible missing application of `obj_fun_kwargs` defaults to the non-CMA cases where relevant. Some of the doc changes have either `ATTN` or `@ntolley` in them so you can grep for them to find them.

I think this is good to go on my end; not only does it pass my test runs but my GUI opt branch passes tests as well with these changes applied.

hnn_core/optimization/general_optimization.py
```diff
@@ -157,49 +158,54 @@ def fit(self, **obj_fun_kwargs):
...
-        target : instance of Dipole (Required if obj_fun='dipole_rmse')
+        target : instance of Dipole (Required if obj_fun='dipole_rmse' or 'dipole_corr')
             A dipole object with experimental data.
-        n_trials : int (Optional if obj_fun='dipole_rmse')
+        n_trials : int (Optional if obj_fun='dipole_rmse' or 'dipole_corr')
             Number of trials to simulate and average.
         f_bands : list of tuples (Required if obj_fun='maximize_psd')
-            Lower and higher limit for each frequency band.
+            Lower and higher limit for each frequency band in Hz. # ATTN @ntolley This
+            is Hz, right?
```
ntolley:
```suggestion
            Lower and higher limit for each frequency band in Hz.
```
Yes this is correct!

hnn_core/optimization/objective_functions.py
```diff
@@ -86,14 +87,27 @@ def _rmse_evoked(
...
+        List to store objective function values.
     tstop : float
         The simulated dipole's duration.
-    target : instance of Dipole
-        A dipole object with experimental data.
-    n_trials : int
-        Number of trials to simulate and average.
-    verbose : bool
-        If True, print build steps and simulation progress to console. Default: True.
+    obj_fun_kwargs : dict
```
ntolley, line 94: Thank you for adding this! Just read is thoroughly and everything seems correct and much clearer

hnn_core/optimization/objective_functions.py
```diff
@@ -153,8 +167,14 @@ def set_params_batch(a, b):
         new_net = initial_net.copy()
         set_params(new_net, params)
 
+        # ATTN: @ntolley: Is this right? I think this is what is run during the non-CMA
```
ntolley: Ah great catch @asoplata!

You're correct I forgot to update the non-batch case with the `obj_fun_kwargs`

hnn_core/optimization/objective_functions.py
```diff
@@ -165,6 +185,13 @@ def set_params_batch(a, b):
 
     obj_values.append(obj)
     print(f"Mean Loss: {np.mean(obj):.2f}; Min Loss: {np.min(obj):.2f}")
+
+    # update best params; this is a "side-effect" that changes the `best` dictionary
+    # in-place in the parent scope
+    if best is not None and obj < best["obj"]:
```
ntolley: I'm not sure if this would work in the batch case

`obj` will be a list and `predicted_params` will be a 2D array which matches the length of that list

I think that there will need to be an `np.argmin(obj)` in the batch case to identify which solution has the lowest score of the population, and then compare that to the current `best["obj"]`

hnn_core/optimization/objective_functions.py
ntolley, line 162:
```suggestion
        obj = [_rmse(dpl, obj_fun_kwargs["target"], tstop=tstop) for dpl in dpls]
        best_obj_idx = np.argmin(obj)
        best_obj = obj[best_obj_idx]
        best_params = predicted_params[best_obj_idx, :]
```
@asoplata one approach would be to add these variables to the batch case (in the non-batch case `best_obj = obj`)

Then you can re-use the code below which compares the current best_obj to all the previous ones

hnn_core/optimization/objective_functions.py
asoplata: Good catch, I forgot that it's multidimensional

hnn_core/optimization/objective_functions.py
asoplata: I'm a bit confused now: `best` is only a parameter in `_run_opt_cobyla`, and it is only that instance of `best` that will be changed by the code here in these objective functions. We know that Cobyla will only ever execute the non-batch block of code. In contrast, in CMA, we already appear to have a way to get the "best" parameters that is different than in the Cobyla case, here: 
https://github.com/carolinafernandezp/hnn-core/blob/531703e6633da8e33f30cb27dfcce49adb305014/hnn_core/optimization/general_optimization.py#L568-L569
where it seems to be taken from the CMA object directly. 

My question now is, doesn't this alternative "best" approach in `_run_opt_cma` obviate (make unnecessary) any `best` handling inside the `is-batch` blocks of the objective functions here? We can totally take the Cobyla's `best` approach and make the changes you've suggested and add this approach to both `_run_opt_cma` and both dipole objective functions. However, first, I just want to ask and make sure that this is necessary, and that `es.result.xbest` isn't already taking care of this for us (I don't know, but I think you added that line).

hnn_core/optimization/objective_functions.py
asoplata: See my comment on the other one

asoplata:
- [x] TODO add test as discussed over `_rmse_evoked`, and indent block and update comment

hnn_core/optimization/objective_functions.py
asoplata, line 390: @ntolley I just noticed that this line was left in as a comment from the Anticorr PR. Was that intentional? Is this code supposed to be live and uncommented, or can it be deleted? Its presence doesn't seem to change the tests, and I don't think `params` is used inside the `is_batch` block. Are you okay if I delete it?

asoplata:
I've done what we discussed, including moving the `best` code inside the non-CMA block in all objective functions, adding some more comments to make those blocks clearer, and added a test of the `best` code. I ended up going with Claude's solution which, since reproducing a consistently non-monotonically-decreasing series of objective function values is extremely difficult, Claude suggested instead to use a mock objective function which mainly exists to test an identical version of the `best` storage. I've reviewed the test and it looks pretty good to me (after some refactoring). I've also done a merge from `master` so all our changes are up to date.

hnn_core/tests/test_general_optimization.py
```diff
@@ -473,3 +473,73 @@ def set_params(net, params):
...
+    tstop = 10.0
+    net = neymotin_2020_model(mesh_shape=(3, 3))
+
+    def set_params(net, params):
+        pass
+
+    constraints = {"mu": (1.0, 6.0), "sigma": (1.0, 3.0)}
+
+    call_log = []
+    call_counter = [0]
```
ntolley:
```suggestion
    call_counter = 0
```

hnn_core/tests/test_general_optimization.py
```diff
@@ -473,3 +473,73 @@ def set_params(net, params):
...
+        predicted_params,
+        update_params,
+        obj_values,
+        tstop,
+        obj_fun_kwargs,
+        best=None,
+    ):
+        call_idx = call_counter[0]
+        obj = obj_sequence[min(call_idx, len(obj_sequence) - 1)]
+        call_counter[0] += 1
```
ntolley:
```suggestion
        call_counter += 1
```
why does the counter need to be a list?

hnn_core/optimization/objective_functions.py
ntolley: yes you can delete it! this was accidentally left in

the goal of `update_params` is to turn an array of parameter values into a dictionary that can be interpreted by `set_params`

in the batch case this is transformed into a similar dictionary a few lines down, but this one is actually passed into BatchSimulate

hnn_core/tests/test_general_optimization.py
asoplata: The answer, as it turns out, is a bit complicated, and has to do with scoping of local variables. Claude was helpful in explaining its "reasoning". 

If I try to run the test with the changes you've applied, I get a `UnboundLocalError`. According to Claude, which is supported by other tutorials such as https://www.datacamp.com/tutorial/scope-of-variables-python (emphasis mine) :

> On line 512, call_idx = call_counter is a read — that works fine. But on line 514, call_counter += 1 is syntactic sugar for call_counter = call_counter + 1, which is an assignment. **The moment Python sees any assignment to a name inside a function, it classifies that name as a local variable for the entire function** — including the read on line 512. Since no local call_counter has been assigned yet at line 512, you get UnboundLocalError.
>
> Python decides at parse time whether a name is local to a function: if there is any assignment statement (=, +=, etc.) targeting that name anywhere in the function body, Python marks the name as local for the entire function. That's why even the read on line 512 fails — Python already decided call_counter is local before it runs.
>
> `nonlocal` tells Python to look up call_counter in the enclosing test_cobyla_best scope rather than treating the += as a local variable definition.

Thus, if we take your changes, but we add `nonlocal call_counter` to the first line of `mock_obj_fun`, that does allow the test to pass and the variable to be used properly. I will add a commit with these changes.

In the case of using a list, however, a list works because we're not rebinding inside the local scope (again, from Claude):

> The connection to mutability is indirect: because integers are immutable, you can't mutate them in place — you're forced to rebind the name (call_counter = call_counter + 1), which triggers the scoping issue. With mutable objects like lists you can sidestep rebinding entirely.
>
> Instead of:
> call_counter += 1   # assignment → name becomes local
>
> You'd write something like:
> call_counter[0] += 1   # assignment to call_counter[0], not to call_counter
>
> Python never sees an assignment to the name call_counter itself, so it looks it up in the enclosing scope and finds the list. The [0] += 1 part is a mutation of the object the name points to, not a rebinding of the name.
>
> So the real rule is: assignment to a name makes it local; method calls on a name do not.

To get into even more detail, something that I was investigating offline was why using mutable types as arguments is bad, which is kind of related. Here's a gist that illustrates why https://gist.github.com/asoplata/bfbc19fa345c1c77566e81275ef9ee65 . It's not directly related, but it does make clear that "inside a local scope, you can freely access and mutate a named Mutable Type (i.e. list, dict, set), but not after you *re-assign* that name. Once you re-assign a name locally, you have just created a new instance of that type that is now both only inside the local scope and independent of the same name in the outer scope" (my words).

So. To summarize, to get around this problem, we can either declare `call_counter` to be `nonlocal` (which I will do, since it's a more explicit solution) or we can force it to be a type that we're allowed to mutate from inside another function (such as list, dict, set) as long as we don't rebind the name.

PR #1248, opened 2026-03-04
[MRG] DOC: add dask[distributed] to parallel extras and document dask backend

Yaswanth8390:
Related to #955 ,
This PR addresses points 1 and 2 from @asoplata's comment:

1. Added dask[distributed] to the parallel as it is specifically required for joblibs dask backend to connect to a client.

2. Documentation: Added a section demonstrating how to use dask backend.During testing, multiple worker threads caused NEURON crashes.Running with --nthreads=1 and multiple workers(--nworkers=8) resolved the issue and all simulations completed successfully.

For single-machine use,loky remains simpler and recommended,while dask is more powerful when scaling across multiple machines.


asoplata:
I'm pretty confused, which `if __name__=='__main__'` are you referencing? If you mean the `if` in `setup.py`, then we definitely do not want to add any `dask`-dependent code (except for what you've already added to the `parallel` extra), since that is the main file used for all package installation and we need to be very careful with it.

It's far better to do all the `dask` setup within the file where you're actually going to use it. I would propose instead to add new code to `plot_batch_simulate.py` only: first, I would add code that detects if `dask` has been installed, and if so, uses `LocalCluster` to run some simulations, in order to show how it's generally used. Then, I would expand your current comment in that file to document what changes need to happen for a user to do the exact same code for a cluster instead of their local computer (you have mostly already done the comment).

asoplata:
Thanks for making the requested changes Yaswanth, and for identifying linux-specific issues. I will work on trying to find a different solution to the Linux tkinter warning issue. Because this is a tutorial notebook, it is best if we still try to move the Dask content after the Loky example and the plotting code. We also need to ensure that the Loky example is run even if Dask is not installed. Feel free to try to make these changes yourself, but don't feel obligated since I will be pushing commits to the branch to resolve these myself.

Yaswanth8390:
I moved the Dask part to run after plotting and moved loky outside the try block so it still runs even if Dask isn’t installed.

Also updated setup.py to include both dask and dask[distributed] as both are required .

And as for the Tkinter warnings on Linux, it seems to be due to multiprocessing (fork) inheriting the Matplotlib/Tkinter state after plotting, which causes cleanup issues when the child processes exit. Looks like a known Matplotlib issue as well: https://github.com/matplotlib/matplotlib/issues/29198.
I’ll also take a look to see if there’s a way to handle this within the same script as well.

Yaswanth8390:
Hey @asoplata, @ntolley ,
Thanks for the changes.

I've tested the code changes after installing `"dask[distributed]"`, and I have got the same tkinter issue as before.

It seems related to Matplotlib’s Tk backend with multiprocessing (fork), where Tk objects are cleaned up later outside the main loop(via garbage collection), causing:
`RuntimeError: main thread is not in main loop`.We can directly call gc.collect() at the end of `plot_results()` but it is an expensive call and might crash the process in some cases.

Hence,I have a created separate processes for plotting such that they execute separetely and execute smoothly without any warnings.

```python
from multiprocessing import Process
p1=Process(target=plot_results, args=(loky_results,))
p1.start()
p1.join()
p2=Process(target=plot_results, args=(dask_results,))
p2.start()
p2.join()
  ```

This isolates the Tk lifecycle per process and avoids the warnings.

That said, if introducing separate processes for plotting is not desirable for the example, I agree that splitting the Loky and Dask workflows into separate scripts/notebooks would be a cleaner and more maintainable approach, especially since the issue seems to arise when combining multiple backends and plotting in the same run.

Before I make any commits to the branch I wanna make sure if this works cleanly in other systems as well.

**Comment by asoplata** [comment]
Argh, yeah it did seem like a really brittle solution. Even if it worked for us, it's better for our users if we keep things as clean as possible and simply separate them into distinct files. Yaswanth do you mind doing that? Once they are split into two files then this PR is probably ready for merge (after final review).

**Comment by asoplata** [comment]
@ntolley What do you think about making `dask[distributed]` a separate `extras` install in `setup.py` from `parallel`? I suggest this because `hnn-core[parallel]` is probably (or at least should) be MOST installs, but `dask` is probably going to be installed by very few users, at least for the near future. 

In other words, in the interest of keeping our dependencies tight, I propose we keep `hnn-core[parallel]` as it is on master, but we add a new `hnn-core[dask]` "extra" install option that requires `hnn-core[parallel]`. What do you think?

Also I'm surprised I missed this, but I realized that this PR adds no testing of the `dask` usage, which it definitely needs to do.

**Comment by Yaswanth8390** [comment]
I've created a separate file for Batch Simulation using Dask, and it works in Linux and other systems without issues.

I’ll add a test specifically for the Dask backend, since `test_run()` already covers the loky backend.

**Comment by asoplata** [comment]
@ntolley @dylansdaniels  barring new errors, I believe this is ready for review and merge. There mostly only documentation and `setup.py` changes applied that are different from Yaswanth's last commit. I've documented what the changes are inside each commit message (mainly better `ImportError` handling in the example scripts).

File: doc/roadmap.md [hunk 1]
```diff
@@ -27,7 +27,7 @@ interactive GUI is essential for all new users of HNN to develop an
 intuition on how to interact with the large-scale computational model to
 study the multi-scale neural dynamics underlying their MEG/EEG data.
 Once this intuition is gained, users who chose to can dive into the
-computational neural modeling code, where further command line utily can
+computational neural modeling code, where further command line utility can
```
ntolley, line 30: might have a merge conflict with #1297

examples/howto/plot_batch_simulate.py
```diff
@@ -146,50 +131,67 @@ def summary_func(results):
 #
 # Stronger synaptic connections (yellow lines) generally show larger
 # amplitude responses and more pronounced features throughout the simulation.
+def plot_results(simulation_results):
```
ntolley, line 135: Very nice! Much cleaner and re-usable

examples/howto/plot_batch_simulate.py
ntolley:
```suggestion
for comprehensive analysis. The BatchSimulate class permits embarrassingly parallel
simulations over multiple CPU cores.
```

examples/howto/plot_batch_simulate_dask_backend.py
```diff
@@ -0,0 +1,220 @@
+"""
+======================================
+08. Batch Simulation with Dask Backend
+======================================
+
+This example shows how to do batch simulations in HNN-core, using the Dask backend,
+allowing users to efficiently run multiple simulations with different parameters for
+comprehensive analysis across potentially very many cores.
```
ntolley:
```suggestion
comprehensive analysis. The BatchSimulate class permits embarrassingly parallel simulations over multiple CPU cores. While Dask can be used on a locally on a personal computer, it is particularly useful for parallelization with high-performance computing clusters (HPC's) which potentially provide access to CPU cores spread across multiple nodes. If HPC's are not being used, the default 'loky' backend is recommended. 
```

examples/howto/plot_batch_simulate_dask_backend.py
```diff
@@ -0,0 +1,220 @@
...
+    plt.xlabel('Synaptic Strength (nS)')
+    plt.ylabel('Dipole Peak Magnitude')
+    plt.title('Min and Max Dipole Peaks across Simulations')
+    plt.legend()
+    plt.grid(True)
+    plt.xscale('log')
+    plt.tight_layout()
+    plt.show()
+    
+if __name__ == "__main__":
```
ntolley: Why is there a main block? This is not a normal python script as it gets turned into a jupyter notebook by sphinx

examples/howto/plot_batch_simulate_dask_backend.py
```diff
@@ -0,0 +1,220 @@
...
+    batch_simulation = BatchSimulate(
+        net=net,
+        set_params=set_params,
+        summary_func=summary_func,
+    )
+
+    ###############################################################################
+    # Using Dask backend for distributed computing
+    # --------------------------------------------
+    # Dask can also be used as a backend for distributed computing.
```
ntolley:
```suggestion
    # Dask can be used as a backend for distributed computing.
```
Why also? We only show the distributed computing use-case

examples/howto/plot_batch_simulate_dask_backend.py
```diff
@@ -0,0 +1,220 @@
...
+    # For local use, ``LocalCluster`` can be used as shown below.
+    # Note that ``threads_per_worker=1`` is required to avoid NEURON conflicts.
```
ntolley: This one comment would have saved by quite a lot of headaches early in grad school 😄

hnn_core/tests/test_batch_simulate.py
```diff
@@ -312,3 +312,36 @@ def test_parallel_execution(batch_simulate_instance, param_grid):
...
+    from dask.distributed import Client, LocalCluster
+
+    param_combinations = batch_simulate_instance._generate_param_combinations(
+        param_grid
+    )
+
+    cluster = LocalCluster(processes=True, n_workers=2, threads_per_worker=1)
+    client = Client(cluster)
+
+    try:
```
ntolley, line 330: Just to be sure, if an error is thrown in the `try` block, does pytest still detect it when there is a `finally` block?

ntolley:
Finished the first round of review @asoplata! Very minor changes requested, should be ready to merge after addressing those and a rebase

examples/howto/plot_batch_simulate_dask_backend.py
ntolley:
```suggestion
```
Can we just remove it an un-indent everything inside?

Yaswanth8390: Without it, Dask's spawn-based workers reimport the module and recursively create new clusters causing a freeze/hang and a runtime error.

hnn_core/tests/test_batch_simulate.py
Yaswanth8390: Finally only handles cleanup; the exception still propagates to the pytest as usual.

examples/howto/plot_batch_simulate_dask_backend.py
asoplata: This may have been added by me when I was trying to debug getting Dask and non-Dask backends working the same script. There's a good chance that it can be taken out now that Dask is in its own script. Additionally, Nick's comment is right since this script is provided both as a raw script and as a Jupyter notebook, where this line will not apply. I'll test that it works without the `if`

PR #1263, opened 2026-03-18
[WIP] Refactor test_hamming_smoothing using pytest parameterization to improve clarity and extensibility

bhumigaddam:
While going through `test_hamming_smoothing`, I noticed that a few similar input cases were being tested using loops, which made the test a bit harder to follow.

In this PR, I’ve refactored those cases using pytest parameterization so each scenario is more clearly visible and easier to extend in the future. Adding new cases now should be much simpler without making the test more complex.

I also added a module-level skip for neuron-dependent tests so they don’t fail in environments where NEURON isn’t installed, making it easier to run the relevant tests locally.

This keeps the existing behavior intact, but makes the test cleaner and more contributor-friendly.


asoplata:
Hello bhumigaddam,

Are you still interested in pursuing the work in this PR? If not, or if we do not hear from you within a month's time, we will close the PR. Thank you for any time you have.

PR #1266, opened 2026-03-22
Fix _hammfilt input validation and improve test coverage

SatvikSaluja:
This PR replaces the use of `assert` in `_hammfilt` with explicit input validation and proper exception handling.

`assert` statements can be removed when Python is run with optimizations, which makes them unsuitable for validating user inputs or runtime conditions. Replacing them with explicit exceptions ensures consistent behavior and improves robustness.

Changes:
- Added type checks for `x` and `winsz`
- Ensured `winsz` is positive and smaller than input length
- Replaced assertion with appropriate exceptions
- Updated tests to use modern `pytest.raises` context manager
- Added additional test cases to cover edge conditions

### DIFF
--- a/hnn_core/tests/test_utils.py
+++ b/hnn_core/tests/test_utils.py
@@ -26,7 +26,7 @@ def test_hamming_smoothing():
     # Tests of data argument
     # ---------------------------------------------------------------
     # Even one sample short should raise an error
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="Window size is too long"):
         smooth_waveform(
             np.arange((1e-3 * default_window_len * default_sfreq) - 1),
             default_window_len,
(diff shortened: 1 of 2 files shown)


asoplata (changes requested):
Thanks for this PR -- I actually did not know about how `python -O` and `-OO` can ignore runtime `asserts`! This is definitely something that we at HNN need to address, since some of our code relies on it for argument validation.

In this PR's specific case, I do not think we actually need to add most of these checks and tests to `_hammfilt`, because most of them are already present when using or testing `smooth_waveform` which is the sole consumer of `_hammfilt`, and `_hammfilt` is also a private function.

SatvikSaluja (commented):
I made a few updates to align the tests and implementation with the current logic. Since the window size is now computed from window_len and sfreq (instead of being used directly), I adjusted the test inputs so that the computed window size still reflects the original condition being tested.

While working on this, I encountered an issue where small window_len values (e.g., 2.5 ms with low sfreq) resulted in a computed window size of 0. This caused invalid behavior during convolution (np.hamming(0) → empty array). To fix this, I added a validation in smooth_waveform to ensure the computed window size is at least 1, and updated the corresponding test to expect a ValueError.

I also restored test cases like None to maintain full coverage for both type and value validation, as intended.

The only slight deviation from the original tests is adjusting parameters to preserve the original test intent under the new computation-based implementation of window size.

asoplata (approved):

Issue #1149, opened 2025-09-24
REF: Remove celltype hardcoding from viz.py


asoplata:
Chetan's addition of metadata fields for different celltypes in #1099 has offered a great solution for the inconsistent customization in `viz.py` across celltypes. Different plotting functions can differ wildly, including reading or not reading a network's celltypes, assigning colors to celltypes explicitly or differently or using different colorschemes, allowing or not allowing support for subsets of celltypes (including with or without color-matching), etc. Therefore, I think the following work should be done:

1. Our default cell metadata (located here https://github.com/jonescompneurolab/hnn-core/blob/0279e31183bb9850f87fce3a9b0a8e413b492c72/hnn_core/network_models.py#L78 ) should be refactored to outside of `jones_2009_model()` for use elsewhere such as in #1119 .
2. Each celltype should have one color to be used consistently, and the same for its "marker" in some uses of `viz.py`
3. `viz.py`'s plotting functions should be updated so that they never make assumptions about celltypes (as much as possible, e.g. sorting alphabetically by celltype name), and they only ever read a celltype's colors instead of setting them inside the function itself.

Tusharjamdade:
Hi @asoplata and @dylansdaniels,
Is this issue still valid? If so, I’d be happy to work on it.

PR #1269, opened 2026-03-23
[MRG] Viz meta hardcode

Chetank99:
wip
will document in a bit


asoplata:
For Nick and I: documentation on the design changes are in https://www.notion.so/jonescompneurolab/MetaData-Comes-to-Viz-Town-32644dbbcce680339464cbb017603db7#34144dbbcce6803da9aee54f0b316ba5

asoplata:
- [x] TODO austin double check other hardocded color sequns in viz (I couldn't find any other hardcoded sequences in viz)

asoplata:
Resolves #1149

asoplata:
Hey Chetan, I've made a secondary PR here https://github.com/Chetank99/hnn-core/pull/2 with some changes I'm wanting to make to this PR. I've added a few more usages of your default colors, but I'm not yet done reviewing `viz.py` for other places where we may want to use the default metadata more. So far, however, after a bit more checking, I think this will be ready for merge -- we can change the colors in the future (since now you've made that easy to do). Please review and accept it if you agree to the changes

hnn_core/viz.py
```diff
@@ -705,14 +730,17 @@ def plot_spikes_raster(
...
+
+    # default colors from cell metadata
+    from .network_models import default_cell_metadata as _meta
 
-    # Set default colors
-    default_colors = plt.rcParams["axes.prop_cycle"].by_key()["color"][
-        : len(cell_types)
-    ]
-    cell_colors = {cell: color for cell, color in zip(cell_types, default_colors)}
+    cell_colors = {cell: _meta.get(cell, {}).get("color", "k") for cell in cell_types}
```
asoplata, line 756: This is a very nice line, going to copy this idiom :)

asoplata:
Finished going over the rest of `viz.py`, and I don't think there's anything left to change. If we decide to merge my secondary PR above ( https://github.com/Chetank99/hnn-core/pull/2 ) and rebase onto master, then I think this is ready for merge.

examples/default_meta_temp_run.py
```diff
@@ -0,0 +1,28 @@
...
+for cell_name, meta in default_cell_metadata.items():
+    print(f"\n  {cell_name}:")
+    for key, value in meta.items():
+        print(f"    {key:20s} = {value}")
+
+# look up a single cell type's metadata directly
+print("\nL5_pyramidal defaults:")
+print(f"  {default_cell_metadata['L5_pyramidal']}")
+
+# fill missing keys from defaults
```
ntolley: I don't totally understand what this is trying to show off or accomplish

examples/default_meta_temp_run.py
ntolley, line 1: I'm not sure if this example is really achieving the goal of helping understand the structure and role of adding cell metadata

Does this actually need it's own example? The details here seem like they would belong as documentation

Chetank99: Hey @ntolley this was just a temp file created to check if existing stuff is being broken or not. It'll not go in the final merge. Something to just for the notion page. This along with the temp place holder comments will go away before the merge..

Chetank99: with the clean up provided by @asoplata it should be good to go while not merging this file

this was initially discussed here https://app.notion.com/p/jonescompneurolab/MetaData-Comes-to-Viz-Town-32644dbbcce680339464cbb017603db7?source=copy_link#32744dbbcce68090804ed9046e70a747

hnn_core/viz.py
Chetank99: Thank you sensei!

hnn_core/viz.py
```diff
@@ -552,6 +577,12 @@ def plot_spikes_hist(
         spike_label: list() for spike_label in np.unique(list(spike_labels.values()))
     }
     spike_color = dict()  # Store colors specified for each spike_label
+    # NOTE: Currently, since `CellResponse` only contains the "type" (aka drive name or
+    # cell name) of what produced each spike, but not whether that drive was proximal or
```
katduecker, line 581: This is not true! Cell types are taken from the cell_types dictionary when gathering trial data

https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/parallel_backends.py#L44

and this info is then used in cell_response._cell_type_names https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/cell_response.py#L132

asoplata: Your point is true only in the case that we are using a `CellReponse` that has just been created by a fresh simulation, but not if we are loading a `CellResponse` object from a spike times file, such as using the partially-deprecated `read_spikes` function https://github.com/katduecker/hnn-core/blob/73d4eb60640669922ad1a53be8876fbca43602c1/hnn_core/cell_response.py#L539 . I say "partially" because unfortunately  we do NOT currently document another way to actually re-load existing spike time data.

PR #1271, opened 2026-03-29
[MRG] ENH: Add support for custom loss functions in Optimizer

### LINKED ISSUE #1147: OPT: Add support for non-RMSE loss functions
(This is copied/edited from an email @ntolley sent)

We should add a feature that permits custom loss functions in the optimizer. Previously the only loss function available was RMSE, which only makes sense if you are fitting signals that have a clear start/end.

Poisson drives are mainly used to produce steady state oscillations, so there needs to be different loss functions suitable for this kind of signal. The “relative power” loss that’s now implemented is one attempt we made at a loss function that makes sense for signals *generated* by Poisson drives. I think a better solution would be to have predefined losses like we already have, with the option for users to supply their own loss function.

vshnvii:
Fixes #1147 

Previously, the `Optimizer` required users to use predefined loss functions like `"dipole_rmse"` (which works best for signals with a clear start/end) or `"maximize_psd"`. While users could supply a custom callable to `obj_fun`, doing so required them to redefine all the internal simulation boilerplate (creating the network copy, running `simulate_dipole`, scaling/smoothing, etc.). 

For cases like steady-state oscillations generated by Poisson drives, users need a straightforward way to swap out just the loss calculation metric without writing the entire simulation loop from scratch.

### Changes
* **Added `_custom_objective_function`** in `hnn_core/optimization/objective_functions.py`. This acts as a generic wrapper that handles standard network parameter updates and simulations.
* **Updated `Optimizer.__init__`** to support the `obj_fun="custom"` flag.
* **Updated `Optimizer.fit()`** to accept and validate a `loss_fun` keyword argument when `"custom"` is chosen.
* **Added `test_custom_loss_fun`** in `test_general_optimization.py` to ensure the custom loss function pathway operates correctly and respects boundary constraints. 

### Usage Example
Users can now simply pass a custom metric logic without doing the dipole simulations themselves:

```python
def my_custom_loss(dpls, obj_fun_kwargs):
    from hnn_core.dipole import average_dipoles
    dpl = average_dipoles(dpls)
    # Define calculation logic returning a float...
    return custom_score 

optim = Optimizer(
    net,
    tstop=10.0,
    constraints=constraints,
    set_params=set_params,
    obj_fun="custom"
)

optim.fit(loss_fun=my_custom_loss, target=target_dpl)
```


asoplata:
Hello Vaishnavi, #1221 has been merged, so would you mind rebasing this on top of the latest `master` branch? This will be a pretty cool feature when it's added!

vshnvii:
done !

hnn_core/optimization/general_optimization.py
```diff
@@ -110,6 +115,9 @@ def __init__(
         elif obj_fun == "dipole_corr":
             self.obj_fun = _anticorr_evoked
             self.obj_fun_name = "dipole_corr"
+        elif obj_fun == "custom":
+            self.obj_fun = _custom_objective_function
+            self.obj_fun_name = "custom"
         else:
             self.obj_fun = obj_fun  # user-defined function
             self.obj_fun_name = None
```
asoplata, line 126: @ntolley If we add the "custom objective function" support from this PR, should we drop this support for "user-defined functions?" I think we should either drop "user-defined function" entirely, or fully document it (which includes a description of what parameters every valid objective function has to take like all the obj funcs in `objective_functions.py`, support for `BatchSimulate` now, etc.). Any "user-defined function" that a user provides (I believe) has to be at least as complicated as Vaishnavi's new `objective_functions.py::_custom_objective_function`. What do think?

hnn_core/optimization/general_optimization.py
```diff
@@ -51,7 +56,7 @@ def __init__(
             The optimizer, 'bayesian', 'cobyla', or 'cma'.
         obj_fun : str | func
             The objective function to be minimized. Can be 'dipole_rmse',
-            'maximize_psd', "dipole_corr", or a user-defined function. The default is
+            'maximize_psd', "dipole_corr", "custom", or a user-defined function. The default is
             'dipole_rmse'.
```
asoplata:
```suggestion
            'maximize_psd', 'dipole_corr', 'custom', or a user-defined function. See the docstring for
            ``Optimizer.fit`` to view the required and optional arguments for each of these cases. The
            default is 'dipole_rmse'.
```

hnn_core/optimization/general_optimization.py
asoplata, line 188:
```suggestion
        loss_fun : callable (Required if obj_fun='custom')
            Objective loss function to be provided by user. The function must accept a list of 
            ``Dipole`` objects as its first argument, and the remaining `kwargs` arguments 
            will be taken from the `obj_fun_kwargs` passed here directly to ``Optimizer.fit``.
        sigma0 : float| array-like (Only used if solver='cma')
```

hnn_core/optimization/general_optimization.py
```diff
@@ -217,6 +225,14 @@ def fit(self, **obj_fun_kwargs):
         ):
             raise Exception("f_bands and relative_bandpower must be specified")
 
+        elif self.obj_fun_name == "custom":
+            if "loss_fun" not in obj_fun_kwargs:
+                raise Exception("loss_fun must be specified when obj_fun='custom'")
+            elif not callable(obj_fun_kwargs["loss_fun"]):
+                raise Exception("loss_fun must be a callable")
+            elif "n_trials" not in obj_fun_kwargs:
+                obj_fun_kwargs["n_trials"] = 1
```
asoplata: @ntolley Should setting default `n_trials` to 1 be "promoted" to be true across all objective functions? Only `_maximize_psd` appears to FORCE it to be 1 (overriding user input, which should be noted in a print statement), while currently all other objective functions use it.

hnn_core/optimization/objective_functions.py
```diff
@@ -384,3 +384,62 @@ def set_params_batch(a, b):
...
+    obj_fun_kwargs : dict
+        Additional arguments to pass to the objective function, must contain
+        'loss_fun', a callable.
+
+    Returns
+    -------
+    obj : float
+        The loss value returned by the user-defined loss function.
+    """
+
```
asoplata: @ntolley My first thought in reviewing this was "this now needs to be updated to include the CMA `is_batch` case, along with all that code..." but I realized that MUCH of your new `is_batch` control flow code in the above pre-existing functions is actually almost identical between each function. I'm going to try to add a secondary PR onto this one that should offer a large amount of refactoring.

hnn_core/tests/test_general_optimization.py
```diff
@@ -473,3 +473,103 @@ def set_params(net, params):
     # The optimization results with different seeds should be different
     assert not np.allclose(optim_seed1.obj_, optim_seed2.obj_)
     assert not np.allclose(optim_seed1.opt_params_, optim_seed2.opt_params_)
+
+
+@pytest.mark.parametrize("solver", ["bayesian", "cobyla"])
```
asoplata:
```suggestion
@pytest.mark.parametrize("solver", ["bayesian", "cma", "cobyla"])
```

This should only be done after batch (CMA) support is added to the custom function caller. I'll try to add a way that works in my own PR onto this one.

asoplata:
@ntolley I've added a secondary PR to Vaishnavi's branch here https://github.com/vshnvii/hnn-core/pull/1 which appears to be a successful attempt at major deduplication of the objective function code, while still supporting Vaishnavi's new custom loss function feature.

hnn_core/optimization/objective_functions.py
asoplata: did so at https://github.com/vshnvii/hnn-core/pull/1

hnn_core/optimization/general_optimization.py
asoplata: made change in https://github.com/vshnvii/hnn-core/pull/1

hnn_core/tests/test_general_optimization.py
asoplata: made it so in https://github.com/vshnvii/hnn-core/pull/1

asoplata:
Hey Vaishnavi, don't worry about responding to the comments I've made in review on this PR, I will apply them and finish out this PR alongside Nick.

hnn_core/optimization/general_optimization.py
asoplata: My opinion now is that we should NOT drop this

hnn_core/optimization/objective_functions.py
```diff
@@ -161,13 +156,64 @@ def set_params_batch(a, b):
...
+    initial_net : instance of Network
+        The network object.
+    initial_params : dict
+        Keys are parameter names, values are initial parameters.
+    set_params : func
+        User-defined function that sets network drives and parameters.
+    predicted_params : list
+        Parameters selected by the optimizer.
+    update_params : func
+        Function to update params.
```
ntolley, line 224:
```suggestion
    update_params : func
        Function to update params.
    obj_values : list
         List of objective values for each epoch (aka iteration) during optimization.
         Updated as a side effect of evaluating the objective function.
```

hnn_core/tests/test_general_optimization.py
```diff
@@ -473,3 +473,102 @@ def set_params(net, params):
...
+            weights_ampa=weights_ampa,
+            synaptic_delays=synaptic_delays,
+        )
+
+    # define constraints
+    constraints = dict()
+    constraints.update({"mu": (1, 6), "sigma": (1, 3)})
+
+    def custom_loss(dpl, obj_fun_kwargs):
+        from hnn_core.dipole import _rmse
```
ntolley: We're trying to cut down on mid-code imports I believe?

asoplata:
Hey @ntolley unfortunately this needs further review, but only for the `best` and `obj_value` changes in  https://github.com/jonescompneurolab/hnn-core/pull/1271/changes/c22dd4d5689f12af462ef77e202b09a4e1881660 . In the process of merging from `master`, I added the new "best objective" support into ALL objective functions. Does that make sense? All tests appear to pass.

asoplata:
Merged, thanks @vshnvii !

PR #1273, opened 2026-03-29
[MRG] Re-enable interactive plotting buttons in GUI

### LINKED ISSUE #1125: BUG: Re-introduce interactive plotting buttons
George claimed that at some point in the past, data plots in the GUI showed their interactive-Python buttons, allowing some features like zoom, etc. These buttons no longer display. George suggests that it may be due to removing some usage of `ipythonmpl` or a similar library. We should re-enable this.

Tusharjamdade:
## Summary
This PR restores interactive plotting behavior in the GUI by improving backend detection in the visualization manager.

Previously, the GUI was incorrectly falling back to the static rendering path (`_static_rerender`) instead of using the interactive rendering path (`_dynamic_rerender`), which caused the interactive plotting buttons to not appear.

To fix this, I updated the backend detection logic in `hnn_core/gui/_viz_manager.py` so that interactive notebook backends are correctly detected before selecting the rendering path.

## Changes
- Updated backend detection in `hnn_core/gui/_viz_manager.py`
- Added detection for interactive notebook backends:
  - `ipympl`
  - `nbagg`
  - `widget`
- Prevents fallback to static rendering when interactive rendering is available

Closes #1125


asoplata:
Thanks Tushar! We will try to get to your PR as soon as we can; we're currently extra busy on account of many of us going to a conference this week, and we have been preparing for a new release for the past several weeks (out very soon). Thank you for your patience, and for your contribution!

asoplata:
@asoplata @dylansdaniels We need to look at how this potential fix interacts with static vs dynamic plot handling that is currently on `master` from Dylan's GUI visual refresh

asoplata:
@dylansdaniels As far as I can tell, there were no real changes relevant in the GUI visual refresh #1243 that are relevant to the dynamic vs static rendering of `_VizManager` here (distinct from dynamic vs static output widgets in `gui.py`), the only exception being a comment here https://github.com/jonescompneurolab/hnn-core/pull/1243/changes#diff-54fec781f6f632abe66e0eaccf3884ea80ddf6a1aaa3821f1c2a36e32efdaaa3R611-R614 . I definitely may have missed something, but that combined with the fact that there was no merge conflict for this PR before or after the visual refresh suggests that there were no changes applied to control of backends or decisions to change dynamic with static (& vice versa) in `_VizManager` during the refresh. If this is true, then this PR may be able to make the changes we want cleanly, which is good news

asoplata:
Hello Tushar, I have rebased your branch onto `master` in order to get it up to date with other GUI changes; you will have to redownload the branch again from `origin` if you want to make any further changes. We have some help on how to do this at https://jonescompneurolab.github.io/hnn-core/stable/contributing.html

asoplata:
@dylansdaniels @ntolley I've made an optional secondary PR onto Tushar's branch here https://github.com/Tusharjamdade/hnn-core/pull/2 .

My PR makes a small but important change: instead of simply detecting the backend and using `ipympl` only if it is both available AND currently chosen automatically (based on the user's system), for the GUI via `_VizManager`, my PR now attempts to force usage of the `ipympl` backend. Similar to the comments I made in the code, I think it is safe to override the user's backend default in this case:

1. `ipympl` is guaranteed to be installed when a user is using the GUI, since it is included in `[gui]` extras.
2. This is only a backend override for use with the GUI itself, not for any API plotting calls.

If `ipympl` is always installed, then I think it's probable that we will NEVER need the static rendering code, as opposed to the dynamic code. However, I've left the static rendering code untouched, and for the same reason (just in case), I've also left in Tushar's fallback in case `ipympl` can't be used for whatever reason.

You can see where this new change to the `ipympl` backend passes the tests at https://github.com/asoplata/hnn-core/actions/runs/26297321416 and https://github.com/asoplata/hnn-core/actions/runs/26297321414

I think the static vs dynamic rendering code is mostly a holdover from the early days when multiple different backends were either unclear or were expected to be used (not to mention version instability around the time when the GUI was first being created). However, if we can be reasonably sure that the GUI is using `ipympl` now with this PR, it's possible that we may no longer need the static rendering code at all. I'm hesitant to delete it in the case of fallbacks however.

I think that with my https://github.com/Tusharjamdade/hnn-core/pull/2 , that this PR of Tushar's is ready for merge. What do y'all think?

PR #1280, opened 2026-04-08
[MRG] ACS: GUI accessibility improvements

asoplata:
This is related to #1261 , where the HNN GUI itself had numerous WAVE accessibility errors. Fortunately, Claude was able to figure out how to fix all of the errors (except for contrast), as far as I can tell. I also got help from a StackOverflow post for this.

@xgao35 I'm out next week, but at some point before May 11, we should both sit down and double check that this (and other web content elsewhere!) satisfies at least having no WAVE errors. I also have some questions about what some of the non-error warnings mean and/or what else we need to fix (if anything).

**edit: I'll add that the GUI code itself is *very* difficult to understand and I wouldn't bother. However, it would be helpful if you checked that this branch's GUI webpage passes the WAVE plugin for you, and take note of anything else related to accessibility that you can see on it that we need to address.

### DIFF
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -34,8 +34,11 @@ _fig_placeholder = HTML(
     value="""
         Visualizations will become available in this window after a simulation has been
         run or data have been loaded
-    """
-).add_class("fig-placeholder")
+    """,
+    description="Visualization placeholder",
+)
+_fig_placeholder.add_class("fig-placeholder")
+_fig_placeholder.add_class("hide-label")
 
 _plot_types = [
     "current dipole",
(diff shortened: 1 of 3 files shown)


xgao35:
The remaining errors I see now are contrast errors and orphaned form labels. They should both be quick fixes when we sit down and go through this together!

asoplata:
TODOs:

- Fix contrast issues
- check contrast of bottom tab under all sim conditions

asoplata:
@xgao35 I think this resolves the remaining "real" contrast issues with the GUI on both Chrome and Firefox. The Chrome WAVE plugin still mentions that there are 2 remaining contrast issues, but it doesn't show where they are, and I can't find anything displayed on any tab that resembles the colors it describes. Therefore, I think those last errors are only for hidden widgets that are not displayed somewhere.

I think this is ready for merge, and I will plan to merge this by the end of the week unless you find a different issue that needs to be resolved on it.

asoplata:
This was reviewed and approved at Code Review by others on 2026-04-30

PR #1281, opened 2026-04-10
Remove celltype hardcoding from visualization and response modules

### LINKED ISSUE #1149: REF: Remove celltype hardcoding from viz.py

vshnvii:
fixes #1149

Previously, both `viz.py` and `cell_response.py` relied on fixed arrays like `["L2_basket", "L2_pyramidal", "L5_basket", "L5_pyramidal"]` for mapping colors, markers, and ordering. This made the code less flexible and caused issues when working with sub-sampled networks or custom cell types.

The goal here is to make visualization fully driven by metadata, so it becomes easier to extend and maintain.

---

#### 1. Global cell metadata

* Moved the default `cell_metadata` from inside `jones_2009_model()` to a module-level variable `default_cell_metadata` in `hnn_core/network_models.py`.
* Added `color` and `marker` fields directly into this metadata.
* Kept the existing color scheme (`m`, `c`, `r`, `b`) and markers (`^`, `x`) so current visual behavior remains unchanged.

---

#### 2. Updates in `viz.py`

* **`plot_cells()`**

  * Removed hardcoded color and marker dictionaries.
  * Now reads values dynamically from `cell_metadata`.

* **`plot_spikes_raster()`**

  * Removed fixed cell type ordering.
  * Uses `unique_spike_types` from the data.
  * Fetches colors from `default_cell_metadata`.
  * Falls back to matplotlib’s default color cycle if a cell type is not registered.

* **`plot_spikes_hist()`**

  * Removed hardcoded cell type arrays.
  * Now derives them dynamically from available metadata.

---

#### 3. Changes in `cell_response.py`

* Replaced hardcoded `network_cell_names` with dynamic retrieval from `default_cell_metadata.keys()`.
* This ensures consistency with the rest of the codebase and avoids duplication.

---

### DIFF
--- a/hnn_core/cell_response.py
+++ b/hnn_core/cell_response.py
@@ -563,7 +563,8 @@ def read_spikes(fname, gid_ranges=None):
             spike_gids += [list()]
             spike_types += [list()]
 
-    network_cell_names = ["L2_basket", "L2_pyramidal", "L5_basket", "L5_pyramidal"]
+    from .network_models import default_cell_metadata
+    network_cell_names = list(default_cell_metadata.keys())
     cell_type_names = list(
         cell_name for cell_name in network_cell_names if cell_name in spike_types
     )
--- a/hnn_core/network_models.py
+++ b/hnn_core/network_models.py
@@ -10,6 +10,45 @@ from .params import _short_name
 from .cells_default import pyramidal_ca, pyramidal, basket
 from .externals.mne import _validate_type
 
+default_cell_metadata = {
+    "L2_basket": {
+        "morpho_type": "basket",
+        "electro_type": "inhibitory",
+        "layer": "2",
+        "measure_dipole": False,
+        "reference": "https://doi.org/10.7554/eLife.51214",
+        "color": "m",
+        "marker": "x",
+    },
+    "L2_pyramidal": {
+        "morpho_type": "pyramidal",
+        "electro_type": "excitatory",
+        "layer": "2",
+        "measure_dipole": True,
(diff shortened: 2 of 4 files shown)


vshnvii:
Hi @asoplata ,
Opened this PR to address #1149, feedback welcome whenever you get a chance.

asoplata:
Hello Vaishnavi, thanks for your contributions but unfortunately this is a duplicate of work that was already begun in #1269, which we forgot to connect to the original issue of #1149. We had previously already done some reviewing of #1269 so we will be moving forward with that PR instead of this one, and I will close this PR. Apologies for the miscommunication!

Issue #1118, opened 2025-09-16
REF/TEST: Large test refactors


asoplata:
There are some large-scale ways that the testing suite could be significantly improved:

- `Network` object setup: Many test files use their own custom ways of setting up the `Network` under test, including before adding any drives, and those ways are not always used consistently in the same file. This has real consequences, in that this also means that certain Network types are *tested inconsistently*, which also means that certain tests are not tested against all of our common "test Networks", causing bugs to slip through the cracks; see #960 for an example. Ideally, what this means is that our existing fixture in `conftest.py` ( https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/tests/conftest.py#L73 ) should be greatly expanded to add options for other kinds of `Network`s, including those with no drives, those of different sizes, etc. Then, most tests (where applicable) should consume those fixtures and run on all compatible Network fixtures.

- Very long tests need to be considered for splitting up: `test_drive.py::test_add_drives` ( https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/tests/test_drives.py#L220 ) is a good example of this: individual drives (and their arguments) should be tested independently from one another, in addition to combinations of them being tested. Separating tests into smaller pieces helps with debugging and is just as important as "integration testing" where we want to see if our code pieces play well together.

- Once we get Codecov working #1112 , or maybe just with `pytest`, we should use that to track down the remaining un-tested code coverage, then add tests.

- IMHO, there are many cases where most of a function's arguments are covered by tests, but not all.

PR #1283, opened 2026-04-14
[WIP] Convert bare pytest.raises calls to context managers in test_extracellular.py

SatvikSaluja:
Refactors functional-style `pytest.raises(ExcType, callable, args)` calls in
`test_extracellular_api` to use the `with pytest.raises(...):` context manager,
bringing them in line with the rest of the test suite.

Additionally, selected cases now include `match=` to assert exception messages,
ensuring tests fail if an incorrect `ValueError` is raised rather than passing
silently.

Part of the test quality improvements tracked in #1118.

No functional changes. All tests pass locally.

### DIFF
--- a/hnn_core/tests/test_extracellular.py
+++ b/hnn_core/tests/test_extracellular.py
@@ -40,7 +40,8 @@ def test_extracellular_api():
     assert (net.rec_arrays["arr1"] == "extArr") is False
 
     # ensure unique names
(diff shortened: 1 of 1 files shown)


SatvikSaluja:
Note.
If this PR is merged , it should be done before  #1284

SatvikSaluja:
Rebased and will open a cleaner PR based on latest upstream.

PR #1284, opened 2026-04-14
( WIP )Fix test params isolation

SatvikSaluja:
While making PR #1283 , I noticed that a module-level `params` dictionary
was being shared across multiple test functions and modified in-place via
`params.update`.

 a `deepcopy(params)` workaround was introduced earlier
(commit 605eb68) to avoid mutation issues in one specific test. However,
other tests continued to operate on the shared dictionary directly, which
means the underlying problem of shared mutable state still remained.

To address this more consistently, each test now creates its own fresh
`params` instance using `read_params(params_fname)`. This removes the need
for `deepcopy` and ensures that tests are fully isolated from each other.

No functional changes. All tests pass locally.

### DIFF
--- a/hnn_core/tests/test_extracellular.py
+++ b/hnn_core/tests/test_extracellular.py
@@ -1,7 +1,6 @@
 # Authors: Nick Tolley <nicholas_tolley@brown.edu>
 #          Christopher Bailey <cjb@cfin.au.dk>
 
-from copy import deepcopy
 import os.path as op
 import numpy as np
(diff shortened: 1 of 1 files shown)


asoplata:
Thanks Satvik! We will try to get to your PR as soon as we can; we're currently extra busy on account of many of us going to a conference this week, and we have been preparing for a new release for the past several weeks (out very soon). Thank you for your patience, and for your contribution!

SatvikSaluja:
Rebased and will open a cleaner PR based on latest upstream.

PR #1287, opened 2026-04-18
add macos-15-intel runner for Python 3.9 tests

### LINKED ISSUE #1134: TEST: Add Intel “macos-15-intel” to some tests
~~I noticed from https://cibuildwheel.pypa.io/en/stable/setup/#github-actions when we were talking with George, the config shown has the following comment:~~

```
# macos-13 is an intel runner, macos-14 is apple silicon .
```

~~For at least the next few years, we may want to consider not just testing macos-latest in our Github Actions, but also macos-13 since it is a different architecture, and our primary users (researchers) tend to not upgrade their laptop hardware frequently.~~

UPDATE: **We should add `macos-15-intel` to our Github Actions runner images used.** For this image, we probably ONLY want to run Python version 3.9 (or whatever is the lowest version we support), instead of the entire Python version matrix in `.github/workflows/unix_unit_tests.yml`.

 `macos-13` itself has reach end-of-life (no security updates), so Github have provided a new `macos-15-intel` runner image in order to run tests on the pre-Apple-Silicon Intel chips, as of https://github.com/actions/runner-images/issues/13045 . This image will only be available until August 2027, after which all MacOS Intel chip images will be dropped.

vshnvii:
fixes #1134

### Changes made

* Updated `.github/workflows/unix_unit_tests.yml` to include `macos-15-intel`
* Restricted the Intel runner to Python 3.9 only using `include:` instead of expanding the full matrix
* Updated macOS-specific `DYLD_FALLBACK_LIBRARY_PATH` handling from a hardcoded `macos-latest` check to pattern matching with `macos*`, so both Apple Silicon and Intel runners are supported correctly

### DIFF
--- a/.github/workflows/unix_unit_tests.yml
+++ b/.github/workflows/unix_unit_tests.yml
@@ -15,8 +15,11 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest,macos-latest]
+        os: [ubuntu-latest, macos-latest]
         python-version: [3.9, 3.11, 3.13]
+        include:
+          - os: macos-15-intel
+            python-version: 3.9
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


asoplata:
Seems to look and work great, thanks a bunch Vaishnavi! I'll add recognition of your contribution and then merge. Apologies for not getting to your other PRs, we are currently swamped with trying to get a new release out, but we will review them soon.

vshnvii:
Thank you so much! No worries at all, I completely understand. Glad this contribution was helpful, and I really appreciate the recognition. Looking forward to contributing more!

PR #1290, opened 2026-04-27
[MRG] MAINT: rename jones2009 to neymotin2020

asoplata:
This is a PR for renaming the ubiquitous `jones_2009_model` function to `neymotin_2020_model`, along with similar changes to downstream products of this function (such as file `jones2009_base.json -> neymotin2020_base.json`). Seems to be passing all tests on my machine. 

I'm currently doing a website/documentation build as well, to test that this change doesn't break anything there either (similar to the CircleCI job for this PR).

This is probably good to go, other than having a second pair of eyes to make sure I'm not forgetting to change anything. I used `ripgrep` to find all uses of the function, so the code should already be completely converted.


asoplata:
Found a bug, do not merge

asoplata:
Fixed bug, probably ready for merge now

dylansdaniels:
LGTM

PR #1293, opened 2026-04-28
[MRG] ACS fix accessibility errors on dev portal

asoplata:
This fixes the missing alt-image text on the "Dev portal" https://jonescompneurolab.github.io/hnn-core/stable/index.html frontpage, and also removes the contrast errors from any displayed code blocks.

We don't need to provide alt-image text for any generated Example pages because those are going to be migrated to the Textbook very soon.

@xgao35 Actually generating a local version of the website for this is a little complicated, but I can show a version I've built tomorrow


asoplata:
TODOs:

- [x] remove contrast-unfriendly highlighting

asoplata:
- [x] TODO: Use WAVE extension to inspect the built HTML artifacts after CircleCI completes in this PR

@xgao35 I think this is the last step needed for the "Dev portal" to be fixed

PR #1295, opened 2026-05-01
[WIP] FEAT Request Add ability to control random seeding

### LINKED ISSUE #1294: [FEAT] Request: Add ability to control random seeding in Bayesian optimization
Currently there is no way to fix/control the random seed during Bayesian optimization. The `bayes_opt` function actually has a parameter to control this:
https://github.com/jonescompneurolab/hnn-core/blob/dd676043bf39cf8c089bb5cb72f9316d838096d7/hnn_core/optimization/general_optimization.py#L445

So exposing this would be a matter of passing it to `optim.fit(..., seed=123)` where the seed is included as an entry in `obj_fun_kwargs`, and then internally gets passed to `bayes_opt()`.

mangodxd:
Added a missing parameter to the function signature.

## Changes
- Added parameter 'eter' to function 'bayes_opt' in hnn_core/externals/bayesopt.py
- Files changed:
- `hnn_core/externals/bayesopt.py`

## Testing
Verify the new parameter is accepted and defaults work correctly.

Fixes #1294

### DIFF
--- a/ai_fix.patch
+++ b/ai_fix.patch
@@ -1,0 +1,13 @@
+diff --git a/hnn_core/externals/bayesopt.py b/hnn_core/externals/bayesopt.py
+index d5cd898..8fb5fe5 100644
+--- a/hnn_core/externals/bayesopt.py
++++ b/hnn_core/externals/bayesopt.py
+@@ -45,7 +45,7 @@ def expected_improvement(gp, best_f, all_x):
+     return (y - best_f) * st.norm.cdf(Z) + y_std * st.norm.pdf(Z)
+ 
+ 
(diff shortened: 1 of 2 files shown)


asoplata:
Hello Nguyễn, thank you for your contribution. However, I think some code is missing from your PR. You've added a new argument to the right function, but that argument does not appear to actually be used anywhere, and thus will not be able to control the random seeding. Can you take a second look? 

Also, you've also included an `ai_fix.patch` file, but there is no reason to include this since we want our PRs to include the changes applied directly to the file. Github automatically allows other developers to get `.patch` files automatically from PRs, so there is no need to include that.

PR #1296, opened 2026-05-04
[WIP] FEAT Add optional 2D cell morphology overlay to plot_csd

carolinafernandezp:
This allows users to overlay an optional 2D projection of the cell morphologies on top of CSD plots (it preserves depth information to align with electrode contacts). The goal is to give anatomical context to the CSD plots making it easier to interpret how the cell morphologies align with current sinks and sources.

**A quick note:** I'm essentially drawing the cell sections as lines because I don't think diameter matters too much for the purposes of aligning the cell morphology to the CSD pattern, but I could plot them with the actual diameters using `_linewidth_from_data_units` if that's preferred.

This figure shows the standard use with the default color:
`net.rec_arrays['shank1'][trial_idx].smooth(window_len=window_len).plot_csd(ax=ax, overlay_cell_morphology=True, net=net)`
overlay_cell_morphology

And this one shows how different colors can be assigned to different cell sections:
`net.rec_arrays['shank1'][trial_idx].smooth(window_len=window_len).plot_csd(ax=ax, overlay_cell_morphology=True, net=net, morphology_color=color_dict)`
overlay_cell_morphology_color_dict

### DIFF
--- a/hnn_core/extracellular.py
+++ b/hnn_core/extracellular.py
@@ -578,6 +578,9 @@ class ExtracellularArray:
         sink="b",
         colorbar=True,
         ax=None,
+        overlay_cell_morphology=False,
+        net=None,
+        morphology_color=None,
         show=True,
     ):
         """Plot laminar current source density (CSD) estimation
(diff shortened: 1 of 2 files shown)


asoplata:
Very cool, thanks Carolina! I was just discussing something similar with @annacatt about this same plotting.

When this code is ready for review and you're done developing it, please let us know with a comment saying so.

asoplata:
Thanks Carolina - after discussing this with @annacatt , it sounds like y'all agreed that Anna would take over and work on finishing her work on a similar implementation. For that, she will be opening a new PR at some point in the future, so I will close this PR unless anyone has a problem with that.

PR #1190, opened 2025-11-17
[MRG]: Feat: Optimization in GUI

asoplata:
This is a *heavily* in-development version of adding a new Optimization tab in the GUI. There's tons of debug stuff around, not to mention TODO comments, etc. However, a minimal working example seems to work correctly. The most difficult part, using a closure to dynamically build a `params` dict and a `set_params` function, seem to work.

todo list:
1. [x] Revert changes made in d2a3f40 to the `weights_ampa` etc. synaptic data structures, since those need to stay as-is. Constraints should be added in parallel to new data structures that mimic them, instead of being added directly. This is needed because the `add_<drive>` functions expect a certain dictionary structure, and we should handle and deliver that cleanly.
2. [x] Test that synaptic weight constraints actually work, since they're a bit more complex than single-variable constraints.
3. [x] Test that drive-unique checkbox selection is working correctly. In prior testing, un-checking the checkbox of the second drive did not seem to disable that parameter from being included in the constraints of the actual optimization. This can be tested easily (probably) with a breakpoint in Optimizer's initializer.
4. [x] Add observed-updating of the ghosted drive variable values, where the ghosted instance of the value in the Optimization tab is updated when the version in the Drive tab is (this might be a pain). Alternatively, we could make them both editable and have them observe each other between the tabs (this would be even more of a pain!).
5. [x] There are probably other TODOs in the code comments that I should do at this point.
6. [x] Investigate if there dependencies between `solver` choices and `obj_fun` choices, then add those to GUI if necessary.
7. [x] Rebase onto the latest `master` which contains Dylan's visual GUI refresh
8. [x] Inspect after rebase
9. [x] Fix save opt history no longer working
10. [x] Re-add Nick's two remaining comments from https://github.com/asoplata/hnn-core/pull/2 here.
11. [x] Begin adding tests, limiting at first to evoked drives. Use a class or something for the setup for this (which will be a fair bit of work), then adapt the existing evoked tests to this situation.
12. [x] Add `dipole_corr` support
13. [x] Add in the new CMA solver support
14. [x] Once I'm confident that the current implementation works for the evoked drive case, then use the same paradigm for all other drive types (boring). Then add similar tests for them.
15. [x] Finally, consult with others and polish the UI.
16. [x] After CMA, add additional target widgets for sigma and pop_size
17. [x] Remove mutable defaults

PR #1298, opened 2026-05-08
[MRG] GUI Print Corr on GUI data comparison plot

asoplata:
@ntolley This is a redo of https://github.com/ntolley/hnn-core/pull/20 which, when a user is plotting both loaded and simulated data in the GUI, the GUI will display not only the RMSE comparison value but also the correlation value.

This should be merged before the release because IIRC this display of the corr in the plot is featured in the JOVE paper.

### DIFF
--- a/hnn_core/gui/_viz_manager.py
+++ b/hnn_core/gui/_viz_manager.py
@@ -25,7 +25,7 @@ from ipywidgets import (
     link,
 )
 
-from hnn_core.dipole import _rmse, average_dipoles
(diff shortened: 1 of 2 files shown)


asoplata:
Resolves #1278

ntolley:
@asoplata would it be better to merge this first? or wait until #1190 is merged?

asoplata:
Better to merge after

asoplata:
Self-merging (after tests pass) after in-person discussion with Nick on 20260512

PR #1304, opened 2026-05-13
[MRG] TEST add progressive installation tests

asoplata:
This adds a few new steps in both the Unix and Windows CI runners between the minimal install and the complete install with all packages. Now, the CI runners do the following in order:

1. test that the minimal install (API-only) works with a test simulation
2. test that the `[opt]` dependencies are sufficient to allow import of `Optimizer`
3. test that `[gui]` depdencies are sufficient to allow proper startup of the gui
4. continue with the remaining tests, where a complete install (except for `[doc]`) is installed, then various checkers are run, then finally the tests as a whole are run.

### DIFF
--- a/.github/workflows/unix_unit_tests.yml
+++ b/.github/workflows/unix_unit_tests.yml
@@ -75,7 +75,7 @@ jobs:
         shell: bash -el {0}
         run: rm -rf build
 
-      - name: Install HNN-Core with Optimization dependencies
+      - name: Install HNN-Core with GUI dependencies
         shell: bash -el {0}
         run: pip install --verbose '.[gui]'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/unix_unit_tests.yml
```diff
@@ -29,29 +29,94 @@ jobs:
...
+
+      - name: Test Optimization dependencies
+        shell: bash -el {0}
+        run: python -c "from hnn_core.optimization import Optimizer"
+
+      - name: Clean Optimization dependencies install
+        shell: bash -el {0}
+        run: rm -rf build
+
+      - name: Install HNN-Core with Optimization dependencies
```
dylansdaniels:
```suggestion
      - name: Install HNN-Core with GUI dependencies
```

PR #1307, opened 2026-05-15
[WIP] Enable probability for tonic bias

jtajchman:
Allows tonic bias to hit a probabilistic proportion of cells of each cell type

### DIFF
--- a/hnn_core/network.py
+++ b/hnn_core/network.py
@@ -1624,18 +1624,13 @@ class Network:
                 bias_seed=bias_seed,
             )
         else:
+            _validate_type(amplitude, dict, "amplitude")
             _validate_type(probability, (float, dict), "probability")
-            if isinstance(amplitude, float):
-                # single amplitude for all cell types
-                amplitude = {
-                    cell_type: float(amplitude) for cell_type in self.cell_types
-                }
             if isinstance(probability, float):
                 # single probability for all cell types
                 probability = {
-                    cell_type: float(probability) for cell_type in self.cell_types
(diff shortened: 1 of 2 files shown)


katduecker:
Happy to look at this when I get back from traveling! I have also implemented something like this for the new model, where I allow a bias to be defined for each cell (based on GID). TBH, I feel like being able to define GIDs for cells that are activated as an input allows users more control than randomly applying the bias, but your functionality is related to this. We should talk about this at a dev meeting...

jtajchman:
> Happy to look at this when I get back from traveling! I have also implemented something like this for the new model, where I allow a bias to be defined for each cell (based on GID). TBH, I feel like being able to define GIDs for cells that are activated as an input allows users more control than randomly applying the bias, but your functionality is related to this. We should talk about this at a dev meeting...

Yeah that sounds like a more generalized and powerful approach. Should we close this PR?

katduecker:
Nooo I think it’s great that you opened this and you should get credit for your contribution! I think we should talk about this at a dev meeting in early June :)

asoplata:
Changing to WIP since it's unclear how this will be merged alongside similar changes in #1227, which we'll hash out when Katharina gets back

asoplata:
@jtajchman We've merged Katharina's branch into `master`, and part of that was a big refactor of `add_tonic_bias`. The bad news is that if you want to add `probability` as an argument, this PR needs to be updated against `master`; the good news is that it should be easier to work with `add_tonic_bias`.

If you don't have time to do this yourself, then let me know and I can help resolve the discrepancies!

PR #1247, opened 2026-03-03
MAINT: Replace os.path with pathlib in test suite

SatvikSaluja:
This PR replaces os.path usage with pathlib.Path within the tests section of the codebase, with only minimal string conversions where necessary.

Changes include:
Replacing op.join, op.dirname, and op.exists with their Path equivalents

Removing import os.path as op where it is no longer needed

Laying groundwork that could be extended to the rest of the codebase

  All test cases passed locally
  tests pass: 225

PR #1311, opened 2026-05-22
[MRG] MIGRATE os path utilities to pathlib

SatvikSaluja:
Builds on #1247 and fixes a small subproblem of #1118 by replacing relevant os path utilities with pathlib equivalents.

For reference on the os.path → pathlib equivalents: [pathlib correspondence to os module tools](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module)

### DIFF
--- a/hnn_core/tests/test_batch_simulate.py
+++ b/hnn_core/tests/test_batch_simulate.py
@@ -206,7 +206,7 @@ def test_save_load_and_overwrite(batch_simulate_instance, param_grid, tmp_path):
 
     batch_simulate_instance._save(results, start_idx, end_idx)
 
-    file_name = tmp_path/ f"sim_run_{start_idx}-{end_idx}.npz"
+    file_name = tmp_path / f"sim_run_{start_idx}-{end_idx}.npz"
     assert file_name.exists()
 
     loaded_data = np.load(file_name, allow_pickle=True)
@@ -257,7 +257,7 @@ def test_load_results(batch_simulate_instance, param_grid, tmp_path):
     end_idx = len(results)
     batch_simulate_instance._save(results, start_idx, end_idx)
 
-    file_name = tmp_path/ f"sim_run_{start_idx}-{end_idx}.npz"
+    file_name = tmp_path / f"sim_run_{start_idx}-{end_idx}.npz"
     assert file_name.exists()
 
     # single result file
--- a/hnn_core/tests/test_batch_simulate.py
+++ b/hnn_core/tests/test_batch_simulate.py
@@ -206,10 +206,10 @@ def test_save_load_and_overwrite(batch_simulate_instance, param_grid, tmp_path):
 
     batch_simulate_instance._save(results, start_idx, end_idx)
 
-    file_name = tmp_path / f"sim_run_{start_idx}-{end_idx}.npz"
-    assert file_name.exists()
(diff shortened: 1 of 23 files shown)


SatvikSaluja:
Note: `os.environ` and `os.walk` were intentionally left unchanged. i knew about the os.enviorn thing ,there isnt any equivalent in pathlib . 

Thanks to **Claude** for information that os.walk remains more suitable here due to dirnames control.

hnn_core/tests/test_dipole.py
```diff
@@ -320,16 +319,16 @@ def test_rmse():
...
         urlretrieve(data_url, "yes_trial_S1_ERP_all_avg.txt")
     extdata = np.loadtxt("yes_trial_S1_ERP_all_avg.txt")
 
     exp_dpl = Dipole(
         times=extdata[:, 0], data=np.c_[extdata[:, 1], extdata[:, 1], extdata[:, 1]]
     )
 
-    hnn_core_root = op.join(op.dirname(hnn_core.__file__))
-    params_fname = op.join(hnn_core_root, "param", "default.json")
+    hnn_core_root = Path(hnn_core.__file__).resolve().parent
```
ntolley: why is `resolve()` necessary here and not above at L189?

hnn_core/tests/test_batch_simulate.py
```diff
@@ -207,8 +206,8 @@ def test_save_load_and_overwrite(batch_simulate_instance, param_grid, tmp_path):
 
     batch_simulate_instance._save(results, start_idx, end_idx)
 
-    file_name = os.path.join(tmp_path, f"sim_run_{start_idx}-{end_idx}.npz")
-    assert os.path.exists(file_name)
+    file_name = tmp_path / f"sim_run_{start_idx}-{end_idx}.npz"
```
asoplata: REFACTOR: In general, when switching to using `Path` objects instead of strings which are filenames, I recommend changing the variables accordingly. For instance, `file_name` here is a `Path` object that contains the filename, but it is NOT simply string that is the filename.

There are some cases where you want to manage `file_name` SEPARATELY from the Path that contains that filename (such as if you're doing fancy stuff with the parent directories, but not the individual filename).

asoplata:
There's some leftover `os.path` usage also in `doc/conf.py` and `setup.py`, please change these

SatvikSaluja:
@asoplata . Hey , should i include the examples script for migrating to pathlib ?

asoplata:
Hello Satvik, don't worry about the examples, but can you please update `doc/conf.py` as well? Thanks

SatvikSaluja:
Hi @asoplata 

all os.path has been migrated to pathlib 

This final commit from my side failed -> https://github.com/jonescompneurolab/hnn-core/pull/1311/changes/578359fd428c4c50704bd50bf3e02e8c1d2669b1

With my limited knowledge in scripting and with some help from Claude . i did the following 2 commits . However that didn't fix it 

Then i thought you might have better idea than me -> so i removed the changes i added in yaml script. 

However now its running and passing all checks

Can u look into it ?
Thanks

asoplata:
Thanks Satvik!

SatvikSaluja:
@asoplata after merging the code has failed some lint checks

Issue #999, opened 2025-02-21
BUG: Recording cellular data from MPI simulations is significantly slower than non-MPI


asoplata:
If one takes the code in the “Plot Firing Patterns” example, https://jonescompneurolab.github.io/hnn-core/stable/auto_examples/howto/plot_firing_pattern.html#sphx-glr-auto-examples-howto-plot-firing-pattern-py , the the time to record either the `vsec` or, if you slightly modify the code, `isec` data is effectively instant.

However, if one replaces the run-simulation line, 

```python
dpls = simulate_dipole(net, tstop=170., record_vsec='soma')
```

with the same thing except for using MPI:

```python
with MPIBackend(n_procs=2):
    dpls = simulate_dipole(net, tstop=170., record_vsec='soma')
```

then the time to record takes *significantly* longer (about a minute on my machine).

Furthermore, if one tries to record the `isec` while running the sim using MPI, like with

```python
with MPIBackend(n_procs=2):
    dpls = simulate_dipole(net, tstop=170., record_isec='soma')
```

then it takes so long that I eventually get an “MPI Simulation failed” error, and it fails to record the data. In this case, the reasons for the failure may be that 1. it takes a long time, and 2. we have a hard-coded timeout on MPI subprocesses. I believe this is the case since, if one modifies the `run_subprocess` timeout in `hnn_core/parallel_backends.py::MPIBackend.simulate`, then the data does get successfully get recorded, but it takes multiple minutes to do so (longer than in the `vsec` case).

I've marked this as a bug since this implies that in certain cases (e.g. the MPI example above when recording `isec`), it is not effectively possible for a user to use both MPI and to record some cellular properties, particularly for simulations that are of any substantial length. It is also not obvious to the user that stopping use of MPI would solve their problem, which it will in this case.

I will also note that the simulation time does not seem to vary with the recording: the issue clearly occurs only after a simulation has run, where the data is probably being aggregated.

asoplata:
Note: in @kmilo9999 's testing, the following script does NOT produce a timeout error when run using Ubuntu WSL on windows, but it DOES produce a timeout error on my local Mac laptop (MacOS Tahoe 26.4.1 (25E253)  :

```python
from hnn_core import neymotin_2020_model, MPIBackend, simulate_dipole

net = neymotin_2020_model()

weights_ampa_p = {"L2_pyramidal": 1.0, "L5_pyramidal": 1.0}
[...]

PR #1312, opened 2026-05-27
[BUG] Fix MPI timeout

kmilo9999:
Solves #999 
`MPIBackend` would hang indefinitely when using `record_vsec='all'`, `record_isec='all'`, or `record_ca='all'`. The simulation itself completed successfully, but results were never returned to the parent process.

The root cause was a pipe deadlock in the original data transport: rank 0 wrote the entire base64-encoded result as a single line to stderr, and the parent thread blocked on `readline()` waiting for the terminating `\n`. The OS pipe buffer (~64 KB on Linux) filled up after the first 64 KB of a more than 600 MB data stream, causing both sides to block indefinitely. The problem is worst with higher `n_procs`  because rank 0 gathers recordings from all ranks via before writing, producing a larger payload each time.

## New implementation

`Rank 0 ` now writes the pickled result directly to a temporary file and signals the parent with a single short line on stderr containing the file path and byte count:

`@data_file:/tmp/hnn_mpi_data_XXXX.pkl:468000000@`

The parent reads this line (It's < 200 bytes, no pipe hang risk), opens the file, unpickles the data, and deletes the temp file.


asoplata:
- [ ] TODO for reviewers @dylansdaniels @ntolley : make sure that the tempfile data is actually what we expect

asoplata:
https://github.com/brown-ccv/hnn-core-ccv/pull/5

hnn_core/tests/test_mpi_child.py
```diff
@@ -77,6 +77,8 @@ def test_extract_data_length():
 
 def test_str_to_net():
     """Test reading the network via a string"""
+    import pickle
```
ntolley: Why not import at top level?

hnn_core/tests/test_mpi_child.py
ntolley, line 68: Maybe a dumb question, what do the numbers a the of the string mean?

ntolley: nevermind! the next test answered my question

hnn_core/tests/test_mpi_child.py
```diff
@@ -139,16 +141,15 @@ def test_child_run():
         with io.StringIO() as buf_err, redirect_stderr(buf_err):
             mpi_sim._write_data_stderr(sim_data)
             stderr_str = buf_err.getvalue()
-        assert "@start_of_data@" in stderr_str
```
ntolley, line 142: Why do these strings go away? is it because the strings are now broken into chunks so it's not guaranteed it will exist in the current chunk?

hnn_core/tests/test_mpi_child.py
```diff
@@ -139,16 +141,15 @@ def test_child_run():
...
+        assert "@data_file:" in stderr_str
 
         # write data to queue
         err_q = Queue()
         err_q.put(stderr_str)
 
         # use _read_stderr to get data_len (but not the data this time)
-        data_len, data = _get_data_from_child_err(err_q)
-        sim_data = _process_child_data(data, data_len)
+        data_len, path_data_file = _get_data_from_child_err(err_q)
```
ntolley: Maybe the returned arguments of `_get_data_from_child_err()` should be flipped to match `_process_child_data()`

hnn_core/tests/test_mpi_child.py
```diff
@@ -157,20 +158,27 @@ def test_child_run():
...
+    data_path = None
+    with pytest.raises(
+        RuntimeError, match="MPI simulation didn't produce a result file"
+    ):
+        _process_child_data(data_path, 0)
 
 
 def test_data_len_mismatch():
     """Test that padded data can be unpickled with warning for length"""
+    import tempfile
```
ntolley: same comment as above, let's move these imports to the top

hnn_core/parallel_backends.py
```diff
@@ -112,13 +112,19 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
...
     """
-    proc_data_bytes = b""
+
     # each loop while waiting will involve two Queue.get() timeouts, each
     # 0.01s. This calculation will error on the side of a longer timeout
     # than is specified because more is done each loop that just Queue.get()
     timeout_cycles = timeout / 0.02
 
-    pickled_obj = base64.b64encode(pickle.dumps(obj))
+    ## Maybe leaving this print for debugging purposes
```
asoplata, line 125: TODO: add a debugging flag for whether to write this to stdout

hnn_core/mpi_child.py
```diff
@@ -107,19 +127,26 @@ def _wait_for_exit_signal(self):
 
     def _write_data_stderr(self, sim_data):
         """write base64 encoded data to stderr"""
+        """writes data to a temp file """
```
ntolley:
```suggestion
"""write base64 encoded data to stderr
     writes data to a temp file"""
```

hnn_core/parallel_backends.py
```diff
@@ -146,17 +154,30 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
...
+
         # loop while the process is running the simulation
         while True:
             child_terminated = proc.poll() is not None
 
             if not data_received:
-                if _echo_child_output(out_q):
+                child_out = _echo_child_output(out_q)
+                if child_out:
+                    if not first_output_received:
```
asoplata, line 205: TODO again add print only if debug flag passed

hnn_core/mpi_child.py
```diff
@@ -107,19 +127,26 @@ def _wait_for_exit_signal(self):
...
-        # be certain it is the border of the signal
-        sys.stderr.write("@end_of_data:%d@\n" % len(pickled_bytes))
+        # rankm 0 writes data to temp file
+        pickled_bytes = pickle.dumps(sim_data)
+        fd, tmp_path = tempfile.mkstemp(prefix="hnn_mpi_data_", suffix=".pkl")
+        try:
+            with os.fdopen(fd, "wb") as f:
+                f.write(pickled_bytes)
+        except Exception:
+            # This exception was necessary in Oscar to make the runtime error more verbose
```
ntolley:
```suggestion
            # This exception is necessary in HPC environments to make the runtime error more verbose
```
Might as well make the comment generally applicable

hnn_core/parallel_backends.py
```diff
@@ -189,15 +210,20 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
             if child_terminated and data_received:
                 # both exit conditions have been met (also we know that
                 # the network has been sent)
+                print(f"process terminated after {time.time() - t_start:.2f}s")
```
asoplata: TODO include in debug

hnn_core/tests/test_mpi_child.py
ntolley: Nevermind I had the original solution in mind where the data was broken up...

hnn_core/parallel_backends.py
```diff
@@ -189,15 +210,20 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
...
                 break
 
             if not child_terminated and count_since_last_output > timeout_cycles:
-                warn(
-                    "Timeout exceeded while waiting for child process output"
-                    ". Terminating..."
-                )
-                kill_proc_name("nrniv")
-                break
+                if sent_network:
```
asoplata, line 254: Should we still include some kind of timeout, just in case that something goes wrong and the simulation hangs?

Another TODO: we should add in here a Warning like "Simulation has taken XXX minutes so far, still waiting to hear back results. If this is taking much longer than expected, there may be a problem with the simulation"

hnn_core/parallel_backends.py
```diff
@@ -245,45 +271,44 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
...
+        raise RuntimeError("MPI simulation didn't produce a result file. ")
+
+    try:
+        with open(path_data_file, "rb") as f:
+            data_bytes = f.read()
+    finally:
+        try:
+            os.unlink(path_data_file)
+        except OSError:
+            pass
```
ntolley, line 365: Why pass if there's an OSError? Wouldn't we want this to fail?

hnn_core/mpi_child.py
```diff
@@ -107,19 +127,26 @@ def _wait_for_exit_signal(self):
...
-        pickled_bytes = _pickle_data(sim_data)
-        sys.stderr.write(pickled_bytes.decode())
-
-        # the parent process is waiting for "@end_of_data:[#bytes]@" with the
-        # length of data. The '@' is not found in base64 encoding, so we can
-        # be certain it is the border of the signal
-        sys.stderr.write("@end_of_data:%d@\n" % len(pickled_bytes))
+        # rankm 0 writes data to temp file
+        pickled_bytes = pickle.dumps(sim_data)
+        fd, tmp_path = tempfile.mkstemp(prefix="hnn_mpi_data_", suffix=".pkl")
```
asoplata, line 129: TODO add two log msgs here before and after, one where "Beginning to write temp file to $LOCATION" and then later "successfully written temp file to location"

hnn_core/parallel_backends.py
```diff
@@ -245,45 +271,44 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
...
+    if path_data_file is None or data_len == 0:
         # This is indicative of a failure. For debugging purposes.
+        raise RuntimeError("MPI simulation didn't produce a result file. ")
+
+    try:
+        with open(path_data_file, "rb") as f:
+            data_bytes = f.read()
+    finally:
+        try:
+            os.unlink(path_data_file)
```
asoplata, line 363: Is this basically the same as `rm <file>` on the command line?

hnn_core/parallel_backends.py
```diff
@@ -292,39 +317,37 @@ def _echo_child_output(out_q):
...
             break
 
+    # Return the output instead of just a boolean value so callers can
+    # tell if anything was received and check the actual content
+    # (logging the first line of nrniv output).
     if len(out) > 0:
         sys.stdout.write(out)
-        return True
-    return False
+        return out
```
ntolley, line 390: Would this increase the RAM usage?

hnn_core/parallel_backends.py
```diff
@@ -292,39 +317,37 @@ def _echo_child_output(out_q):
         except Empty:
             break
 
+    # Return the output instead of just a boolean value so callers can
```
asoplata, line 385: I don't really understand this but i probably just need to spend more time reading

hnn_core/parallel_backends.py
```diff
@@ -292,39 +317,37 @@ def _echo_child_output(out_q):
...
-    if len(extracted_data) > 0:
-        # _extract_data only returns data when signals on
-        # both sides were seen
-
-        err = err.replace("@start_of_data@", "")
-        err = err.replace(extracted_data, "")
-        data_length = _extract_data_length(err, "data")
-        err = err.replace("@end_of_data:%d@\n" % data_length, "")
-        data_bytes = extracted_data.encode()
+    # check for file signal @data_file
```
ntolley, line 405: Is the main reason we don't need start/end flags because when writing to a file you completely control what's written? Then there's no uncertainty what part of the data is relevant

hnn_core/parallel_backends.py
```diff
@@ -292,39 +317,37 @@ def _echo_child_output(out_q):
...
+    file_match = re.search(r"@data_file:(.+):(\d+)@", err)
```
asoplata, line 409: TODO a reviewer should check this regex

hnn_core/parallel_backends.py
```diff
@@ -292,39 +317,37 @@ def _echo_child_output(out_q):
...
+    if file_match is not None:
+        data_file = file_match.group(1)
+        data_length = int(file_match.group(2))
+        err = err[: file_match.start()] + err[file_match.end() :]
```
asoplata, line 413: i'm not famililar enough with `re` results to understand this

hnn_core/mpi_child.py
```diff
@@ -54,14 +50,27 @@ class MPISimulation(object):
     """
 
     def __init__(self, skip_mpi_import=False):
+        signal.signal(signal.SIGTERM, self._handle_sigterm)
```
asoplata: Are we sure that sigterm errors weren't being propagated up already?

hnn_core/mpi_child.py
asoplata: many of the docstrings need some help (including preexisting funcs)

hnn_core/parallel_backends.py
dylansdaniels: so if network is sent here and count increments past a certain point, we just manually reset it to 0?

hnn_core/mpi_child.py
asoplata: To be clear, both this function name and its docstring are no longer accurate and need to be updated

asoplata:
one TODO: this part of the contribution page needs to be updated: https://jonescompneurolab.github.io/hnn-core/stable/contributing.html#notes-on-mpi-for-contributors . This is controlled by https://github.com/jonescompneurolab/hnn-core/blob/master/doc/contributing.md

hnn_core/parallel_backends.py
```diff
@@ -245,45 +271,44 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
...
         The unpickled data.
     """
-    if not data_len == len(data_bytes):
+    if path_data_file is None or data_len == 0:
         # This is indicative of a failure. For debugging purposes.
+        raise RuntimeError("MPI simulation didn't produce a result file. ")
+
+    try:
+        with open(path_data_file, "rb") as f:
+            data_bytes = f.read()
```
asoplata, line 352: Add an except that, if there's an error opening or reading the file, then the tempfile is saved permanently to the local directory, AND the user is informed of what happened, AND both the file and passed data lengths are reported

hnn_core/parallel_backends.py
asoplata: Discused this in code review: in this case, it is a failure to delete the temporary file. In which case, we don't want to report an error or stop the program, because we don't actually care if we can't delete the temp file for any reason

hnn_core/parallel_backends.py
asoplata: We shold add a comment that this is trying to find the appropriate string in `mpi_child.py` (currently L149 in the new version)

hnn_core/parallel_backends.py
asoplata: Update: we decided that in the error case, we do not need to save the tempfile. However, the error should still be reported to the user (along with data length comaprison)

hnn_core/parallel_backends.py
asoplata: this is a comment for myself, not camilo

hnn_core/parallel_backends.py
asoplata: Refactor: remove both the timeout_cycles and the counts (since IIRC they're no longer needed fi we do not have a timeout)

asoplata: Discussion: we're not going to drop timeout_cycles and count_since_last_output at this time, and instead Camilo will leave a comment for a future refactor

kmilo9999:
@asoplata The tests are failing on Python 3.9. Could you take a look at the logs when you get a chance? I wasn’t able to reproduce the issue locally, but it appears to be related to the pip version.

asoplata:
Hey @kmilo9999 it looks like there was an upstream dependency problem with Anaconda itself https://github.com/ContinuumIO/anaconda-issues/issues/13504 but it seems to have been fixed recently. Now it seems that other tests actually related to your code are still failing

hnn_core/tests/test_mpi_child.py
kmilo9999: I applied this change.

hnn_core/tests/test_mpi_child.py
kmilo9999: I moved it to the top

hnn_core/parallel_backends.py
kmilo9999: @asoplata Shouldn't we always log the size of the network being sent to NEURON?

hnn_core/parallel_backends.py
kmilo9999: I added exception messages for the potential errors that the code could raise.

hnn_core/mpi_child.py
kmilo9999: I added docustring to `_write_data_stderr` and `_wait_for_exit_signal`

hnn_core/mpi_child.py
kmilo9999: You are very right. They are being handle properly. This was just left overs of my tests. I removed the code.

hnn_core/parallel_backends.py
kmilo9999: I added a comment above this line describing what it does.

hnn_core/parallel_backends.py
kmilo9999: The previous start/end flags were string flags that the `parallel_backends `code used to verify the integrity of the data returned by rank 0. That made sense when data was being streamed between processes. However, the new implementation simply writes the data to a file and returns the file path and size, so those markers are no longer necessary.

hnn_core/parallel_backends.py
kmilo9999: Not in a meaningful way. The output from NEURON is something like `Trial 1: 0.03 ms...,` so it's quite small. Although, the caller functions allocate memory for a value the same size as out, the reference count should be decremented once `_echo_child_output()` exits. So only one copy of the value should be retained in memory.

hnn_core/parallel_backends.py
kmilo9999: Yes

hnn_core/mpi_child.py
kmilo9999: I added the messages as requested.

hnn_core/parallel_backends.py
kmilo9999: I left a comment on top of the`  white true` describing what to expect from a future improvement of that code.

hnn_core/parallel_backends.py
kmilo9999: Done

hnn_core/mpi_child.py
kmilo9999: Done

hnn_core/parallel_backends.py
kmilo9999: Done

hnn_core/mpi_child.py
kmilo9999: Fixed the docusting as requested

hnn_core/mpi_child.py
```diff
@@ -106,21 +122,41 @@ def _wait_for_exit_signal(self):
                     break
 
     def _write_data_stderr(self, sim_data):
```
asoplata: Camilo: Please split this function into two funcs, one which writes tempfile and one which communicates path to stderr

hnn_core/parallel_backends.py
```diff
@@ -189,15 +240,21 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
             if child_terminated and data_received:
```
asoplata: Would you mind moving this if block to after the `if not child_terminated and count_since_last_output > timeout_cycles:` block?

hnn_core/parallel_backends.py
asoplata: This also shouldn't be an issue since we're no longer using standard IO to transfer the data, so `out` should only be log messages, i.e. tiny

hnn_core/parallel_backends.py
```diff
@@ -245,45 +302,52 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
...
+        raise RuntimeError("MPI simulation didn't produce a result file. ")
+
+    try:
+        with open(path_data_file, "rb") as f:
+            data_bytes = f.read()
+    except FileNotFoundError:
+        warn("Result file not found in path %s" % path_data_file)
+        warn("data length expected %d" % data_len)
+        raise RuntimeError("MPI result file missing: %s" % path_data_file)
+    except PermissionError:
```
asoplata, line 357: Can you please add new tests that ensure that these are caught?

asoplata:
@kmilo9999 Forgot to mention, please add the new changes as a separate commit instead of squashing - I forgot to tell you this at the beginning, but we always "Squash&Merge" from PRs into `master` so that all your commits will be squashed anyways

hnn_core/mpi_child.py
kmilo9999: Done 👍

hnn_core/parallel_backends.py
kmilo9999: Done 👍 
I also added a message indicating that the process is still running and can be canceled if necessary.

hnn_core/parallel_backends.py
kmilo9999: Done 👍 
I added two new tests on `test_mpi_child.py`

hnn_core/tests/test_mpi_child.py
```diff
@@ -179,3 +188,26 @@ def test_data_len_mismatch():
...
+
+
+def test_file_not_found():
+    """_process_child_data raises RuntimeError and warns when file is missing"""
+    with pytest.warns(UserWarning, match="Result file not found"):
+        with pytest.raises(RuntimeError, match="MPI result file missing"):
+            _process_child_data("/nonexistent/path/hnn_mpi_data.pkl", 42)
+
+
+def test_permission_error():
```
ntolley, line 205: Can the docstring be updated to better reflect what functionality is being tested?

To my eyes, it seems like the test is ensuring the "MPI result file is unreadable" error is displayed in the case that there's a permission denied error

If that's correct, then maybe the docstring could be "If error X happens, raise/warn Y"

hnn_core/mpi_child.py
```diff
@@ -51,17 +46,31 @@ class MPISimulation(object):
...
         if skip_mpi_import:
             self.rank = 0
         else:
             from mpi4py import MPI
 
             self.comm = MPI.COMM_WORLD
             self.rank = self.comm.Get_rank()
+            size = self.comm.Get_size()
+
+            if verbose_subprocess:
```
ntolley, line 65: Could a small test be written to ensure that the logger is turned on appropriately when `verbose_subprocess=True`?

hnn_core/mpi_child.py
```diff
@@ -105,22 +121,54 @@ def _wait_for_exit_signal(self):
...
+
+        return tmp_path, len(pickled_bytes)
+
+    def _signal_data_file_stderr(self, tmp_path, pickled_size):
+        """Signal parent process with temp file path
+        and byte count on stderr."""
+
+        # Signal the parent process the file path @data_file:/path/to/file:SIZE@
+        # solves pipe buffering problems
+        # TODO make sure that this string is under 64kb right?
```
ntolley: Could this just be a if/else block that raises an error if `pickled_size` is to big?

hnn_core/mpi_child.py
```diff
@@ -148,9 +209,10 @@ def run(self, net, tstop, dt, n_trials):
     import traceback
 
     rc = 0
+    verbose_subprocess = "--verbose-subprocess" in sys.argv
 
     try:
-        with MPISimulation() as mpi_sim:
+        with MPISimulation(verbose_subprocess=verbose_subprocess) as mpi_sim:
             # XXX: _read_net -> _read_obj, fix later
```
ntolley: this is from before this PR, seems like it's potentially no longer relevant
```suggestion
```

hnn_core/mpi_child.py
kmilo9999: `pickled_size ` represents the size of the file. At what point would you consider a file to be large?

hnn_core/mpi_child.py
kmilo9999: Done

hnn_core/tests/test_mpi_child.py
kmilo9999: I updated the `doctring ` as requested

hnn_core/tests/test_mpi_child.py
```diff
@@ -139,16 +145,15 @@ def test_child_run():
...
+        assert "@data_file:" in stderr_str
 
         # write data to queue
         err_q = Queue()
         err_q.put(stderr_str)
 
         # use _read_stderr to get data_len (but not the data this time)
-        data_len, data = _get_data_from_child_err(err_q)
-        sim_data = _process_child_data(data, data_len)
+        data_len, data_path = _get_data_from_child_err(err_q)
```
ntolley:
```suggestion
        data_len, data_path = _get_data_info_from_child_err(err_q)
```
Perhaps more precise since we're not getting the data anymore

hnn_core/parallel_backends.py
```diff
@@ -145,18 +158,59 @@ def run_subprocess(command, obj, timeout, proc_queue=None, *args, **kwargs):
...
+                                f"nrniv startup time: {time.time() - t_start:.2f}s | "
+                                f"first output: {repr(child_out[:200])}",
+                                flush=True,
+                            )
+                        first_output_received = True
+                    ## end Camilo's code
                     count_since_last_output = 0
                 else:
                     count_since_last_output += 1
                 # look for data in stderr and print child stdout
```
ntolley:
```suggestion
                # look for data length and file location in stderr and print child stdout
```

asoplata:
Thanks a ton for working on this @kmilo9999 !!! This will be really great for us :)

PR #1315, opened 2026-06-02
[DOCS] Refactor MPIBackend simulation process steps

kmilo9999:
Updating `Notes on MPI for contributors` section with the new implementation.


asoplata:
Merging since this passes tests locally, and it's only a website-documentation change that should be completely safe. I believe the issue why our tests are failing has to do with Anaconda itself, not the code changes.

Issue #1317, opened 2026-06-08
[BUG] GUI log toggle button caret no longer visible


dylansdaniels:
The accessibility update to the GUI introduced a bug whereby the caret for toggling the log window is no longer visible due to the "font-size: 0" CSS being added to hide the newly added description="Toggle Log View" to self._log_toggle_btn . The unintended side effect of adding "font-size: 0" is that this hides the caret button as well as the text.

Additionally, there is a pixel shift in the visualization tab when opening/closing the log that is not present in the other tabs. I thought this issue was previously solved, but perhaps I missed it. Regardless, I'll add the fix here since it's small and related to the log. The fix is to add a "gutter" to reserve space for the scrollbar to pop in/out as the log is opened/closed.

dylansdaniels:
the shift in visualization tab broke when the new optimization tab was added, as fixing the shift relief on the tab number:

.param-window-tabs-widget >
.widget-tab-bar:has(.lm-TabBar-tab:nth-child(4).lm-mod-current) +
.widget-tab-contents

note "nth-child(4)", corresponding to the 4th tab. The visualization tab is now the 5th tab

PR #1318, opened 2026-06-08
[MRG] GUI: fix caret coloring and pixel shifting when opening/closing the log

dylansdaniels:
solves #1317


asoplata:
LGTM, merging. Found a new tiny issue in Safari, but it's unrelated to this. Thanks for catching this Dylan!

PR #1319, opened 2026-06-12
[WIP] Place synapses only on sections targeted by connectivity

SatvikSaluja:

PR #1322, opened 2026-06-22
[WIP] Second approach segment location

SatvikSaluja:
This PR builds upon PR https://github.com/jonescompneurolab/hnn-core/pull/1319.

Work done in this PR:

1. Decreases the number of synapses by removing connections to sections that do not receive connections. For a network without drives, the number of synapses is reduced from **6,610 to 2,275**. After adding drives, the number of synapses increases linearly in both cases.

2. Added a connectivity DataFrame, which will replace the existing connectivity object after a future deprecation cycle.

3. Added testing for the connectivity DataFrame.


hnn_core/network_builder.py
```diff
@@ -472,10 +472,35 @@ def _create_cells_and_drives(
...
+                        seg_x[loc] = val
+
+                if seg_x:
+                    cell.seg_x = seg_x  # this gives the value into build() to _create_synapses()
+
                 if src_type_metadata.get("measure_dipole", False):
                     cell.build(sec_name_apical="apical_trunk")
                 else:
                     cell.build()
+                if src_type == 'L5_pyramidal':
```
SatvikSaluja: Just a debug message . Ignore

asoplata:
@SatvikSaluja Can you please add the runscripts that you are using to explore your code (such as the code that produces https://app.notion.com/p/jonescompneurolab/Satvik-GSOC-2026-Main-Page-35744dbbcce680709a98d78640f9c976?source=copy_link#39844dbbcce680ccb20df3a5825c3db1 ) to this PR? Don't worry, we can delete it before we merge the PR into `master`, but seeing the latest version of the runscripts helps us to understand what's happening and help with debugging. Thanks!

Feel free to add them in this branch under something like `hnn-core/satvik_runscripts` or just `hnn-core` (not under `hnn_core`)

hnn_core/network.py
```diff
@@ -1890,11 +1924,14 @@ def add_connection(
...
         if probability != 1.0:
             _connection_probability(conn, probability, conn_seed)
-
         conn["probability"] = probability
         conn["allow_autapses"] = allow_autapses
-
         self.connectivity.append(deepcopy(conn))
+        #i had earlier tried to call build_synapse_trees before line 1934 but it couldnt access connectivity, so it is always accesed after net.connectivity is incremented filled
+        if(not(self.orignal_synapse_creation)):
+            self.build_synapse_tree()
```
ntolley: summary from meeting conversation

We can use the `@getter` decorator and have it's method run `self.build_synapse_tree()` every time `net.synapse_tree` is accessed

hnn_core/network.py
```diff
@@ -1689,6 +1691,67 @@ def _rename_cell_types(self, name_mapping: Dict[str, str]):
     def gid_to_type(self, gid):
         """Reverse lookup of gid to type."""
         return _gid_to_type(gid, self.gid_ranges)
+    '''
```
SatvikSaluja: Ignore this commented out code  from line 1694 to 1754 i Just for keepsakes. Will delete later

hnn_core/cell.py
```diff
@@ -608,15 +607,75 @@ def _compute_section_mechs(self):
...
+        that target same gid at same section and at same seg_x.
+        
+        This opposed the thing we discussed in meeting where we decided on target_type instead of source_type.
+        '''
+        from hnn_core import simple
+        for _, row in target_df.iterrows():
+            target = row['target_type']
+            sec_name = row['actual_section']
+            receptor = row['receptor']
+            segment = row['segX']
```
ntolley:
```suggestion
            segX = row['segX']
```
or `segx`

Should this be more explicit?

hnn_core/network_builder.py
```diff
@@ -566,6 +586,60 @@ def _connect_celltypes(self):
...
+            src_type = row.src_type
+            target_type = row.target_type
+            receptor = row.receptor
+            sec_name = row.actual_section
+            segX = row.segX
+
+            target_cell = self._cells[gid_to_idx[target_gid]]
+
+            connection_name = (
+                f"{_short_name(target_type)}_{_short_name(target_type)}_{receptor}"
```
SatvikSaluja: target_type twice

hnn_core/cells_default.py
```diff
@@ -412,21 +412,21 @@ def _cell_L5Pyr(override_params, pos=(0.0, 0.0, 0), gid=0):
     }
     p_mech = _get_mechanisms(p_all, "L5Pyr", ["soma"] + section_names, mechanisms)
 
+    syn_sections = ["apical_tuft", "apical_oblique", "basal_2", "basal_3"]
```
katduecker, line 506: syn_sections should be read from the data frame for both L2 and L5

hnn_core/network_builder.py
```diff
@@ -384,7 +382,10 @@ def _build(self):
         self._all_spike_gids = h.Vector()
 
         self._record_spikes()
-        self._connect_celltypes()
+        if isinstance(self.net.use_data_frame,bool) and not self.net.use_data_frame:
```
katduecker: This doesn't cover a case in which net.use_data_frame doesn't exist, which could happen when running an old model with the new HNN version. 

add getattr(self.net, "use_data_frame", False) before this line.

hnn_core/cells_default.py
katduecker: Quick thought but I think this would mean that in network_models when creating the cell_types dictionary, target_df would have to be an input to the cell types function 

please double check if this is the way

katduecker: Okay, I see that that is effectively what you're doing because you're just using target_df to build the cell in network_builder. I was worried that syn_sections would somehow relate to get_syn_propos but it doesn't (necessarily, it just reads out the properties of the synapses). But, syn_sections documents which synapses are at which section (see lines 427 following). So I think we'd have to read out from the df what types of synapses can occur on which section (even if they aren't on every section for every cell...)

katduecker: Decided in Meeting Aug 18 to leave this in and adjust only for cell s that will be used in networks with probabilistic synapse distributions. This is only used in add_connection to check if a synapse at a certain location is allowed. If probabilistic, allow all sections.

hnn_core/network.py
```diff
@@ -1890,11 +1897,42 @@ def add_connection(
...
+                    rows.append(
+                        {
+                            "src_gid": src_gid,
+                            "target_gid": target_gid,
+                            "src_type": self.gid_to_type(src_gids[0]),
+                            "target_type": self.gid_to_type(target_gids[0]),
+                            "receptor": receptor,
+                            "template_loc": loc,
+                            "actual_section": section,
+                            "segX": 0.5,  # today it is  hardcodes 0.5 for every synapse
```
```suggestion
                            "segX": 0.5, # synapse is placed in the middle of the section by default
```

hnn_core/network.py
```diff
@@ -2327,6 +2365,39 @@ def __repr__(self):
 
         return entr
 
+import inspect
+import warnings
```
```suggestion
```
These imports should go to the top of the file

hnn_core/network.py
```diff
@@ -471,17 +472,18 @@ def __init__(
                 DeprecationWarning,
                 stacklevel=1,
             )
-
+        self.conn_dataframe = pd.DataFrame()
```
ntolley: I would vote for changing this to `connectivity_df`

`conn_df` is also fine but not easily understandable to a new user

hnn_core/network.py
```diff
@@ -439,6 +439,7 @@ class Network:
     def __init__(
         self,
         params,
+        use_data_frame=False,
```
```suggestion
        use_dataframe=False,
```
I usually see this without the underscore

hnn_core/cell.py
```diff
@@ -609,14 +608,67 @@ def _compute_section_mechs(self):
...
-    def _create_synapses(self, sections, synapses):
+    def create_synapses_orignal(self, sections, synapses):
         """Create synapses."""
         for sec_name in sections:
             for receptor in sections[sec_name].syns:
                 syn_key = f"{sec_name}_{receptor}"
                 seg = self._nrn_sections[sec_name](0.5)
                 self._nrn_synapses[syn_key] = self.syn_create(seg, **synapses[receptor])
 
+    """
```
katduecker: can this go now?

hnn_core/cell.py
```diff
@@ -609,14 +608,67 @@ def _compute_section_mechs(self):
...
+
+        IGNORE THIS COMMENTED OUT BLOCK
+
+        Only reason to keep this function is because it currently reads from NEURON and then modify seg values
+        correctivel .
```
ntolley:
```suggestion
        correctly.
```
?

hnn_core/cell.py
```diff
@@ -609,14 +608,67 @@ def _compute_section_mechs(self):
...
+            
+        def create_synapses_using_synapse_trees(self, syn_tree_gid):
```
ntolley: I think we can remove the synapse trees function now

hnn_core/cell.py
```diff
@@ -609,14 +608,67 @@ def _compute_section_mechs(self):
...
+                            actual_locations.append(actual_loc)
+                            syn_key = f"{source}_{sec_name}_{receptor}_{actual_loc}"
+                            self._nrn_synapses[syn_key] = syn
+
+                        syn_tree_gid[source][receptor][sec_name] = actual_locations
+    """
+
+    def create_synapses_using_connectivity_dataframe(self, target_df):
+        """
+        we would have to first have all the connectiions which are unique
```
ntolley:
```suggestion
        we would have to first have all the connections which are unique
```

hnn_core/network.py
```diff
@@ -1494,15 +1495,21 @@ def _instantiate_drives(self, tstop, n_trials=1):
...
-                        )
+                        if self.use_data_frame:
+                            target_types = set(
+                                self.conn_dataframe.loc[
+                                    self.conn_dataframe["src_gid"] == drive_cell_gid,
+                                    "target_type",
+                                ]
+                            )
+                        else:
+                            conn_idxs = pick_connection(self, src_gids=drive_cell_gid)
```
katduecker, line 1789: Interesting.. it looks like we may not need pick_connection in the future?

hnn_core/cells_default.py
```diff
@@ -252,15 +252,16 @@ def _cell_L2Pyr(override_params, pos=(0.0, 0.0, 0), gid=0):
     }
     p_mech = _get_mechanisms(p_all, "L2Pyr", ["soma"] + section_names, mechanisms)
 
+    syn_sections = ["apical_tuft", "apical_oblique", "basal_2", "basal_3"]
```
ntolley, line 346:
```suggestion
    # This variable allows for specifying where synapses are allowed to be placed
    syn_sections = ["apical_tuft", "apical_oblique", "basal_2", "basal_3"]
```

hnn_core/network.py
```diff
@@ -1494,15 +1495,21 @@ def _instantiate_drives(self, tstop, n_trials=1):
...
+                            target_types = set(
+                                [
+                                    self.connectivity[conn_idx]["target_type"]
+                                    for conn_idx in conn_idxs
+                                ]
+                            )
                         for target_type in target_types:
```
ntolley, line 1753:
```suggestion
                        # loop over drives (one for each target cell
                        # population) and create event times
                        for target_type in target_types:
```

hnn_core/cell.py
```diff
@@ -609,14 +608,67 @@ def _compute_section_mechs(self):
                         p_mech[attr] = [seg_xs, seg_vals]
         return self.sections
 
-    def _create_synapses(self, sections, synapses):
+    def create_synapses_orignal(self, sections, synapses):
```
annacatt: create_synapse_original

hnn_core/network.py
```diff
@@ -1890,11 +1897,42 @@ def add_connection(
         # Probabilistically define connections
         if probability != 1.0:
             _connection_probability(conn, probability, conn_seed)
-
         conn["probability"] = probability
         conn["allow_autapses"] = allow_autapses
-
-        self.connectivity.append(deepcopy(conn))
+        if isinstance(self.use_data_frame,bool) and not self.use_data_frame :
```
ntolley: There should just be a type check on `self.use_dataframe` inside of `network.__init__()`

you can use `_validate_type()` to ensure it is bool

hnn_core/network_models.py
```diff
@@ -125,7 +131,10 @@ def neymotin_2020_model(
         params = op.join(hnn_core_root, "param", "default.json")
     if isinstance(params, str):
         params = read_params(params)
-
+    if isinstance(use_data_frame, str):
```
katduecker: use_data_frame likely going to be replaced by something like conn_type or alike, which can be conn_type="probabilistic" or even a filename. Need to discuss with team

hnn_core/network.py
```diff
@@ -2327,6 +2365,39 @@ def __repr__(self):
 
         return entr
 
+import inspect
+import warnings
+
+
+class ConnectivityList(list):
+    DEPRECATION_MSG = (
```
```suggestion
class ConnectivityList(list):
    """This is a temporary class used to warn users about an API change
        whenever they attempt to access or modify the connectivity."""
    DEPRECATION_MSG = (
```

hnn_core/network_builder.py
ntolley:
```suggestion
        if not self.net.use_data_frame:
```
We can assume the type is correct once `validate_type()` is implemented in `network.__init__()` for this attribute

hnn_core/network.py
katduecker: The warning message says that net.connectivity now lives in net.connectivity_df, but the dictionary is actually called net.conn_dataframe. Please change this to net.connectivity_df

hnn_core/network_builder.py
```diff
@@ -418,20 +419,24 @@ def _gid_assign(self, rank=None, n_hosts=None):
...
-                    target_gids = set()
-                    for conn_idx in conn_idxs:
-                        gid_pairs = self.net.connectivity[conn_idx]["gid_pairs"]
-                        if src_gid in gid_pairs:
-                            target_gids.update(
-                                self.net.connectivity[conn_idx]["gid_pairs"][src_gid]
-                            )
-
+                    if self.net.use_data_frame:
+                        target_gids = set(
```
ntolley, line 426: one suggestion to make this more readable
```python
src_mask = self.net.conn_dataframe["src_gid"] == src_gid
target_gids = set(self.conn_dataframe[src_mask]["target_gid"].values)
```

Or alternatively something like
```python
self.conn_dataframe[src_mask]["target_gid"].unique()
```

hnn_core/network_builder.py
```diff
@@ -546,8 +556,7 @@ def _connect_celltypes(self):
                     pos_idx = src_gid - net.gid_ranges[_long_name(src_type)][0]
                     # NB pos_dict for this drive must include ALL cell types!
                     nc_dict["pos_src"] = net.pos_dict[_long_name(src_type)][pos_idx]
-
-                    # get synapse locations
+                    # get synapse locations( now we modify it as we use synapse_tree)
```
annacatt: Is this synapse_tree reference still needed now that synapse locations come from the data frame?

ntolley:
```suggestion
                    # get synapse locations
```
I'm not sure I understand why this extra note is necessary here since it seems like the code below isn't changed with the new `connectivity_df` updates

hnn_core/network_builder.py
```diff
@@ -556,7 +565,6 @@ def _connect_celltypes(self):
                     # Targeting individual section like soma or apical_tuft
                     else:
                         syn_keys = [f"{loc}_{receptor}"]
-
```
ntolley, line 572: was this a `ruff` formatting change?

we can leave the space if not

hnn_core/network_builder.py
```diff
@@ -566,6 +574,59 @@ def _connect_celltypes(self):
                         )
                         self.ncs[connection_name].append(nc)
 
+    def _connect_celltypes_using_dataframe(self):
```
ntolley, line 595: Can this get a proper docstring briefly summarizing what this function does?

asoplata:
Finished adding the merge-commit that brings in the Duecker model and `master` into here. I also added a small fix that was preventing any use of the classic model functions (e.g. `neymotin_2020_model`) from working at all. Note that **tests are expected to continue failing** as in the most recent pre-merge commit https://github.com/jonescompneurolab/hnn-core/actions/runs/32291636426/job/96193561491 , but that is to be expected.

Let me know @katduecker if you run into issues that you think are tied to the merge commit instead of the code and I'll try to help fix

SatvikSaluja:
Hey @annacatt @asoplata , could you point to the transmembrane current V1 PR? I’ll use my PR, (which has Katharina’s new model changes and my changes), as the base and merge it with the transmembrane current V1, resolving the merge conflicts. Ill raise then it as a new PR .

asoplata:
Hey Satvik! No need to worry about the transmembrane current V1 PR at all. It's a bit confusing, but we do *not* want that PR merged into your branch, and instead we want this PR to continue to ONLY be about the dataframe / connectivity stuff that you have worked so hard on for GSoC.

I've made a custom branch for Anna on her own fork to use that has all of the necessary code like we discussed this week, but to be clear, *we do not expect her branch to actually be merged into `master`, even in the future*. In fact, we do not want the TM current V1 PR merged into `master` at all, and therefore do NOT want it merged into this PR (which do very much want merged into `master`!). Anna's branch is only for her own scientific work, not for general updates to the `hnn-core` library like yours is. Eventually, Dylan will finish a Version 2 of the TM current work, and we will merge that into Anna's branch, at which point we will no longer need the TM V1 PR code at all, and it will never be merged.

Long story short: there's no work that you or Katharina need to worry about with respect to the transmembrane code, either on this or another PR. Thanks for offering to help though!

hnn_core/tests/conftest.py
```diff
@@ -106,6 +106,7 @@ def _run_hnn_core_fixture(
             add_drives_from_params=True,
             legacy_mode=legacy_mode,
             mesh_shape=mesh_shape,
+            use_dataframe=True,
```
asoplata, line 109: Until we actually remove the non-dataframe code (at least one version from now, probably two), we need to test BOTH `use_dataframe=True` and `use_dataframe=False`. This means that anywhere we use `run_hnn_core_fixture`, we need to find a way to test and use both variants, both in the fixture definition here AND every single place where it's used throughout the code.

For here, the fixture definition, the easiest way (assuming this is how Pytest wants us to do to it) is probably to use `request.param` like here https://github.com/jonescompneurolab/hnn-core/blob/e117514ec5fc41be5e4761be413fcffe4ba2638a/hnn_core/tests/test_batch_simulate.py#L18-L58 . I think this works automatically where the fixture is used.

Note that this will also approximately double the time/compute our tests take.

satvik-runscripts/dipole_comparison.py
```diff
@@ -0,0 +1,46 @@
+import pickle
+import numpy as np
```
asoplata, line 2: This and all other `satvik-runscripts` files need to be adapted into the existing test suite.

satvik-runscripts/dipole_comparison.py
```diff
@@ -0,0 +1,46 @@
...
+
+with open("dpl1_orignal.pkl", "rb") as f:
+    dpl1 = pickle.load(f)
+
+with open("dpl2_syanpse_tree.pkl", "rb") as f:
```
asoplata, line 7: It's totally fine if we add tests that write data, and then read in that same data, but only if we also add those filenames to the `.gitignore` so that they're not downloaded with the source code.

hnn_core/network.py
```diff
@@ -312,6 +312,50 @@ def pick_connection(net, src_gids=None, target_gids=None, loc=None, receptor=Non
 
     return sorted(conn_set)
 
+def pick_connection_from_dataframe(net, src_gids=None, target_gids=None, loc=None, receptor=None):
```
SatvikSaluja, line 315: Equivalent function of pick_connection made for dataframe

hnn_core/network.py
annacatt, line 1697: This doesn't work when use_dataframe=True and the model also includes a distal input made of two receptors (unlike the proximal input, which only has one). The reason is that self.connectivity is empty in that case and cannot access self.connectivity[-2]. A workaround is to only access this when use_dataframe=False:

if receptor_idx > 0:
if isinstance(self.use_dataframe, bool) and not self.use_dataframe:
self.connectivity[-1]["src_gids"] = self.connectivity[-2]["src_gids"]

katduecker: Why would you say that the proximal input only has one input? That is not generally true, proximal drives can (and should!) be defined with ampa and nmda weights. I know this is somehow still not the case for the ERP drives we distribute, although I know Darcy and Dylan have been working on an updated drive pattern that has both, and I thought that was part of hnn-data as well.

I would absolutely not regress to not being able when using the dataframe in these instances, we need to find a way to make it work.

katduecker (changes requested):

hnn_core/network.py
SatvikSaluja: we would need an equivalent one pick_connection_from_dataframe

hnn_core/network.py
SatvikSaluja: Can u share the script .i can try to see to it as i have introduced a counter variable which acts same as conn_index. Thanks

katduecker:
Great work @SatvikSaluja! The only thing we thought would be nice is 
1. Figure out if we can safely remove the source gid enforcement (Nick will work on that).
2. Split connectivity_df into external_connectivity_df and recurrent_connectivity_df. I might make a start on this in a secondary PR. net.external_drive will still stay as it is right now.
4. We still need to ponder about how we will store actual connectivity between gid pairs in the dataframe, so not to worry about this now.

hnn_core/cell.py
annacatt, line 844: With use_dataframe=True, net.cell_response.isec ends up empty, no synaptic currents get recorded at all. The synapse filter assumes the format sec_name_receptor, but that naming has changed with the dataframe.

annacatt: Suggested solution:
Instead of
for key in self._nrn_synapses.keys()
if key.startswith(f"{sec_name}_")

use if key.startswith(f"{sec_name}_") or f"_{sec_name}_" in key. 

I'm going to check if this works as intended.

satvik-runscripts/script.py
```diff
@@ -1,11 +1,11 @@
-from hnn_core import neymotin_2020_model, simulate_dipole
+from hnn_core import neymotin_2020_model, simulate_dipole , simple
 net_a = neymotin_2020_model(use_dataframe=True)
 net_a.add_evoked_drive(
     "evprox",
     mu=40,
     sigma=5,
     numspikes=1,
-    location="proximal",
+    location="distal",
```
katduecker, line 8: according to the name this is a proximal drive, why change it?

SatvikSaluja: Sorry ,I might have pushed without reviewing.

SatvikSaluja:
Hey all,
As Austin mentioned an operation in pandas can be done in multiple ways, so to study first and find out better methods, I read from references some sites and documentation by Pandas official. 

I have linked below those I personally found extremely useful for the changes i made and also for in future for new contributors.

For the changes -
1.https://pandas.pydata.org/docs/user_guide/groupby.html
2.https://pandas.pydata.org/docs/user_guide/basics.html
3.https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.equals.html

 

For learning and visualization - 
Pandas Tutor -> https://pandastutor.com/

SatvikSaluja:
As we are taking input dataframe , we will have to validate it . Also, in one of the previous meetign someone suggested we will do it twice , one at loading time(network_creation) and one at simualtion time . 

Also, a quick question . How to validate data frame during network creation time.There can be two instances / time frames

1. Before network_models calls network.__init__  ( ie . i. e.There Doesnt exists a network object , there exists gid_ranges , cell_types)
2. After network_models calls network.__init__ . ( i. e.there exists a network object , there exists gid_ranges , cell_types)


The problem in 1st approach is we wouldnt have a network , so net.cell_types , gid ranges and everything , wouldnt be present . But if we still want to do 1st approach , i can just hardcode it . 
 
I have written code for validating dataframe using 2nd approach.

happy to go either way as instructed

SatvikSaluja:
14 tests failing for now. some of these tests are the ones which modify a parameter and test whether it has changed . Should i go ahead with changing these functions so the change is reflected in data frame also ?

Reasons -
1. set_global_synaptic_gains -> this doesnt update Data Frame. So due to that  test_update_weights_metadata fails. 
2. connectivity_df never reset before JSON reload
3, _rename_cell_types() updates connectivity list but not connectivity_df 

Still finding out for rest :)

hnn_core/cell.py
annacatt: @dylansdaniels @asoplata This is the change I was mentioning to make isec not being empty

katduecker:
@SatvikSaluja Thanks for being on it, here are my thoughts:

1. Handling pandas: can you make some concrete suggestions on how you would do it that we can discuss?
2. Validating the network: 
Some scenarios:
- Yes check at build-time (network_models): user provides dataframe to neymotin_2020_model -> all add_connection code is skipped. In this case, we should validate that the cell_types and gid_ranges are consistent between df and model. This can happen as the elif statement to your current isinstance(use_dataframe, bool) .
- If use_dataframe = True:
- We said we would add columns realized_weight and realized_delay to the df, this would have to happen when the weight and delay are calculated in network_builder, _connect_celltypes_using_dataframe, so at runtime. Please also add a column is_drive to indicate whether the connection is a drive or not.
- At network build time, check if realized values exist and raise warning (in  _connect_celltypes_using_dataframe): There are scenarios in which users might create a network, run it, then change the weights and run again, such as optimization. In this case, realized_weight and realized_delay columns exist before network_builder is called. I think we should stick to always using the realized values if they exist, as that gives users more freedom in changing the connectivity. We should then raise a warning (again  _connect_celltypes_using_dataframe). Checking back and forth between weight and delay and realized values is really tedious, IMHO. I think a warning will make it clear.
- Regular optimization: Typically during optimization, users will want to change weight (maybe delay) and not the realized values. I think the easiest way to allow users to do this is to adjust the set_params function in our optimizer to first to remove those columns if they exist, then change the parameters based on the prediction in that iteration. realized values will again be calculated automatically at runtime.

In short: validation at build-time when df is provided. At runtime, we check if realized values exist and use those, raise a warning that the column already exists and those values will be used. For scenarios in which users might try to change the maximum weight in the context of an optimization, we provide clear instructions/functions on how to do it.

@ntolley what do you think?

SatvikSaluja:
Hey @katduecker 

I actually had this function lying around in one of my script files. I am just committing it as an individual file cos I don't know what the best place for it would be in codebase .I have used AI in this only for the formatted string and errors outputs it would show,as I couldn't think of how I would show multiple errors at same time. 

 However, there still might exist some loopholes which are not covered by this validation 

File - https://github.com/jonescompneurolab/hnn-core/pull/1322/changes#diff-1e97e4b8ecd6a6ad9cf1beaaeb3749fef5fc50fdc98e13435f556ff78421e01b

asoplata:
Decision at Code review 2026-09-24: Yes @SatvikSaluja please use the "parametrize" followed by "if then" paradigm you did in https://github.com/jonescompneurolab/hnn-core/pull/1322/changes/95c8322d0c2d2a2e32da6cacf141c26764b5b5ae#diff-917031bc32ffd62b87f2c0d20be335588f8ff4dab3c6439c3b0739b647ea6af4 for other tests.

katduecker:
@SatvikSaluja the function looks good to me at first glance! :) 

W.R.T. to optimization, I just realized that we don't provide a set_params function, so never mind! So if others are okay with it, I think what's left to do is to add the realized values columns and is_drive. 
And if you could look at viz.py too as a few functions are using connectivity and should use connectivity_df. I did that in my secondary PR as well which we won't merge into this because we are no longer splitting connectivity into intranet and drives... would that be okay?

SatvikSaluja:
Sure, I will look at viz.py . 

Also, 

> As we are taking input dataframe , we will have to validate it . Also, in one of the previous meetign someone suggested we will do it twice , one at loading time(network_creation) and one at simualtion time .
> 
> Also, a quick question . How to validate data frame during network creation time.There can be two instances / time frames
> 
> 1. Before network_models calls network.**init**  ( ie . i. e.There Doesnt exists a network object , there exists gid_ranges , cell_types)
> 2. After network_models calls network.**init** . ( i. e.there exists a network object , there exists gid_ranges , cell_types)
> 
> The problem in 1st approach is we wouldnt have a network , so net.cell_types , gid ranges and everything , wouldnt be present . But if we still want to do 1st approach , i can just hardcode it .
> 
> I have written code for validating dataframe using 2nd approach.
> 
> happy to go either way as instructed

PR #1327, opened 2026-07-16
[MRG] REF/FIX consolidate v_inits and use on read

asoplata:
This makes two changes for the sake of fixing a current bug/regression on `master` that was introduced via
https://github.com/jonescompneurolab/hnn-core/blob/5ca983a3dd53b6e749c377b24bcba64773debbd3/hnn_core/hnn_io.py#L174 as part of #1185.

1. In `hnn_io.py`, when we introduced Katharina's voltage-initialization updates, I included some backwards-compatibility code that was incorrect.

    - The above code in `_read_cell_types`, when reading in an older Network JSON file that does not have a `v0` entry (which is any Network JSON file created before #1185 was merged), would default to setting that cell type's `v0` to be -65. This was probably a misunderstanding on my part at the time of how `h.finitialize` works. Currently, on `master`, this leads to an undetected bug that @annacatt found, where older Network JSON files such as https://github.com/jonescompneurolab/hnn-data/blob/main/workshops/2025-04-09-HNN-online_workshop/gamma_gui_walkthrough/gamma_L5weak_L2weak.json have their voltages initialized to -65, which is NOT correct. Previously, since there was no `v0` key in the Network JSON, these voltages would be initialized at simulation time to their proper, diverse values.

    - In order to fix this while still supporting `v0` usage in Network JSON files, when a Network JSON is read in, it needs to have its correct, corresponding voltage initializtion applied to its `v0` key at read-in time. This is because `hnn_io._read_cell_types` itself creates the network's `Section` objects, instead of relying on `cells_default.py` to do so, which is now (after #1185) where voltage initialization takes place.

    - However, there's a small issue: currently, the different sections' hardcoded initial voltage values are embedded throughout `cells_default.py` itself. This leads to the next change in this PR:

2. So, in `cells_default.py`, similarly to how we manage default network metadata in `network_models.py`, I've extracted all the hardcoded initial voltage values to live in a module-level dictionary. Throughout `cells_default.py`, instead of the initial voltages living inside each subfunction, the subfunctions access the global dictionary.

3. Finally, this allows us to solve the original issue, which is, when reading in a Network JSON with `hnn_io`, we apply the hardcoded default initial voltage values for each default cell type.

This approach is not perfect since it does visually group some parameters of the sections away from their individualized functions, but I think it's the best approach to be able to concisely communicate those initial voltage parameters cleanly to `hnn_io`.


hnn_core/cells_default.py
```diff
@@ -209,7 +240,7 @@ def _cell_L2Pyr(override_params, pos=(0.0, 0.0, 0), gid=0):
         p_all = compare_dictionaries(p_all, override_params)
 
     # All sections of this cell type use the same initial membrane voltage:
-    all_v_init = -71.46
+    v_init = _default_v_init["L2_pyramidal"]
 
     section_names = [
```
ntolley, line 335: Should there be a check to ensure that the keys of `_default_v_init` match `section_names`?

hnn_core/hnn_io.py
```diff
@@ -171,7 +172,9 @@ def _read_cell_types(cell_types_data):
                 diam=section_data["diam"],
                 cm=section_data["cm"],
                 Ra=section_data["Ra"],
-                v0=section_data.get("v0", -65),  # for backwards compatibility
+                v0=section_data.get(
+                    "v0", _default_v_init[cell_name][section_name]
```
ntolley, line 154: Should the unit test be passing a cell_types_data dict where we explicitly `.pop()` the `v0` element and ensure it updates to the default value in the dictionary?

ntolley: This is not found in a our current tests because the current `.json` correctly specify `v0`

I could definitely be convinced this doesn't need a unit test since these are the formats we are distributing going forward, I'll leave it to your call!

ntolley:
Great catch @asoplata! Just two small comments about unit tests and verifying metadata matches, everything else is good with me

hnn_core/hnn_io.py
asoplata: I'm averse to popping the `v0` in case this default dict gets used elsewhere later (which it does not currently)

ntolley: I mean specifically in a unit test where we load in a `.json` file

I made the comment in the spirit of writing a test that triggers the original bug that gave rise to the issue

Let me know if you think this and the other suggested test should be added, I can write them myself and push to the PR, if you don't think they're necessary then we can go ahead and merge

hnn_core/cells_default.py
asoplata: Added check for both L2_pyramidal and L5_pyramidal. Not doing so for basket cells since they're only one section

asoplata: Added a validation check for this in a new commit.

hnn_core/hnn_io.py
asoplata: Added a new test that uses an existing test asset to make sure that the `v0` defaults are used correctly when reading in a network JSON that doesn't have existing `v0` values.

asoplata:
@ntolley I think I've addressed your two comments, so if you approve of the most recent changes, then this should be merge-ready.

asoplata:
@ntolley Can you please check to see if you are okay with merging this?

ntolley:
Perfectly done @asoplata!!

ntolley:
FYI I think there might be a rebase error based on the unit test that failed

asoplata:
Yeah you're right, since Katharina's branch was merged first, I need to update this PR to work with her model as well.

asoplata:
@ntolley I've finished updating this to work with `master` and Katharina's new code, so I think this is ready for re-review and merge now.

I briefly thought about making Katharina's initial voltages a global default dict, but 1) those values are used far less frequently, and they are easily readable inside their appropriate cell-defining functions, and 2) Any network using the Duecker model going forward should be guaranteed to have `v0` values for each section inside their network files, since the big PR to merge the Duecker model has that present. Therefore, I don't think we need to include a "safe default" of providing Duecker's `v0` values; if `model_variant` is Duecker, then it should never not have `v0` values where they need to be.

PR #1328, opened 2026-07-20
[Feat] simulations data store

kmilo9999:
Introduces `_simulations_data_store.py ` containing a DataStore singleton that serves as the single source of truth for all GUI session data, replacing the previously scattered defaultdict instances passed around as function arguments.

#### New file: `hnn_core/gui/_data_store.py`
- Defines `DataStore(dict)` — a singleton subclass of `dict` with three namespaced `defaultdict` stores:
  - `run_simulations` — simulation data from the Run tab
  - `loaded_data` — externally loaded dipole data
  - `networks` — (reserved for future use)
- Each namespace auto-initializes entries to `{"net": None, "dpls": []}` on first access via `defaultdict`
- Exposes convenience properties: `run_simulations`, `loaded_data`, `simulation_names`, `loaded_data_names`, `all_simulation_names`
- Module-level singleton `data_store` is imported by both `gui.py` and `_viz_manager.py`

#### `gui.py`
- Removed passing simulation data dicts as deep function arguments
- `HNNGUI.run_simulations` and `HNNGUI.loaded_simulations` properties now return `data_store` namespaces directly (no more `{"simulation_data": ...}` wrapper)
- `run_button_clicked` and `run_opt_button_clicked` receive the `defaultdict` directly, eliminating the previous `KeyError: 'default'` bug caused by the wrapper
- `update_opt_tab_target_widgets` uses `data_store.simulation_names`
#### `_viz_manager.py`
- `_VizManager.data` property cleaned up — removed `data_store`-related entries, keeping only local viz state (`fig_idx`, `figs`, layout refs)
- All simulation lookups use `data_store` directly (`data_store.run_simulations`, `data_store.loaded_data`, `data_store.all_simulation_names`)
- `add_figure`, `_simulate_edit_figure`, `reset_fig_config_tabs` use `data_store` properties instead of `self.data["simulations"]`

---


hnn_core/gui/_simulations_data_store.py
```diff
@@ -0,0 +1,50 @@
...
+    @property
+    def networks(self):
+        return self["networks"]
+
+    @property
+    def all_simulation_names(self):
+        return list(self["run_simulations"].keys()) + list(self["loaded_data"].keys())
+
+    def reset(self):
+        """Clear all stored simulation/loaded data ( GUI reinitialization)."""
```
ntolley:
```suggestion
        """Clear all stored simulation/loaded data (i.e. GUI reinitialization)."""
```

hnn_core/gui/_simulations_data_store.py
```diff
@@ -0,0 +1,50 @@
...
+    @property
+    def loaded_data(self):
+        return self["loaded_data"]
+
+    @property
+    def networks(self):
+        return self["networks"]
+
+    @property
+    def all_simulation_names(self):
```
ntolley:
```suggestion
    def all_simulation_and_data_names(self):
```
Perhaps too verbose...

hnn_core/gui/_viz_manager.py
```diff
@@ -715,28 +723,17 @@ def _clear_axis(
...
-                if _is_simulation(data["simulations"][sim_name])
-            ),
-            0,  # Default value if no such simulation is found
-        )
-
+    last_loaded_data = simulation_names[-1]
     simulation_selection = Dropdown(
         options=simulation_names,
-        value=simulation_names[sim_index],
+        value=last_loaded_data,  ## Hopefully this works
```
ntolley: Should this actually be the `run_simulations_names` property?

Basing this on the comment from the previous version: "# Find the last simulation with a non-None 'net'"

hnn_core/gui/_simulations_data_store.py
kmilo9999: I renamed the function as you suggested. But, I am open to new names.

hnn_core/gui/_viz_manager.py
kmilo9999: I think the function logic needs to be reworked. This function is called when adding figures and axes from both simulations and data. In the previous implementation, it would fall back to `simulation_names[0]` whenever it couldn't find a non-`None` `net` simulation.

It appears that one of the function's prerequisites is that at least one simulation or data source must exist. If that assumption is valid, it may be better to enforce it explicitly rather than relying on a fallback to the first simulation.

hnn_core/gui/_gui_utils.py
```diff
@@ -0,0 +1,10 @@
+# clear empty trash simulations
+def clear_empty_trash_simualtions(simulation_data):
```
asoplata: typo

hnn_core/gui/_viz_manager.py
```diff
@@ -1420,56 +1418,61 @@ def _simulate_edit_figure(
...
-            idx = config_name_idx[conf_key]
-            conf_widget = ax_control_tabs.children[ax_idx].children[idx]
-            conf_widget.value = conf_val
-
-        buttons = ax_control_tabs.children[ax_idx].children[-2]
-        if operation == "plot":
-            buttons.children[0].click()
-        elif operation == "clear":
-            buttons.children[1].click()
+        try:
```
ntolley, line 1549: Just curious, was there a specific experience that motivated turning this into a try/except block?

hnn_core/gui/_viz_manager.py
```diff
@@ -1080,8 +1077,8 @@ class _VizManager:
 
     Parameters
     ----------
-    gui_data : dict
-        A dict containing all simulation data
+    simulation_store : SimulationDataStore
```
ntolley: this is not currently and argument of the __init__()

hnn_core/gui/_simulations_data_store.py
```diff
@@ -0,0 +1,50 @@
...
+        super().__init__(
+            {
+                "run_simulations": defaultdict(lambda: dict(net=None, dpls=list())),
+                "loaded_data": defaultdict(lambda: dict(net=None, dpls=list())),
+                "networks": {},
+            }
+        )
+
+    @property
+    def run_simulations(self):
```
asoplata: Possible rename this to `simulated_data`? That makes it clear that it's referring to the simulated dipole data

hnn_core/gui/_simulations_data_store.py
```diff
@@ -0,0 +1,50 @@
...
+    @property
+    def loaded_data(self):
+        return self["loaded_data"]
+
+    @property
+    def networks(self):
+        return self["networks"]
+
+    @property
+    def all_simulation_and_data_names(self):
```
asoplata: Given rename of `run_simulations` to `simulated_data`, then we can also rename this to `all_data_names`

hnn_core/gui/_viz_manager.py
ntolley: perhaps this could just be removed?

hnn_core/gui/_viz_manager.py
```diff
@@ -1157,7 +1154,7 @@ def __init__(self, gui_data, viz_layout, fig_default_params):
         # data
         self.fig_idx = {"idx": 1}
         self.figs = {}
-        self.gui_data = gui_data
+        self._simulation_store = data_store.loaded_data
```
ntolley:
```suggestion
```
Seems like this doesn't need to be here since data_store is accessed directly

hnn_core/gui/_gui_utils.py
ntolley:
```suggestion
def clear_empty_trash_simulations(simulation_data):
```

hnn_core/gui/gui.py
```diff
@@ -52,6 +51,8 @@
 from hnn_core.cells_default import _exp_g_at_dist
 from hnn_core.dipole import _read_dipole_txt, average_dipoles
 from hnn_core.gui._logging import logger
+from hnn_core.gui._simulations_data_store import data_store
+from hnn_core.gui._gui_utils import clear_empty_trash_simualtions
```
ntolley:
```suggestion
from hnn_core.gui._gui_utils import clear_empty_trash_simulations
```

hnn_core/gui/_viz_manager.py
asoplata: I don't think we need `VizManager`'s `_external_data_widget` anymore, including this function, its usage in `VizManager`, and references to `VizManager._external_data_widget` in `gui.py` such as https://github.com/jonescompneurolab/hnn-core/blob/5c229cfceced2ef57b094fd4c58a63b7d23658a3/hnn_core/gui/gui.py#L2369-L2376 . Please replace all uses of the attribute/function with usage of your new, better data store. This "external data widget" business was recently added by me in a PR a few months ago, but the problem it was trying to solve is completely obviated by your new data store.

hnn_core/gui/gui.py
```diff
@@ -4895,13 +4904,11 @@ def run_button_clicked(
...
     """Run the simulation and plot outputs."""
-    simulation_data = all_data["simulation_data"]
+    # simulation_data = data_store.run_simulations()
     with log_out:
         try:
             # clear empty trash simulations
-            for _name in tuple(simulation_data.keys()):
-                if len(simulation_data[_name]["dpls"]) == 0:
-                    del simulation_data[_name]
+            clear_empty_trash_simualtions(simulation_data)
```
ntolley:
```suggestion
            clear_empty_trash_simulations(simulation_data)
```

hnn_core/gui/gui.py
```diff
@@ -5902,17 +5909,18 @@ def run_opt_button_clicked(
         try:
             # Sim data setup (and related input validation)
             # --------------------------------------------------------------------------
-            simulation_data = all_data["simulation_data"]
+
+            clear_empty_trash_simualtions(simulation_data)
```
ntolley:
```suggestion
            clear_empty_trash_simulations(simulation_data)
```

hnn_core/gui/_viz_manager.py
```diff
@@ -153,14 +154,14 @@
 
 
 def check_sim_plot_types(new_sim_name, plot_type_selection, target_selection, data):
-    if not _is_simulation(data["simulations"][new_sim_name]):
+    if new_sim_name in data_store.loaded_data_names:
```
asoplata: I'm confused by the argument name `new_sim_name` (this is pre-Camilo). It seems that this conditional only happens if `new_sim_name` is NOT simulated data, but instead is loaded data. If the first argument to this function can be either simulated data or loaded data, then it needs to be renamed to reflect this big backend change that we're making.

hnn_core/gui/gui.py
```diff
@@ -1358,12 +1368,11 @@ def _run_opt_button_clicked(b):
...
                 self._log_out,
                 self.opt_drive_widgets,
-                self.data,
+                self.run_simulations,
+                self.loaded_simulations,
                 self.widget_dt,
                 self.widget_tstop,
                 self.fig_default_params,
-                self.widget_default_smoothing,
-                self.widget_default_scaling,
```
ntolley, line 1366: Why were these two removed?

hnn_core/gui/_viz_manager.py
asoplata: Furthermore, it's not clear to me which of these three dropdowns on the GUI this is used in (or if it used in multiple) 

Image

@ntolley @dylansdaniels I think this ties into some small GUI design decisions about how we provide the user to choose between datasets for visualization.

1. For the first of the three highlighted widgets (at the top in the above screenshot, currently says `Simulation:` ), perhaps it should just say `Data:` and offer a dropdown that lets a user select EITHER simulated OR loaded data.

2. For the second widget (in the middle, currently says `Simulation Data`), perhaps we have it continue to say the same thing **but only allow the user to select from simulated data, NOT loaded data**.

3. For the third widget (in the middle, currently says `Data to compare`), perhaps we have it instead say `Loaded data` or `Experimental data`, and **only allow the user to selected from loaded data, but NOT simulated data**.

This has the implication that **every figure can display one and only one simulated data**, and also **can display one and only one loaded data**. Are we okay with that?

hnn_core/gui/gui.py
```diff
@@ -5180,13 +5187,14 @@ def serialize_simulation(simulations_data, simulation_name):
...
     """
-    simulation_data = simulations_data["simulation_data"]
     csv_trials_output = []
     # CSV file headers
     headers = "times,agg,L2,L5"
     fmt = "%f, %f, %f, %f"
 
-    for dpl_trial in simulation_data[simulation_name]["dpls"]:
+    ## retrieve simulation by name
+    simulation_data = simulations_data[simulation_name]
```
ntolley:
```suggestion
    simulation_data = all_simulation_data[simulation_name]
```
perhaps this is easier to read?

hnn_core/gui/_viz_manager.py
```diff
@@ -579,7 +580,11 @@ def _plot_on_axes(
     # freeze plot type
     widgets_plot_type.disabled = True
 
-    single_simulation = data["simulations"][sim_name]
+    ## It must look up both run_data and laoded_data dicts for sim_name
+    ## Maybe throw exception if there's no simulation?
+    single_simulation = data_store.run_simulations.get(
```
asoplata: This is another GUI-oddity where the new data store paradigm means we should rethink how we want our widgets to guide users.

Currently (on `master`), I believe the widget that `sim_name` is related to is the one highlighted in this screenshot

Image

Where the name is `Simulation Data`. In the example of the screenshot, I have NOT run ANY simulations, but instead only loaded an experimental data file. Currently on `master`, this results in the name of the loaded experimental data being used in the `Simulation Data:` widget, which is a bit confusing since it's not simulated data.

Similar to my comment here https://github.com/jonescompneurolab/hnn-core/pull/1328/changes#r3685862238 , I think that in this function's code (assuming I understand what's happening), we should not simply load in whichever is the most recent data. Instead, if the user has run a new simulation, then `Simulation Data` should be updated according to the appropriate simulation data name. Additionally, if a user has loaded new experimental data, then `Simulation Data` should read `None`, and `Data to Compare` should INSTEAD be updated according to the new experimental loaded data name.

hnn_core/gui/_viz_manager.py
```diff
@@ -597,11 +602,14 @@ def _plot_on_axes(
     # If target_simulations is not None and we are plotting a dipole,
     # we need to plot the target dipole as well.
     if (
```
asoplata, line 624: If we decide to go with the logic I've suggested in prior comments above, then this logic will also probably need to be updated (this whole function will probably have to be refactored):

Currently, in `master`, in the screenshot of https://github.com/jonescompneurolab/hnn-core/pull/1328/changes#r3685862238 , `Simulation Data:` must always have a value (regardless of simulated or loaded), while `Data to Compare` may be `None` or filled. In contrast, if we do what I suggested in the comment, we'll need to redo the control flow inside this function to deal with 3 cases:

1. `Simulation Data == None` and `Data to Compare` is selected with a loaded data.
2. `Simulation Data` is selected with a simulated data and `Data to Compare == None`.
3. Both widgets are filled with a corresponding data type.

hnn_core/gui/_viz_manager.py
asoplata: See my comment at https://github.com/jonescompneurolab/hnn-core/pull/1328/changes#r3685862238 as a possible solution (but one that warrants future work). I think we'll want to maintain two "last data" vars, one for `last_loaded_data` and one for `last_simulated_data`.

hnn_core/gui/_viz_manager.py
```diff
@@ -1174,16 +1171,17 @@ def data(self):
         """Provides easy access to visualization-related data."""
         return {
             "use_ipympl": self.use_dynamic_rendering,
-            "simulations": self.gui_data["simulation_data"],
             "fig_idx": self.fig_idx,
+            "simulations": data_store.run_simulations,  # ---> some tests need this entry
```
asoplata: If `VizManager.data["simulations"]` always refers to non-specific elements of `data_store`, then this key of `VizManager.data` should probably be dropped entirely. I'm not sure it is actually used anywhere in Camilo's PR here.

hnn_core/gui/_viz_manager.py
```diff
@@ -1295,7 +1292,7 @@ def _layout_template_change(self, template_type):
             # Add only simulated data
             sim_names = [
                 simulations
-                for simulations, sim_name in self.data["simulations"].items()
+                for simulations, sim_name in data_store.run_simulations.items()
                 if sim_name["net"] is not None
             ]
```
asoplata: I believe L1293-L1297 to set `sim_names` can be replaced with `DataStore.simulated_data_names` or the equivalent

hnn_core/gui/_viz_manager.py
```diff
@@ -1314,15 +1311,16 @@ def _layout_template_change(self, template_type):
     @unlink_relink(attribute="figs_config_tab_link")
     def add_figure(self, b=None):
         """Add a figure and corresponding config tabs to the dashboard."""
-        if len(self.data["simulations"]) == 0:
+        if not data_store.all_simulation_and_data_names:
             logger.error("No data has been loaded")
```
asoplata:
```suggestion
            logger.error("No data has been simulated or loaded")
```

hnn_core/gui/_viz_manager.py
```diff
@@ -1314,15 +1311,16 @@ def _layout_template_change(self, template_type):
...
             return
 
         template_name = self.widgets["templates_dropdown"].value
         is_data_template = _check_template_type_is_data_dependant(template_name)
+        ## Only checking run simulations
         if is_data_template:
             sim_name = self.widgets["dataset_dropdown"].value
-            if sim_name not in self.data["simulations"]:
+            if sim_name not in data_store.run_simulations:
```
asoplata: See comment here https://github.com/jonescompneurolab/hnn-core/pull/1328/changes#r3685862238 about the first widget, and whether or not it should support BOTH simulation and loaded data.

hnn_core/gui/gui.py
```diff
@@ -711,7 +714,10 @@ def __init__(
         )
 
         # In-memory storage of all simulation and visualization related data
-        self.simulation_data = defaultdict(lambda: dict(net=None, dpls=list()))
+        # self.simulation_data = defaultdict(lambda: dict(net=None, dpls=list()))
+
+        ## Not sure if we need this redirection
+        # self._simulation_store = data_store.run_simulations
```
asoplata: I agree that we can take out redirections like this. I found the redirects VERY confusing when you actually wanted to access the raw data... and fortunately now we won't have to deal with that!

hnn_core/gui/gui.py
```diff
@@ -1259,9 +1264,14 @@ def analysis_config(self):
         }
 
     @property
-    def data(self):
-        """Provides easy access to simulation-related data."""
-        return {"simulation_data": self.simulation_data}
+    def run_simulations(self):
+        """Provides easy access to run simulation data."""
+        return data_store.run_simulations
```
asoplata: Is this a good idea? I personally don't like redirecting a local class's attribute to refer to another class's attribute, since I think always referring to the other class makes it clear that the other class is the one mutating state. However, I don't have much experience dealing with multiple simultaneous interacting Python objects with strong interdependencies and significant state change. In your experience, is this good practice (lowkenuinely curious)?

hnn_core/gui/gui.py
```diff
@@ -1259,9 +1264,14 @@ def analysis_config(self):
...
+
+    @property
+    def loaded_simulations(self):
+        """Provides easy access to loaded simulation data."""
```
asoplata: I think this name was a typo, since the VizManager code at least seems to assume that "loaded" and "simulated" data are mutually exclusive

hnn_core/gui/gui.py
```diff
@@ -4643,32 +4652,32 @@ def get_cell_param_default_value(cell_type_key, param_dict):
     return param_dict[cell_type_key]
 
 
-def on_upload_data_change(change, data, viz_manager, log_out):
+def on_upload_data_change(change, loaded_simulations, viz_manager, log_out):
```
asoplata: See comment https://github.com/jonescompneurolab/hnn-core/pull/1328/changes#r3686099070 . Even though this is not using the `HNNGUI.loaded_simulations` attribute, we should probably refrain from using the name `loaded_simulations` for the same reasons as in that comment.

hnn_core/gui/gui.py
```diff
@@ -4643,32 +4652,32 @@ def get_cell_param_default_value(cell_type_key, param_dict):
...
     # the upload file widget.
     data_dict = change["new"][0]
     dict_name = data_dict["name"].rsplit(".", 1)
-    data_fname = dict_name[0]
+    data_filename = dict_name[0]
     file_extension = f".{dict_name[1]}"
 
     # If data was already loaded return
-    if data_fname in data["simulation_data"].keys():
+    if loaded_simulations.get(data_filename) is not None:
```
asoplata: This should probably check against `DataStore.loaded_data_names`, not an argument, but I may be missing something.

hnn_core/gui/gui.py
```diff
@@ -4643,32 +4652,32 @@ def get_cell_param_default_value(cell_type_key, param_dict):
...
+            logger.error(f"Found existing data: {data_filename}.")
         return
 
     # Read the file
     ext_content = data_dict["content"]
     ext_content = codecs.decode(ext_content, encoding="utf-8")
     with log_out:
         # Write loaded data to data object
-        data["simulation_data"][data_fname] = {
+        loaded_simulations[data_filename] = {
```
asoplata: So it looks like this is appending to the `DataStore` via the `HNNGUI.loaded_simulations` redirect. However, I think it would be more clear if we removed the `loaded_simulations` argument and mutated the state of the `DataStore` directly here in the code. Is that good or bad practice? Is it recommended to use internal class attributes as redirects in order to control the interface between objects?

hnn_core/gui/gui.py
```diff
@@ -4895,13 +4904,11 @@ def run_button_clicked(
     global_gain_textfields,
 ):
     """Run the simulation and plot outputs."""
-    simulation_data = all_data["simulation_data"]
+    # simulation_data = data_store.run_simulations()
```
asoplata: Feel free to just delete these lines instead of commenting them out (unless you're keeping them commented for quick debugging, which I like to do)

hnn_core/gui/gui.py
```diff
@@ -5835,12 +5843,11 @@ def run_opt_button_clicked(
     widget_simulation_name,
     log_out,
     opt_drive_widgets,
-    all_data,
+    simulation_data,
```
asoplata: See my other comments about whether we should remove these arguments and `HNNGUI` redirect attributes (and refer to `DataStore` explicitly), or keep them in.

hnn_core/gui/gui.py
```diff
@@ -6301,7 +6314,9 @@ def main_execution():
                 opt_result["psd_relative_bandpower"] = psd_relative_bandpower
             return optimized_config, opt_result
 
-        except Exception:
+        except Exception as e:
+            print(repr(e))
+            traceback.print_exc()
```
asoplata: Nick and I should manually test this before we merge the PR

hnn_core/gui/_simulations_data_store.py
```diff
@@ -0,0 +1,50 @@
+from collections import defaultdict
+
+
+class SimulationDataStore(dict):
```
asoplata: Maybe just rename to `DataStore` since it's not only containing simulated data

hnn_core/gui/_simulations_data_store.py
```diff
@@ -0,0 +1,50 @@
...
+    def reset(self):
+        """Clear all stored simulation/loaded data ( GUI reinitialization)."""
+        self["run_simulations"] = defaultdict(lambda: dict(net=None, dpls=list()))
+        self["loaded_data"] = defaultdict(lambda: dict(net=None, dpls=list()))
+        self["networks"] = {}
+
+
+# Module-level singleton should be imported by classes that
+# want to save data in DataStore
+data_store = SimulationDataStore()
```
asoplata: We'll definitely want some tests of `DataStore` (probably best to just use 2 or 3 fixtures that set test versions up), but they can be limited to `DataStore.reset()` and consumption of `DataStore` by `HNNGUI` and `VizManager`

hnn_core/gui/_simulations_data_store.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @property
+    def all_simulation_and_data_names(self):
+        return list(self["run_simulations"].keys()) + list(self["loaded_data"].keys())
+
+    def reset(self):
+        """Clear all stored simulation/loaded data ( GUI reinitialization)."""
+        self["run_simulations"] = defaultdict(lambda: dict(net=None, dpls=list()))
+        self["loaded_data"] = defaultdict(lambda: dict(net=None, dpls=list()))
+        self["networks"] = {}
```
asoplata, line 48: FYI @ntolley the `"networks"` attribute of `DataStore` (c.f. approximately `HNNGUI.simulation_data[_sim_name]["net"]`) does not appear to be fully implemented yet. I can't find anywhere where `"networks"` is updated with the `"net"` of simulated networks in Camilo's PR, though I may have just missed it.

asoplata:
Okay I think I've finished my review of what's current. I think @ntolley @dylansdaniels and I need to make a design decision on how the dropdowns handle loaded vs simulated data vis a vis my comment here https://github.com/jonescompneurolab/hnn-core/pull/1328/changes#r3685862238

hnn_core/gui/_viz_manager.py
kmilo9999: Thank you Austin.
I believe this is related to a bug @dylansdaniels mentioned a few weeks ago regarding the dropdown manipulation logic. Let me know what is expected behavior for this workflow. Once I have the specifications, I'll address it in a separate child PR rather than including it in this one.

hnn_core/gui/_viz_manager.py
kmilo9999: The callbacks in the conditionals:
```
            if operation == "plot":
                buttons.children[0].click()
            elif operation == "clear":
                buttons.children[1].click()
```
May throw exceptions. I wasn't able to read the stack trace or print them in the gui console

hnn_core/gui/_simulations_data_store.py
kmilo9999: I am planning on using it for saving/loading networks. I'm adding a todo comment

hnn_core/gui/_viz_manager.py
kmilo9999: I get your idea here and I can apply this change on a child PR. I would like this one to be focused only on including data_store for now.

hnn_core/gui/_viz_manager.py
kmilo9999: @asoplata From what I can see, this approach would only allow comparisons between simulations and loaded data. However, the existing code appears to support comparing two simulations or two loaded datasets as well. Any thoughts on that?

hnn_core/gui/gui.py
kmilo9999: `defaultdict.get()` looks up runs in O(1). Lookup using `loaded_data_names` runs on O(n).
Maybe I should check more places in the code where we can use this mini-optimized code

hnn_core/gui/gui.py
kmilo9999: Removed

hnn_core/gui/gui.py
kmilo9999: Done

hnn_core/gui/gui.py
kmilo9999: This is one of those exceptions where I couldnt see the message. I removed the print

hnn_core/gui/_simulations_data_store.py
kmilo9999: Done

hnn_core/gui/_simulations_data_store.py
kmilo9999: Done

hnn_core/gui/_simulations_data_store.py
kmilo9999: Applied

hnn_core/gui/_simulations_data_store.py
kmilo9999: I added three data_store related tests

hnn_core/gui/_viz_manager.py
asoplata: In Dev meeting 2026-08-03, we agreed:

- The "second" widget in the highlighted screenshot should remain `Simulation Data` and simply only list entries of simulated data. Also, `None` should be a new option for this widget.
- The "third" widget in the highlighted screenshot should be renamed to `Loaded Data` and simply only list entries of loaded data.

- For the "first" widget in the highlighted screenshot, it's a bit more complicated. First, `Figure Template` should have a new entry for "Loaded Data". Under this new `Figure Template` type, when it is selected, instead of there being a widget for selection of `Simulation Data`, there should be a widget for selection of `Loaded Data`. For example, if `Figure Template` is set to `Drive-Dipole`, then the `Simulation Data` widget is shown; if a user then changes `Figure Template` to `Loaded Data`, then the `Simulation Data` widget goes away, and instead a `Loaded Data` widget appears where the users selects only from Loaded data.

hnn_core/gui/_viz_manager.py
kmilo9999: I created PR https://github.com/brown-ccv/hnn-core-ccv/pull/7  applying this change.

hnn_core/gui/_viz_manager.py
kmilo9999: Changes in this PR:
https://github.com/brown-ccv/hnn-core-ccv/pull/8

hnn_core/gui/_viz_manager.py
```diff
@@ -1157,7 +1149,7 @@ def __init__(self, gui_data, viz_layout, fig_default_params):
         # data
         self.fig_idx = {"idx": 1}
         self.figs = {}
-        self.gui_data = gui_data
+        # self._simulation_store = data_store.loaded_data
```
asoplata:
```suggestion
```

hnn_core/gui/_data_store.py
```diff
@@ -0,0 +1,51 @@
...
+    def simulated_data(self):
+        return self["simulated_data"]
+
+    @property
+    def simulated_data_names(self):
+        return self["simulated_data"].keys()
+
+    @property
+    def loaded_data_names(self):
+        return self["loaded_data"].keys()
```
asoplata: @ntolley @dylansdaniels Naming: Do we like "Loaded data" or "Experimental data"? Loaded is technically more accurate (since you can load simulated data), but I feel like "Experimental data" is more beginner-user-friendly. "Experimental" also communicates what it's *intended* for, and I think makes screenshots a tiny bit clearer about what's going on. There's pros/cons for both though.

hnn_core/gui/_viz_manager.py
asoplata: Camilo's relevant PR is https://github.com/brown-ccv/hnn-core-ccv/pull/8

hnn_core/gui/_viz_manager.py
asoplata: As Dylan explained, we are NOT limited to only plotting one of each type: You can always use the `Add plot` button inside the plot manager to add a new time series of either Simulated or Loaded. So we're not restricted to ONLY 2

hnn_core/gui/_data_store.py
asoplata: Decided in code review 2026-08-06 that we would like to change the name to Experimental data instead of Loaded

ntolley: sounds good to me!

hnn_core/gui/_viz_manager.py
asoplata: This change should wait until after https://github.com/brown-ccv/hnn-core-ccv/pull/8 is completed

hnn_core/gui/_viz_manager.py
asoplata: I think this makes sense

hnn_core/gui/gui.py
asoplata: Camilo was right to remove this, this was a mistake by me: the Optimization tab has its own smoothing and scaling widgets (which are affected by, and affect, the Simulation tab's same widgets), but the Run Optimization function itself uses the Optimization tab's version of widgets (which provide the same value). See here https://github.com/jonescompneurolab/hnn-core/blob/a5014172419379f3cbe96655cda8714bf202f767/hnn_core/gui/gui.py#L6239-L6240 . Ergo, Camilo is right, these two arguments are unnecessary.

hnn_core/gui/gui.py
```diff
@@ -711,7 +714,7 @@ def __init__(
         )
 
         # In-memory storage of all simulation and visualization related data
-        self.simulation_data = defaultdict(lambda: dict(net=None, dpls=list()))
+        # self.simulation_data = defaultdict(lambda: dict(net=None, dpls=list()))
```
asoplata:
```suggestion
```

hnn_core/gui/gui.py
```diff
@@ -1258,10 +1260,15 @@ def analysis_config(self):
...
-        return {"simulation_data": self.simulation_data}
+    # @property
+    # def run_simulations(self):
+    #     """Provides easy access to run simulation data."""
+    #     return data_store.simulated_data
+
+    # @property
+    # def loaded_simulations(self):
+    #     """Provides easy access to loaded simulation data."""
+    #     return data_store.loaded_data
```
asoplata: TODO delete once we're sure we don't need this

hnn_core/gui/_viz_manager.py
kmilo9999: @asoplata @ntolley  @dylansdaniels 
I am confused about the third point. Let's look at the the case when the user:
-  Selects the new entry  `Loaded Data` from the `Figure Template` dropdown
-  The new `Loaded data` dropdown shows up (`Simulation Data` dropdown gets hide ) with the list of loaded data. 
-  The user selects a loaded data from the new dropdown

What happens if the user clicks on `Make figure` ?

dylansdaniels: Per 08-17 mtg, we're not sure if we want a template for loaded data. but the way `Make figure` should work is essentially the same as if we were to create a blank figure, set loaded data to some file, and then click `add plot`. default smoothing 0 scaling 1

dylansdaniels: And also, please keep the spacing as it was previously around sub tabs and the data selection dropdown that pops in :) 

we'll try to keep visual changes separate from the functionality. And also, in this case, trying to minimize "jumping" of visual elements

hnn_core/gui/_data_store.py
```diff
@@ -0,0 +1,51 @@
...
+    @property
+    def simulated_data(self):
+        return self["simulated_data"]
+
+    @property
+    def simulated_data_names(self):
+        return self["simulated_data"].keys()
+
+    @property
+    def loaded_data_names(self):
```
ntolley:
```suggestion
    def experimental_data_names(self):
```
@kmilo9999 I think this will amount to a Ctrl-f replace for just about every instance of `loaded` based on the discussion below

hnn_core/gui/_viz_manager.py
ntolley: @asoplata let's discuss this point when you're back, @dylansdaniels made a good case for not having a separate "Loaded data" template but @kmilo9999 already has a nice implementation so it'd be good for you to have a chance to play around with it before making a final decision

hnn_core/gui/gui.py
ntolley: @kmilo9999 I think it's safe to delete this now?

hnn_core/gui/_gui_utils.py
```diff
@@ -0,0 +1,10 @@
+# clear empty trash simulations
+def clear_empty_trash_simulations(simulation_data):
```
ntolley: I actually wonder if this is still an issue with the `data_store` implementation...

hnn_core/gui/_viz_manager.py
```diff
@@ -534,7 +535,7 @@ def _plot_on_axes(
 
     Parameters
     ----------
-    b : ipywidgets.Button
+    button : ipywidgets.Button
     widgets_simulation : ipywidgets.Dropdown
         A dropdown widget that contains all the simulation names.
```
ntolley, line 548:
```suggestion
    button : ipywidgets.Button
       Button that plots new data on top of existing figure.
   widgets_simulation : ipywidgets.Dropdown
       A dropdown widget that contains all the simulation names.
```
?

hnn_core/gui/_viz_manager.py
```diff
@@ -572,14 +573,18 @@ def _plot_on_axes(
...
+    # disable 'add plots' button for types that do not support overlay
     if plot_type in _no_overlay_plot_types:
-        b.disabled = True
+        button.disabled = True
 
     # freeze plot type
     widgets_plot_type.disabled = True
 
-    single_simulation = data["simulations"][sim_name]
+    ## It must look up both run_data and laoded_data dicts for sim_name
```
ntolley:
```suggestion
    ## It must look up both run_data and loaded_data dicts for sim_name
```

hnn_core/gui/_viz_manager.py
```diff
@@ -596,12 +601,12 @@ def _plot_on_axes(
...
-    if (
-        data_widget.value in data["simulations"].keys()
-        and plot_type == "current dipole"
-    ):
+    if data_widget.value in data_store.all_data_names and plot_type == "current dipole":
         target_sim_name = data_widget.value
-        target_sim = data["simulations"][target_sim_name]
+        target_sim = data_store.simulated_data.get(
+            target_sim_name
+        ) or data_store.loaded_data.get(target_sim_name)
```
ntolley:
```suggestion
        )
```
Could there be an issue if you
1) Load in target data named `'example_name'`
2) Run a simulation also named `'example_name'`?

ntolley: For the `target_sim` seems like in this scenario you should look at `data_store.loaded_data` first before checking `simulated_data`

But if I'm understanding correctly maybe there shouldn't be an `or` statement at all as the `target_sim` should exclusively be located in `loaded_data`

hnn_core/gui/_viz_manager.py
ntolley:
```suggestion
    target_names = simulation_names[:-1]
```

hnn_core/gui/_gui_utils.py
kmilo9999: It seems the function is no longer necessary. It was implemented before my time, so I'm not sure what issue it was originally intended to address. The tests pass without it, so I'm going to comment out the code for now in case we need to restore it later.

hnn_core/gui/_viz_manager.py
kmilo9999: Let's talk about this later. It's in the middle of the new dropdown logic, and that is on a different PR. Let's merge the sub PR first, and then we come back to this.

hnn_core/gui/_viz_manager.py
ntolley: @kmilo9999 final decision is to keep the loaded data template functinality, and rename the dropdown option to "Experimental Data - Dipole"

asoplata:
@kmilo9999 Is the order that we should merge these the following?

1. First, merge https://github.com/brown-ccv/hnn-core-ccv/pull/13 into your branch `maint-viz-tab-dropdown`
2. Next, merge https://github.com/brown-ccv/hnn-core-ccv/pull/8 into your branch `feat-sim-data-store`
3. THEN merge this PR itself (1328).

The above seems to the be the order that makes sense to me, since it's newest code first. Thoughts?

kmilo9999:
> @kmilo9999 Is the order that we should merge these the following?
> 
> 1. First, merge [[MAINT] Rename `load data`  to ` experimental data`  brown-ccv/hnn-core-ccv#13](https://github.com/brown-ccv/hnn-core-ccv/pull/13) into your branch `maint-viz-tab-dropdown`
> 2. Next, merge [[MRG] Viz tab fig template dropdown logic brown-ccv/hnn-core-ccv#8](https://github.com/brown-ccv/hnn-core-ccv/pull/8) into your branch `feat-sim-data-store`
> 3. THEN merge this PR itself (1328).
> 
> The above seems to the be the order that makes sense to me, since it's newest code first. Thoughts?

@asoplata Agree!  Let's follow this plan.

asoplata:
Reproducing some TODO ideas that we decided in code review on 2026-08-27 here, since they should be done in this PR instead of any secondary PRs.

Decision: ban duplicates of names such that simulated data and experimental data cannot have an entry of the same name. In the case where they try to load a data file that has the same name as a pre-existing simulation, OR if they try to run a simulation that has the same name as a pre-existing experimental data file, the status bar at the bottom of the GUI should turn red and they should get a log message saying that they need to change the name of the file and/or simulation name (appropriately). Then they should b eable to try again with a new name and the GUI should continue to function.

Decision: In the case where a user wants to upload an experimental data file that is of the same name of an existing experimental data file, then 
1. the new data should overwrite the prior data of the same name
2. The GUI status bar at the bottom should switch to state that looks yellow (indicating "warning")
3. The log should have a warning that the user has OVERWRITTEN the data of that name, and so will need to redo any existing visualizations or optimizations using the new version of the data, since the prior results may no longer be accurate.

asoplata:
Things still left to do (by Austin) from https://github.com/brown-ccv/hnn-core-ccv/pull/8 :

- [x] https://github.com/brown-ccv/hnn-core-ccv/pull/8#discussion_r3816881390 
- [x] some renames
- [x] experimental template list is displaying

hnn_core/gui/_viz_manager.py
asoplata: We discussed this briefly in https://github.com/jonescompneurolab/hnn-core/pull/1328#issuecomment-5444471136 , where no duplicates are allowed, and how we should respond accordingly. These changes still need to be implemented.

hnn_core/gui/_gui_utils.py
asoplata: Deleted in a recent commit, and all tests pass locally for me.

hnn_core/gui/_viz_manager.py
asoplata: I think this was resolved by the addition of `_get_dropdowns_initial_values` at https://github.com/brown-ccv/hnn-core-ccv/blob/63fad788dc7e0de97f552ec2d4f9c71591b3c73a/hnn_core/gui/_viz_manager.py#L957

asoplata:
Pending the last commit's tests, I think I've finished resolving all the trivial comments that were resolved either by the merge ofhttps://github.com/brown-ccv/hnn-core-ccv/pull/8 into here or in my commits after that point (mostly just renames and cleanup).

IMHO, the remaining work is:

- [x] Camilo implements https://github.com/jonescompneurolab/hnn-core/pull/1328#issuecomment-5444471136 
- [x] We finish discussing any unresolved comments
- [x] Make an attempt to build out more testing of GUI/VizMgr, at least for changes in this PR (assuming there are test coverage gaps)

kmilo9999:
> Reproducing some TODO ideas that we decided in code review on 2026-08-27 here, since they should be done in this PR instead of any secondary PRs.
> 
> Decision: ban duplicates of names such that simulated data and experimental data cannot have an entry of the same name. In the case where they try to load a data file that has the same name as a pre-existing simulation, OR if they try to run a simulation that has the same name as a pre-existing experimental data file, the status bar at the bottom of the GUI should turn red and they should get a log message saying that they need to change the name of the file and/or simulation name (appropriately). Then they should b eable to try again with a new name and the GUI should continue to function.
> 
> Decision: In the case where a user wants to upload an experimental data file that is of the same name of an existing experimental data file, then
> 
> 1. the new data should overwrite the prior data of the same name
> 2. The GUI status bar at the bottom should switch to state that looks yellow (indicating "warning")
> 3. The log should have a warning that the user has OVERWRITTEN the data of that name, and so will need to redo any existing visualizations or optimizations using the new version of the data, since the prior results may no longer be accurate.

@asoplata  
Sorry, I miss this comment. Let me finish the job on this PR.

hnn_core/gui/gui.py
```diff
@@ -52,7 +51,10 @@
 from hnn_core.cells_default import _exp_g_at_dist
 from hnn_core.dipole import _read_dipole_txt, average_dipoles
 from hnn_core.gui._logging import logger
-from hnn_core.gui._viz_manager import _idx2figname, _VizManager
+from hnn_core.gui._data_store import data_store
+
+# from hnn_core.gui._gui_utils import clear_empty_trash_simulations
```
ntolley:
```suggestion
```

hnn_core/gui/gui.py
```diff
@@ -4893,15 +4913,19 @@ def run_button_clicked(
...
             # clear empty trash simulations
-            for _name in tuple(simulation_data.keys()):
-                if len(simulation_data[_name]["dpls"]) == 0:
-                    del simulation_data[_name]
+            # Camilo: It seems the function below is not necessary
+            # it's implementation was before my time, so
+            # I am not sure what it supposes to fix.
+            # Test as passing without it. I am going go comment the
+            # code for now in case is needed again.
+            # clear_empty_trash_simulations(simulation_data)
```
ntolley:
```suggestion
```
Let's remove since it seems we don't need this?

hnn_core/gui/gui.py
```diff
@@ -4893,15 +4913,19 @@ def run_button_clicked(
...
     cell_parameters_widgets,
     global_gain_textfields,
+    target_dipole_data_widget,
 ):
     """Run the simulation and plot outputs."""
-    simulation_data = all_data["simulation_data"]
+    simulation_data = data_store.simulated_data
     with log_out:
         try:
             # clear empty trash simulations
```
ntolley:
```suggestion
```

hnn_core/gui/_viz_manager.py
```diff
@@ -1259,78 +1365,84 @@ def build_viz_tab_contents(self):
...
-    def add_figure(self, b=None):
+    def add_figure(self, b=None, ui_action=None):
         """Add a figure and corresponding config tabs to the dashboard."""
-        if len(self.data["simulations"]) == 0:
-            logger.error("No data has been loaded")
+        if not data_store.all_data_names:
+            logger.error("No data has been simulated or loaded")
             return
 
+        sim_name = None
```
ntolley: Found this with @asoplata 

We should change this to `data_name` since it can be either a simulation or experimental data

```suggestion
        data_name = None
```

hnn_core/gui/gui.py
kmilo9999: Done

hnn_core/gui/gui.py
kmilo9999: Done

hnn_core/gui/gui.py
kmilo9999: Done

kmilo9999:
@asoplata @ntolley 
I added the new rule on overwriting experimental data. Test it on your side and let me know your thoughts.

asoplata:
Thanks for adding the warning variant of https://github.com/jonescompneurolab/hnn-core/pull/1328#issuecomment-5444471136 . Can you please add the other variants discussed in the comment, where

1. Loading an experimental data file with the same name as a pre-existing simulation causes an expected failure mode, where the hotbar turns red to indicate an error, and an appropriate log message is added.
2. Trying to run a simulation with the same name as a pre-existing loaded experimental dataset causes an expected failure mode, where the hotbar turns red to indicate an error, and an appropriate log message is added.

kmilo9999:
@asoplata 
I added the other two cases and a test.
I believe `run_button_clicked ` can be defined as a member function of the `HNNGUI `class. I ran a few tests and it seems to work. I know it's not part of this PR, but I can include that change as well if you'd like.

asoplata:
Thanks Camilo! For refactoring `run_button_clicked` (and its cousin, `run_opt_button_clicked`), those should definitely be a second PR, since it would be better do most of our function <-> method exchanging in a single PR.

I made a small change for the failure-mode error status bar: Now, all status bar failures still use the old simulation version where it says `Simulation failed` (now called the status bar's `simulation_failed` case, formerly just `failed` in https://github.com/brown-ccv/hnn-core-ccv/blob/f4fdec1e2878fc8d4e579db41f726f98eb465605/hnn_core/gui/gui.py#L646 ) except for when you try to load in a data file with the same name, wherein you get a similar status bar using `loading_failed` which displays different text.

@ntolley I think the test coverage is complete enough and that this PR is now ready for merge. Unless you want further changes, feel free to merge it anytime between now and thursday Code Review. If not by then, then we'll merge it then.

kmilo9999:
There's a small refactor I'd like to include in this PR. I dont like how the status bar style is defined in the `_simulation_status_contents  `  dictionary (the line you mentioned in your previous comment) .  If you want I can do the refactor in this PR, otherwise I can do it on a different one, so you can proceed to merge this one.

asoplata:
Sure Camilo, feel free to add the refactor in this PR. We'll probably then end up reviewing and merging it at Code Review tomorrow on Thursday.

hnn_core/gui/gui.py
```diff
@@ -2371,9 +2438,9 @@ def update_opt_tab_target_widgets(self):
         #
         # Note: this is what first creates the `VizManager` object's
         # `_external_data_widget` attribute!
-        self.viz_manager._external_data_widget = self.opt_target_widgets[
-            "target_dipole_data"
-        ]
+        # self.viz_manager._external_data_widget = self.opt_target_widgets[
+        #     "target_dipole_data"
+        # ]
```
ntolley:
```suggestion
```
I think this is the last lingering commented out block!

hnn_core/gui/gui.py
```diff
@@ -2371,9 +2438,9 @@ def update_opt_tab_target_widgets(self):
         #
         # Note: this is what first creates the `VizManager` object's
         # `_external_data_widget` attribute!
```
ntolley:
```suggestion
```

hnn_core/gui/gui.py
```diff
@@ -642,99 +646,29 @@ def __init__(
...
-                    padding-left:10px;
-                    color:black;
-                '>
-                    Experimental Data Loaded
-                </div>
-            """,
+        #
+        # Refactor 9/2/2026
+        # The dict maps a key to  the css class defined in gui_styles.css
+        # and the display message(i.e .sim-status-box.status-*)
```
asoplata:
```suggestion
        # _status_box_config maps a status-bar key to both a CSS class defined in 'gui_styles.css'
        # and to a display message (i.e. .sim-status-box.status-*).
```

hnn_core/gui/gui.py
```diff
@@ -1310,28 +1272,39 @@ def _delete_drives_clicked(b):
...
+        def _on_upload_experimental_data_cb(change):
+            ## this function should get data and  take care of the state of the widgets
+            # Check state of the widget
+            if not change["owner"].value:
+                return
+            try:
+                self._on_upload_experimental_data(file_path=change["new"][0])
+            except Exception:
+                self._simulation_status_bar.value = self._simulation_status_contents[
+                    "simulation_failed"
```
asoplata:
```suggestion
                    "loading_failed"
```

hnn_core/gui/gui.py
```diff
@@ -1515,6 +1488,100 @@ def _opt_solver_change(value):
             (self.widget_opt_scaling, "value"),
         )
 
+    def _on_upload_experimental_data(self, file_path):
+        # Parsing path into filename and extension
+        data_dict = file_path
```
ntolley:
```suggestion
    def _on_upload_experimental_data(self, data_dict):
        # Parsing path into filename and extension
```

Issue #995, opened 2025-02-13
DOCS: Fix inconsistencies in existing parameter docstrings


asoplata:
This issue consists of three efforts:

1. There are some cases where the order or presence of parameters in the docstring is inconsistent with the order of parameters in the function itself, such as https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/viz.py#L81-L121 . All docstrings should be updated to be accurate with respect to the argument order in every function.
2. For parameters which have a default value, but their default is different than `None`, these *optional* parameters should have `, default=<value>` added to the end of the parameter's first line in the docstring. As an example, this
```
    color : tuple of float | str
        RGBA value to use for plotting. By default, 'k' (black)
```
should become
```
    color : tuple of float | str, default='k'
        RGBA value to use for plotting. By default, 'k' (black)
```
since this implies *both* that the parameter is optional, and that it has a specific default value. This is in agreement with https://numpydoc.readthedocs.io/en/latest/format.html#parameters .

3. For parameters whose default value is `None`, these parameters should have `, optional` added to the end of the parameter's first line in the docstring, in a similar manner to above.

Note that any amount of work for this is appreciated! Contributors should not feel like they have to fix every single instance of this issue before making a PR and contributing changes.

dikshant182004:
i have updated some of the funtions  docstring if that looks good we can scale it up throughout the code base .

PR #1332, opened 2026-07-23
Fix docstring parameter order and defaults in plot_laminar_lfp and plot_dipole

ArneshBanerjee:
Addresses #995

`plot_laminar_lfp` and `plot_dipole` in `viz.py` had their docstring parameter listings out of order compared to the actual function signatures, and optional parameters were missing the `, default=<value>` or `, optional` annotations required by the numpydoc format. Fixed both functions to match the signature order and added the missing annotations throughout.

### DIFF
--- a/doc/whats_new.md
+++ b/doc/whats_new.md
@@ -60,10 +60,14 @@ merged into `master`! Use `git log` instead and cross-reference instead. -->
 ### People who contributed to this release:
 
 - [Vaishnavi Baghel][]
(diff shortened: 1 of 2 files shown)


asoplata:
Thanks Arnesh! I've made some slight changes to your docs but they are very welcome.

ArneshBanerjee:
Thanks for merging!
Happy to contribute.

PR #1333, opened 2026-07-27
[MRG] MAINT cap ruff version due to breaking changes

asoplata:
This is a temporary dependency version block for `ruff`, which introduced many breaking changes in its latest update https://astral.sh/blog/ruff-v0.16.0 . Basically, `ruff` is expanding the list of rules which both their formatter and linter are checking, which is a good thing. However, it will take time for us to analyze how the new rules change our code, and whether we agree with all the rules or want to exclude some, etc.

It's best if this is tabled for now, since this season is particularly busy.

Edit: In testing, it seems the new ruff formatting doesn't require any changes. However, the linter, accessed via `ruff check hnn_core` reports 746 errors, most of which seem to be code-style (such as using the literal `[]` instead of `list()`).

### DIFF
--- a/.github/workflows/unix_unit_tests.yml
+++ b/.github/workflows/unix_unit_tests.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Install ruff for use before ANY other tests
         shell: bash -el {0}
         run: |
-          pip install --verbose ruff
+          pip install --verbose "ruff<0.16.0"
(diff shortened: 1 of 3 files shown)


ntolley (approved):
Great solution! We'll definitely get with the new ruff times once we get past these bigger PR's :)

PR #1336, opened 2026-08-13
[MRG] ENH Control Bayesian optimizer seed

### LINKED ISSUE #1294: [FEAT] Request: Add ability to control random seeding in Bayesian optimization

floze-the-genius:
## Summary

- Forward the seed passed to Optimizer.fit into Bayesian optimization.
- Use that random state for candidate sampling as well as the Gaussian process.
- Document Bayesian seed support and cover repeated and distinct seeds with a regression test.

Fixes #1294

## Validation

- pytest -q hnn_core/tests/test_general_optimization.py (34 passed)
- make format-check lint spell

### DIFF
--- a/hnn_core/externals/bayesopt.py
+++ b/hnn_core/externals/bayesopt.py
@@ -13,6 +13,7 @@ Adopted from http://atpassos.me/post/44900091837/bayesian-optimization
 
 import warnings
 from sklearn import gaussian_process
+from sklearn.utils import check_random_state
 import numpy as np
 
 import scipy.stats as st
@@ -63,7 +64,7 @@ def bayes_opt(func, x0, cons, acquisition, maxfun=200, debug=False, random_state
     debug : bool, optional
         The default is False.
     random_state : int, optional
-        Random state of the GaussianProcessRegressor. The default is None.
+        Random state of the optimizer. The default is None.
 
     Returns
     -------
@@ -83,13 +84,14 @@ def bayes_opt(func, x0, cons, acquisition, maxfun=200, debug=False, random_state
     best_x = X[np.argmin(y)]
     best_f = y[np.argmin(y)]
     gp = gaussian_process.GaussianProcessRegressor(random_state=random_state)
+    rng = check_random_state(random_state)
 
     if debug:
         print("iter", -1, "best_x", best_x, best_f)
 
     for i in range(maxfun):
         # draw samples from distribution
-        all_x = np.random.uniform(
+        all_x = rng.uniform(
             low=[idx[0] for idx in cons],
             high=[idx[1] for idx in cons],
             size=(10000, len(cons)),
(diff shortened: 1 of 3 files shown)


ntolley:
Thanks for the PR @floze-the-genius !! 

Unfortunately it seems like github's servers are down right now preventing me from seeing the code changes but I will definitely keep an eye out for when they return

hnn_core/optimization/general_optimization.py
```diff
@@ -480,6 +480,7 @@ def _obj_func(predicted_params):
         cons=constraints,
         acquisition=expected_improvement,
         maxfun=max_iter,
+        random_state=obj_fun_kwargs.get("seed"),
```
ntolley, line 483:
```suggestion
        random_state=obj_fun_kwargs.get("seed", 123),
```
I think we will actually have this be a default seed

hnn_core/externals/bayesopt.py
ntolley, line 49:
```suggestion
def bayes_opt(func, x0, cons, acquisition, maxfun=200, debug=False, random_state=123):
```
Related to the change below, we will fix the default seed here

hnn_core/externals/bayesopt.py
```diff
@@ -83,13 +84,14 @@ def bayes_opt(func, x0, cons, acquisition, maxfun=200, debug=False, random_state
     best_x = X[np.argmin(y)]
     best_f = y[np.argmin(y)]
     gp = gaussian_process.GaussianProcessRegressor(random_state=random_state)
+    rng = check_random_state(random_state)
```
ntolley, line 87:
```suggestion
    rng = np.random.default_rng(random_state)
```

hnn_core/externals/bayesopt.py
```diff
@@ -13,6 +13,7 @@
 
 import warnings
 from sklearn import gaussian_process
+from sklearn.utils import check_random_state
```
ntolley, line 16:
```suggestion
```

ntolley (changes requested):
Awesome work @floze-the-genius!!

I only have a few minor suggestions but otherwise this PR is good to go 🚀

asoplata:
I reviewed along with Nick, and I agree with all the suggested changes by Nick. After these changes are made, this PR should be good to merge. Thanks Floze!

PR #1316, opened 2026-06-08
[WIP] Implement Synapse creation at every Segment

SatvikSaluja:
Detailed Documentation in Notion page 

Some files included just for display of code and what simulation i ran to get to the output . (will remove later)

PR #1340, opened 2026-08-28
[WIP] ENH add load_erp_drives to neymotin_2020_model

### LINKED ISSUE #1337: [BUG] Deprecate `add_erp_drives_to_jones_model` and update canonical drive invocation
## Problem

There are currently two common ways to use the HNN-Core API to invoke the "default" model (neymotin 2020) while also invoking the "default" version of the **drives** used to reproduce the common ERP behavior:

1. `neymotin_2020_model(..., add_drives_from_params=True)` which ultimately calls https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/drives.py#L156
2. `neymotin_2020_model(..., add_drives_from_params=False [the default])` followed by calling `add_erp_drives_to_jones_model(...)`:  https://github.com/jonescompneurolab/hnn-core/blob/master/hnn_core/network_models.py#L515

Most developers use the `add_drives_from_params` flag, but several of our tutorials use the `add_erp_drives_to_jones_model` function.

Dylan has recently pointed out that these two ways actually use slightly different seeds, and therefore produce slightly different simulations. While the qualitative results should be robust to differences in seed value, this caused some confusion recently when Dylan was re-doing the ERP tutorial since, for the sake of the tutorial, it's easiest if the seed is kept the same so that the output is EXACTLY the same. In addition, it is confusing to users to have multiple ways to do the same thing (use the "canonical" ERP drives).

I haven't verified that the seeds are different myself, but Dylan reported this, and a reading of the seeds for method 1 (from here https://github.com/jonescompneurolab/hnn-core/pull/1180#issuecomment-3524054931 ) versus method 2 (see the code in `network_models.py` above) do appear to be different at a glance.

## Solution

As discussed in Dev meeting on 2026-08-24, here's what we should do:

1. [ ] Add a THIRD way to add the canonical ERP drives which loads them from a specific hierarchical Network JSON file. This will become the new default way that all documentation should reflect. For the API, this can be a new flag such as `neymotin_2020_model(..., load_erp_drives=True)`.
2. [ ] Do a deprecation cycle for `add_erp_drives_to_jones_model`, and remove it from all tutorial documentation. According to Nick, this was added a long time ago in the way of constructing tutorials before saving Networks to file was possible, and therefore is no longer necessary.
3. [ ] Do a deprecation cycle for `neymotin_2020_model(..., add_drives_from_params=True)`, and remove it from all tutorial documentation. This still uses the old `params.py` code which we eventually want to deprecate (again, see https://github.com/jonescompneurolab/hnn-core/pull/1180#issuecomment-3524054931 ), and *dynamically* creates the drives in a way that is somewhat difficult to follow and is not fixed to a file.

Souravrajvi0:
## Summary

Fixes #1337 by introducing a single canonical way to add default ERP drives to `neymotin_2020_model`.

- Reproduced the reported issue on `master`: `add_drives_from_params=True` and `add_erp_drives_to_jones_model()` use different `event_seed` values and produce different simulation outputs.
- Added packaged `hnn_core/param/neymotin2020_erp_drives.json` (extracted from `neymotin2020_base.json`).
- Added `neymotin_2020_model(..., load_erp_drives=True)` as the recommended API.
- Deprecated `add_erp_drives_to_jones_model()` and `add_drives_from_params=True` with `FutureWarning`.
- Updated tutorial examples and added regression tests.

## Test plan

- [x] Reproduced seed mismatch between legacy ERP drive APIs on `master`
- [x] `pytest hnn_core/tests/test_drives.py` (15 passed)
- [ ] `pytest hnn_core/tests/test_network.py::test_network_models` (passed)
- [ ] Verified `load_erp_drives=True` uses identical seeds across mesh shapes (3x3 vs 10x10)
- [ ] Full `make test` run as well

### DIFF
--- a/hnn_core/drives.py
+++ b/hnn_core/drives.py
@@ -556,4 +556,5 @@ def _load_erp_drives(net, fname=None):
                 f"ERP drive configuration is missing required drive "
                 f"'{drive_name}' in {fname}."
             )
+        # Adds drive to network using `net.add_<drive_type>()` method
         _read_external_drive(net, external_drives[drive_name], read_output=False)
--- a/hnn_core/drives.py
+++ b/hnn_core/drives.py
@@ -4,11 +4,20 @@
 #          Sam Neymotin <samnemo@gmail.com>
 #          Christopher Bailey <bailey.cj@gmail.com>
 
+import json
+from pathlib import Path
+
+import hnn_core
 import numpy as np
 
 from .params import (
     _extract_bias_specs_from_hnn_params,
     _extract_drive_specs_from_hnn_params,
+)
+
+CANONICAL_ERP_DRIVE_NAMES = ("evdist1", "evprox1", "evprox2")
+DEFAULT_ERP_DRIVES_FNAME = (
+    Path(hnn_core.__file__).parent / "param" / "neymotin2020_erp_drives.json"
 )
 
 
@@ -515,3 +524,36 @@ def _create_bursty_input(
         t_array = np.ravel([t_array + cycle_events_isi * cyc for cyc in cycle])
 
     return t_array
+
+
+def _load_erp_drives(net, fname=None):
+    """Load canonical ERP drives from a hierarchical JSON file.
+
+    Parameters
+    ----------
+    net : instance of Network
+        Network object that will be updated with ERP drives in place.
+    fname : str | Path | None
+        Path to a JSON file containing canonical ERP drive definitions.
+        If None (default), loads from the packaged
+        ``neymotin2020_erp_drives.json`` file.
+    """
+    from .hnn_io import _read_external_drive
+
+    if fname is None:
+        fname = DEFAULT_ERP_DRIVES_FNAME
+    else:
+        fname = Path(fname)
+
+    with open(fname, "r") as file:
+        drive_data = json.load(file)
+
+    external_drives = drive_data.get("external_drives", drive_data)
+
+    for drive_name in CANONICAL_ERP_DRIVE_NAMES:
+        if drive_name not in external_drives:
+            raise ValueError(
+                f"ERP drive configuration is missing required drive "
+                f"'{drive_name}' in {fname}."
+            )
+        _read_external_drive(net, external_drives[drive_name], read_output=False)
--- a/doc/whats_new.md
+++ b/doc/whats_new.md
@@ -43,7 +43,9 @@ merged into `master`! Use `git log` instead and cross-reference instead. -->
 
 - `neymotin_2020_model(..., load_erp_drives=True)` loads canonical ERP drives
   from the packaged ``neymotin2020_erp_drives.json`` configuration file. This
-  provides a single, reproducible way to add default ERP drives.
+  provides a single, reproducible way to add default ERP drives. Canonical ERP
+  tutorials and tests now use this flag instead of
+  ``add_drives_from_params=True``.
 - Users can now specify their own custom objective function in two different ways:
     1. Passing a `obj_fun='custom'` to `Optimizer`. If passing `custom`, then users can
       provide a function that simply accepts a single `Dipole` object and accepts
--- a/doc/whats_new.md
+++ b/doc/whats_new.md
@@ -33,9 +33,17 @@ merged into `master`! Use `git log` instead and cross-reference instead. -->
 - User-facing deprecation notices now emit `FutureWarning`, making them visible under
   Python's default warning filters,
   by [William Kang][] in {gh}`1339`. This was their first PR, thanks William!
+- `add_erp_drives_to_jones_model()` is deprecated in favor of
+  `neymotin_2020_model(..., load_erp_drives=True)`.
+- `add_drives_from_params=True` is deprecated in favor of
+  `neymotin_2020_model(..., load_erp_drives=True)` for adding canonical ERP
+  drives.
 
 ### New Features
 
+- `neymotin_2020_model(..., load_erp_drives=True)` loads canonical ERP drives
+  from the packaged ``neymotin2020_erp_drives.json`` configuration file. This
+  provides a single, reproducible way to add default ERP drives.
 - Users can now specify their own custom objective function in two different ways:
     1. Passing a `obj_fun='custom'` to `Optimizer`. If passing `custom`, then users can
       provide a function that simply accepts a single `Dipole` object and accepts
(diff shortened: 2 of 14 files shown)


asoplata:
Thanks Sourav! We plan to take a look at this very soon.

asoplata:
@dylansdaniels @ntolley Note that this will require us choosing EITHER:

1. The seeds used by the `add_drives_from_params=True` argument to `neymotin_2020_model` as described here https://github.com/jonescompneurolab/hnn-core/pull/1180#issuecomment-3524054931 upon which only a small fraction of the tutorials, but quite a few tests are based on
2. OR the seeds from `add_erp_drives_to_jones_model` https://github.com/jonescompneurolab/hnn-core/blob/49f210fb41481ab859537e9b3e32a74117ae95fc/hnn_core/network_models.py#L936 , which Dylan's absolutely massive ERP tutorial rework is based on (among other things that use it, such as other tests and  https://jonescompneurolab.github.io/textbook/content/08_using_hnn_api/plot_firing_pattern.html albeit implicitly).

Option 1 requires less manual work, so we should probably go with that.

hnn_core/drives.py
```diff
@@ -515,3 +524,36 @@ def _create_bursty_input(
...
+
+    external_drives = drive_data.get("external_drives", drive_data)
+
+    for drive_name in CANONICAL_ERP_DRIVE_NAMES:
+        if drive_name not in external_drives:
+            raise ValueError(
+                f"ERP drive configuration is missing required drive "
+                f"'{drive_name}' in {fname}."
+            )
+        _read_external_drive(net, external_drives[drive_name], read_output=False)
```
ntolley, line 560:
```suggestion
        # Adds drive to network using `net.add_<drive_type>()` method
        _read_external_drive(net, external_drives[drive_name], read_output=False)
```

asoplata:
DECISION at Code Review on 2026-09-03, we decided that the specific drive seeds to use for the new `load_erp_drives` argument should be the same as the drives used in https://jonescompneurolab.github.io/textbook/content/05_erps/erp_multiscale_interpretation_01.html#multiscale-interpretation-part-1 . These drive seeds are what Dylan used to record his considerable GUI videos and intensive ERP tutorial, and in theory are the same as the seeds used by default in the GUI. All other tests, tutorials, and documentation need to be updated in tandem with this change to the default seeds and the method to invoke the new default ERP drives.

Souravrajvi0:
Hi all , I rebased this branch onto the latest master to clear the merge conflict,
The only conflict was in (test_network.py). I kept the new update_cell_positions tests from master and removed the old block that used set_cell_positions and add_erp_drives_to_jones_model in that test. That section wasn’t testing load_erp_drives, so I don’t think anything important from this pr was lost there.
The load_erp_drives work itself is unchanged, including the GUI/textbook seeds (272 / 507 / 777). 
Nick’s suggestion in (drives.py) has been added as well.
Let me know if you’d like the broader test/tutorial migration to load_erp_drives=True in this PR

Souravrajvi0:
Hi @asoplata @ntolley, per the Sep 3 decision, I've migrated the canonical ERP tutorials/tests to load_erp_drives=True (plot_connectivity.py, plot_simulate_evoked.py, plot_firing_pattern.py, plus the position/reset/API/pickle tests).

I left add_drives_from_params=True where it tests the params-file loader, custom t_ev timing, legacy-mode dipole comparison (conftest.py), calcium_model, and deprecation tests.

Textbook pages are outside this repo.

asoplata:
Hello Sourav, apologies for not attending the monthly public HNN dev meeting, as it fell on a national US holiday here. In the future, I will amend the description of it on our Contributing Guide to reflect what should happen for holidays.

That said, anyone is welcome at our regular weekly Dev meetings if they're interested -- if you are interested then let me know via the email on my Github profile!

asoplata:
Thanks for all the updates Sourav! There are some changes that I will be pushing directly to your branch that we developers have decided on. I'll let you know if we need help with anything.

Souravrajvi0:
> Thanks for all the updates Sourav! There are some changes that I will be pushing directly to your branch that we developers have decided on. I'll let you know if we need help with anything.

Thanks that sounds good. I’ll review the changes once they’re pushed and let you know if I have any questions.

asoplata:
@ntolley @dylansdaniels I think this is a bigger issue than we thought. In a prior dev/review meeting, we agreed that if `load_erp_drives=True`, we will load `hnn_core/param/neymotin2020_base.json` including its drives, just like the GUI does. However, this raises another issue with the `neymotin_2020_model` API itself: the use of `params`.

If we load `neymotin2020_base.json` using `hnn_io.read_network_configuration`, then when the `Network` is initialized, it's done so using a hardcoded, bare-bones `params` from here https://github.com/jonescompneurolab/hnn-core/blob/e117514ec5fc41be5e4761be413fcffe4ba2638a/hnn_core/hnn_io.py#L506-L509 . However, this is a different value of `params` from what the user may pass in to the `neymotin_2020_model(params,...)` call directly!

We therefore have a conflict and need to make a Decision: do we either:

1. If a user passes in custom `params` via `neymotin_2020_model(params,...)` (either by value or file), do we merge in those params with those of `neymotin2020_base.json`?
2. Or, do we begin to deprecate the `params` argument in `neymotin_2020_model(params,...)` entirely? We've been talking about beginning to deprecate the old parameter usage for a long time now, but this may be the time to do it.

ntolley:
Following up from a conversation with @asoplata on 09/14/26, I think should move towards deprecating the `params` argument entirely

@katduecker has pointed out how modifying network properties beyond the drives remains really difficult from a user perspective, and that the values in `params` are much more approachable

To that point, I think PR's like #1316 and making the API centered around dataframes will resolve this issue while avoiding the problems of maintaining a flat `.json`

asoplata:
In separate discussions with Nick, Dylan, and myself, we've agreed that before we begin "really" deprecating `params`, we need to make a notebook that illustrates how to use the API to change ideally *at least* every parameter that can be easily changed using the legacy "flat JSON" format.

However, for the sake of this PR, we'll decide today at Code Review about whether or not we should make the new `load_erp_drives` argument mutually exclusive with the existing `params`, `add_drives_from_params`, and `legacy_mode` arguments (which we probably will).

asoplata:
DECISION made a code review today: we should make the new load_erp_drives argument mutually exclusive with the existing params, add_drives_from_params, and legacy_mode arguments.

PR #1341, opened 2026-09-01
[WIP] [do not merge yet] Addition of Diesburg & Waller PFC network models

darcywaller:
Adds network function instantiations for two PFC network models: one just with connectivity changes for PFC (following Diesburg et al., 2024 paper) and one with connectivity changes and changes that follow Law et al., 2022 to test contributions of beta events combined with evoked drives to dipoles and spikes. The second is more complex, with added basket population to hold a segregated tuft connection, switching in the destexhe gabab .mod, and some tweaks to the default PFC pyramidal cell distance scaling to replicate exactly results from the old GUI. I have tried to keep these edits limited to appropriate modules.

Leaving this as WIP for now but would love dev team recommendations if I need to configure anything differently. Plan to merge at time of eventual publication and would be happy to add a tutorial for these network configurations then.


ntolley:
This is great! Excited to see the model collection expanding

One thing we might need to think through in the dev meeting is how to organize `cell_default.py`

With @katduecker's model there's some new cells as well (see #1227). One idea is to include the intended publication in the name, but the issue is that we often re-use the same cell models across several publications which could be confusing. However, it at least makes it very clear what reference you should inspect to understand how that cell model was developed

darcywaller:
Sure, that makes sense. Regardless of the naming conventions you all decide on, I can also add details in the docstring of the new pyramidal_PFC function indicating which network functions it's getting called from.

asoplata:
Hey Darcy, I've merged in the latest version of master, made a bunch of small changes, and added some tests for your two new models here. Fortunately I don't think there's a lot of more work to do, but I do have some questions and comments, mostly related to the `waller_pfcbeta_model`:

- First of all, for both of the new models, what is the "standard" drive configuration that you commonly use? I ask just so we can add tests that test the closest thing to your intended use case of the model. In the current tests, I've just added some random drive configs used elsewhere just to have something.
- Do the models need to support `legacy_mode`, or have you never used the option? If you have not, then it may be best if we don't allow its usage, as we are eventually planning to deprecate that option and remove it in the future (in general).
- For `waller_pfcbeta_model` in particular:
    - I think `add_drives_from_params` is incompatible with the waller model; when I've tried to use it, it attempts to add drives at the time of calling `neymotin_2020_model` (different name but same effect of using (`jones_2009_model`). However, `_insert_gabab_population`, which is called immediately after, appears to be incompatible with there being any existing drives in the network. You're not using this argument, are you? If not, we should simply disable it, so that drives are required to be added in a different way. In a recent commit, I have removed this argument and hardcoded its use with `neymotin_2020_model` to be false.
    - Similar issue with `mesh_shape`: Possibly due to the use of `delta` in `_insert_gabab_population`, the only compatible mesh shape that appears to work (for me) with `waller_pfcbeta_model` is (10, 10). I tried (2, 3) and (3, 3), which we use a lot in testing, but these seemed to fail due to certain gabab synapses not being created, probably because of the way that `_insert_gabab_population` runs. I do **not** think we need to change `_insert_gabab_population` to work with multiple mesh shapes, as long as we enforce that `waller_pfcbeta_model` must always have a mesh shape of a particular size, which is what I've done in a recent commit (hard coding it to (10, 10)). If this is a mistake, and you DO want the waller model to be usable by different mesh shapes, then we'll need to fix this in `_insert_gabab_population`.
    - The argument `gabab_params` appears unused (probably leftover from an older implementation), so I've removed it in a recent commit. LMK if this is a mistake.
    - There may possibly be a bug related to how the two following code sections in the main body of `waller_pfcbeta_model` interact . This one that sets the gabab taus

```python
# LAW MODEL-following CHANGES start here
# Update biophysics (increase gabab duration of inhibition)
net.cell_types["L2_pyramidal"]["cell_object"].synapses["gabab"]["tau1"] = 45.0
net.cell_types["L2_pyramidal"]["cell_object"].synapses["gabab"]["tau2"] = 200.0
net.cell_types["L5_pyramidal"]["cell_object"].synapses["gabab"]["tau1"] = 45.0
net.cell_types["L5_pyramidal"]["cell_object"].synapses["gabab"]["tau2"] = 200.0
```

versus this one that overwrites the dictionary of `...["gabab"]` to be a dictionary of just the `mechname`:

```python
# Swap in the realistic GABAb synapses
for cell_entry in net.cell_types.values():
    cell = cell_entry["cell_object"]
    if "gabab" in cell.synapses:
        cell.synapses["gabab"] = {"mechname": "gabab_neymotin2016"}
```

(First, note that `gabab_destexhe.mod` has had its name changed to `gabab_neymotin2016` in the current `master` branch, which has been merged in; the behavior of the mechanism has not changed.) The problem is that the second code block (that sets the `mechname`) actually *removes* the settings that are set in the first code block. For example, after you've run the second code block, if you tried to access one of the `tau` parameters that were set during the first block, such as via

```
net.cell_types["L2_pyramidal"]["cell_object"].synapses["gabab"]["tau1"]
```

then you would get a `KeyError`, since the entirety of `net.cell_types["L2_pyramidal"]["cell_object"].synapses["gabab"]` has been replaced with only `{"mechname": "gabab_neymotin2016"}`.

In other words, the "**bug**" in this case is that the **biophysics changes of the first code block above are never actually used**, since their `...["gabab"]` dictionaries are **overwritten**. Put another way, I think the first code block is never actually used, and is therefore unnecessary. This appears to be intentional, since if you DO try to set `tau1` and `tau2` using the new `gabab_neymotin2016` mechanism, then I get an error since these are **not valid parameters** of the updated gabab mechanism.

I wanted to highlight this just in case you thought that the first code block, which sets the `tau` parameters, is being incorporated, but as far as I can tell, it is not. If you instead do intend to be applying the parameter changes from this block, then there may be a bug that we need to fix.

darcywaller:
Hi Austin - thanks for looking this over and for testing. Answers to your questions:
- Standard drive configurations can be found in the attached code block. (I've included the standards for the evoked drives, but additional inputs via Poisson and beta are needed to match the empirical ERPs - will add a tutorial including this at the time of publication when we merge.)
- I do not use `legacy_mode`, `add_drives_from_params`, or `mesh_shape`, so the solutions for those you proposed sound good to me.
- Thanks for noting the discrepancies with the gabab_params. Indeed, that named variable is probably leftover from a previous implementation, as are the set tau values. tau1 and tau2 are only relevant in the context of the Exp2Syn synapse, which the gabab_destexhe synapse replaces. That .mod file handles the time constants differently than Exp2Syn, and so we do not need to set the taus explicitly. I forgot about that when writing this network function, so we can simply remove those initial lines that are overwritten by the .mod mechs.

[diesburg_2024_net_drivedefaults.py](https://github.com/user-attachments/files/32210255/diesburg_2024_net_drivedefaults.py)
[waller_pfcbeta_drivedefaults.py](https://github.com/user-attachments/files/32210496/waller_pfcbeta_drivedefaults.py)

hnn_core/network_models.py
```diff
@@ -933,6 +936,282 @@ def duecker_ET_model(
...
+    See Also
+    --------
+    neymotin_2020_model
+    diesburg_2024_model
+    law_2021_model
+
+    Notes
+    -----
+    This model differs from ``diesburg_2024_model`` in the following ways:
+    1) Increased gabaB duration of inhibition by increasing tau1/2.
```
asoplata, line 1134:
```suggestion
    1) Increased gabaB duration of inhibition (see file `hnn_core/mod/gabab_neymotin2016.mod`)
```
@darcywaller What do you think about this substitution for this note, now that tau1/tau2 are removed?

darcywaller: sensible to me!

asoplata:
For the Diesburg 2024 model, I've added a new argument to `diesburg_2024_model` named `add_default_drives` that does like it suggests, using the drives from the file you gave.

@darcywaller  For the Waller model, it's a tad more complicated and we need to discuss. The drives that you gave in the file above are dependent on an input variable `stimulus_start` in your script:

1. In the case that a typical user IS expected to understand and provide their own stimulus start time, then I've made a "secondary" PR here https://github.com/darcywaller/hnn-core/pull/2 that, when merged, will go directly onto this PR's branch. That secondary PR has a new function `waller_add_default_drives(net, stimulus_start)` which adds the drives appropriately, given a stimulus start time.

2. Alternatively, the easiest possible way to do it for a user (e.g. someone interested in reproducing or reusing your work) is to do it like I did for `diesburg_2024_model` above (add an argument), but simply hardcode when the `stimulus_start` time is. If you do not, in your paper, expect to be using multiple different stimulus start times, then this is the preferred option, since it's easiest for users to setup.

3. A third option would be a combination of the above: we add two new arguments to `waller_pfcbeta_model`: one which is whether to include default drives, and another that is the `stimulus_start` used to define the drives. This is a hybrid approach of options 1 and 2 above.

darcywaller:
Hi Austin - that's my fault for the confusion. I don't think we need to add a function to the codebase that adds these, i was just using that one to add my evoked drives in my notebook code. I neglected to note the actual start time in the previous code snippet I sent. For now, I think it's easiest to set defaults with just evoked drives, following what is implemented now for the diesburg model. Here are those drives in the attached code.

[waller_pfcbeta_drivedefaults.rtf](https://github.com/user-attachments/files/32245357/waller_pfcbeta_drivedefaults.rtf)

asoplata:
@darcywaller Is there anything else that you think should be added before this is merged to `master`? It might be good to go as-is.

Forget what I said, we shouldn't merge this anytime soon ~~@ntolley @dylansdaniels Added y'all for review since I've added tests and made a few changes. Like I've said to each of y'all in our 1-on-1's, I am planning next (beginning this week) to try to do a comprehensive test-fixture refactor across our entire test suite, so that we can more easily apply test network model variants across different tests. It's easier to work on the refactor if this PR gets merged first.~~

hnn_core/network_models.py
```diff
@@ -933,6 +936,282 @@ def duecker_ET_model(
     return net
 
 
+def diesburg_2024_model(params=None, add_default_drives=False, mesh_shape=(10, 10)):
```
asoplata, line 939: @katduecker Does it make sense to have an `add_default_drives` argument to your `duecker_ET_model` function that works similarly to this, where it appends the "standard" drive set to your Network? We don't have to do this now, especially if you aren't sure about a standard drive set (or if it doesn't scientifically make sense), but just wanted to highlight this approach, which would make things very easy for users.

darcywaller:
Hi - considering that having this incorporated will make it easier to accomplish upcoming dev goals, I'm OK with merging to the development branch. However, I'd like not to advertise or include it as part of a new official release yet given that things may change prior to or during peer review. If you think it won't cause undue user confusion if they do change (which I assume the dev branch is wont to do anyway?), I'm cool with that.

I also will hold off on the tutorial including fitting the model to data with both the default ERP drives + some other add-on Poisson/beta drives until we're ready to include it in an official release, which is the only other thing I'd want to add in advance of a full release for these models.

asoplata:
Hey Darcy, my bad, I totally forgot that you wanted to hold off on actually merging this. It can stay here inside the PR until you actually want it merged (presumably a couple months from now), you just let us know.

There's definitely no need to merge it onto the development branch, so there's no pressure to merge until you actually want to. The big test refactor I was talking about does not depend on this PR. In fact, it would be easier to get my test refactor done independently, then merge that into master, then "backport" it to this PR, so that we would then have a much, much easier way of including your model in our existing tests.

PR #1343, opened 2026-09-03
[WIP] Remove bsl_cor argument and add baseline correction function to network object

katduecker:
This PR removes the bsl_cor argument from simulate_dipole and replaces it with baseline_correction=bool.
The baseline_correction function is now part of the network object.

@ntolley  @asoplata could you check if the logic of this implementation is okay with you? It's a little heavy-handed but I feel it may be necessary for backwards compatibility. 

I also noticed that my baseline correction didn't consider the number of pyramidal neurons so I updated that too.

Will add tests once implementation is final.


asoplata:
I'm going to make a secondary PR on top of this one due to the amount of refactors I'd like to do (and there's a lot of code to fix broken tests, etc.)

asoplata:
Note: as of 2f1ebb9 (i.e. before I added anything), the changes DO seem to produce differences in the test of the actual output of the Duecker model (e.g. using `hnn_core/tests/test_parallel_backends.py::test_compare_duecker_model_output[`). I'm not sure whether this is expected or not, owing to the changes that have been made to `bsl_corr_duecker_ET.json`

asoplata:
@ntolley @katduecker Please see my secondary PR here https://github.com/katduecker/hnn-core/pull/20 which is ready for merge into here (I have no more work planned for it). The details are described there.

katduecker:
> I'm going to make a secondary PR on top of this one due to the amount of refactors I'd like to do (and there's a lot of code to fix broken tests, etc.)

Like I said, I was going to fix tests and write new ones after getting feedback from you ... I'll look at the changes you made.

asoplata:
> > I'm going to make a secondary PR on top of this one due to the amount of refactors I'd like to do (and there's a lot of code to fix broken tests, etc.)
> 
> Like I said, I was going to fix tests and write new ones after getting feedback from you ... I'll look at the changes you made.

I also have not added complete testing that covers all the changes, adding new tests is still a TODO

asoplata:
I think the testing was complete for this in the secondary PR I just merged, but I think @ntolley wanted to do additional review here in this PR on the actual math

PR #1344, opened 2026-09-03
[MRG] FIX move Optimizer docstring to before init

asoplata:
Currently, Optimizer's API docs on the API website https://jonescompneurolab.github.io/hnn-core/stable/generated/hnn_core.optimization.Optimizer.html#hnn_core.optimization.Optimizer do not actually show anything about the arguments etc. for the Optimizer class. This is because the docstring for the class (and its initializer) is incorrectly located after the initializer signature, not after the class signature. Sphinx apparently can't tell the difference, so we need to move the docstring of `__init__` to below the class signature line, not below `__init__` itself.

### DIFF
--- a/hnn_core/optimization/general_optimization.py
+++ b/hnn_core/optimization/general_optimization.py
@@ -23,6 +23,65 @@ import pickle
 
 
 class Optimizer:
+    """Parameter optimization.
+
+    Parameters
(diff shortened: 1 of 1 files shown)


ntolley (approved):
Great catch!! Merging now

PR #1350, opened 2026-09-08
[MRG] MAINT Refactor `test_viz.py`

kmilo9999:
In this PR.

1. Test function` test_viz.py::est_dipole_visualization ` has been split in 5 different functions:
- `test_dipole_viz_decimation_options`
- `test_dipole_viz_dipole_mutiple_layers`
- `test_dipole_viz_multiple_tfr`
- `test_dipole_viz_no_data_in_raster_plt`
- `test_dipole_viz_cell_response_plot_spikes_hist`

The last function `test_dipole_viz_cell_response_plot_spikes_hist` has been also refactored running a number of simulations needed  to complete the test.

### DIFF
--- a/hnn_core/tests/test_viz.py
+++ b/hnn_core/tests/test_viz.py
@@ -46,44 +46,6 @@ def setup_net():
     return net
 
 
-@pytest.fixture
-def run_simulation(setup_net):
-    net = setup_net
-    weights_ampa = {"L2_pyramidal": 5.4e-5, "L5_pyramidal": 5.4e-5}
-    syn_delays = {"L2_pyramidal": 0.1, "L5_pyramidal": 1.0}
-
-    net.add_bursty_drive(
-        "beta_prox",
-        tstart=0.0,
-        burst_rate=25,
-        burst_std=5,
-        numspikes=1,
-        spike_isi=0,
-        n_drive_cells=11,
-        location="proximal",
-        weights_ampa=weights_ampa,
-        synaptic_delays=syn_delays,
-        event_seed=14,
-    )
-
-    net.add_bursty_drive(
-        "beta_dist",
-        tstart=0.0,
-        burst_rate=25,
-        burst_std=5,
-        numspikes=1,
-        spike_isi=0,
-        n_drive_cells=11,
-        location="distal",
-        weights_ampa=weights_ampa,
-        synaptic_delays=syn_delays,
-        event_seed=14,
-    )
-
-    dpl = simulate_dipole(net, tstop=100.0, n_trials=2, record_vsec="all")
-    return net, dpl
-
-
 def _fake_click(fig, ax, point, button=1):
     """Fake a click at a point within axes."""
     x, y = ax.transData.transform_point(point)
@@ -195,6 +157,43 @@ def test_network_visualization(setup_net):
 
 
 class TestDipoleViz:
+    @pytest.fixture
+    def run_simulation(setup_net):
+        net = setup_net
+        weights_ampa = {"L2_pyramidal": 5.4e-5, "L5_pyramidal": 5.4e-5}
+        syn_delays = {"L2_pyramidal": 0.1, "L5_pyramidal": 1.0}
+
+        net.add_bursty_drive(
+            "beta_prox",
+            tstart=0.0,
+            burst_rate=25,
+            burst_std=5,
+            numspikes=1,
+            spike_isi=0,
+            n_drive_cells=11,
+            location="proximal",
+            weights_ampa=weights_ampa,
+            synaptic_delays=syn_delays,
+            event_seed=14,
+        )
+
+        net.add_bursty_drive(
+            "beta_dist",
+            tstart=0.0,
+            burst_rate=25,
+            burst_std=5,
+            numspikes=1,
+            spike_isi=0,
+            n_drive_cells=11,
+            location="distal",
+            weights_ampa=weights_ampa,
+            synaptic_delays=syn_delays,
+            event_seed=14,
+        )
+
+        dpl = simulate_dipole(net, tstop=100.0, n_trials=2, record_vsec="all")
+        return net, dpl
+
     def test_decimation_options(self, run_simulation):
         """Test basic dipole visualisations and decimation."""
         _, dpls = run_simulation
--- a/hnn_core/tests/test_viz.py
+++ b/hnn_core/tests/test_viz.py
@@ -153,7 +153,6 @@ def test_network_visualization(setup_net):
     _fake_click(fig, ax_src, [pos[0], pos[1]])
     pos_in_plot = ax_target.collections[2].get_offsets().data[0]
     assert_allclose(pos[:2], pos_in_plot)
-    plt.close("all")
 
 
 class TestDipoleViz:
@@ -308,8 +307,6 @@ def test_drive_strength(setup_net):
 
     any_plot = any(ax.lines or ax.patches or ax.images for ax in figure.axes)
     assert any_plot  # At least one axis contains graphical elements
-
-    plt.close("all")
 
 
 class TestCellResponsePlotters:
@@ -801,8 +798,6 @@ class TestCellResponsePlotters:
         )
         _check_inverted_spike_axes(fig)
 
-        plt.close("all")
-
 
 def test_network_plotter_init(setup_net):
     """Test init keywords of NetworkPlotter class."""
@@ -829,7 +824,6 @@ def test_network_plotter_init(setup_net):
     assert net_plot.vsec_array.shape == (159, 1)
     assert net_plot.color_array.shape == (159, 1, 4)
     assert net_plot._vsec_recorded is False
-    plt.close("all")
 
 
 def test_network_plotter_simulation(setup_net):
@@ -864,7 +858,6 @@ def test_network_plotter_simulation(setup_net):
     assert net_plot.color_array.shape == (159, 21, 4)
     assert net_plot._vsec_recorded is True
     assert isinstance(net_plot._cbar, Colorbar)
-    plt.close("all")
 
 
 def test_network_plotter_setter(setup_net):
@@ -914,7 +907,6 @@ def test_network_plotter_setter(setup_net):
 
     with pytest.raises(RuntimeError, match="Network must be simulated"):
         net_plot.trial_idx = 1
-    plt.close("all")
 
 
 def test_network_plotter_export(tmp_path, setup_net):
@@ -931,5 +923,3 @@ def test_network_plotter_export(tmp_path, setup_net):
     net_plot.export_movie(path_out, dpi=200, decim=100, writer="pillow")
 
     assert path_out.is_file()
-
-    plt.close("all")
--- a/hnn_core/tests/test_viz.py
+++ b/hnn_core/tests/test_viz.py
@@ -195,7 +195,7 @@ def test_network_visualization(setup_net):
 
 
 class TestDipoleViz:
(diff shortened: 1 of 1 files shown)


hnn_core/tests/test_viz.py
```diff
@@ -257,13 +230,62 @@ def test_dipole_visualization(setup_net):
...
+        n_drive_cells=11,
+        location="distal",
+        weights_ampa=weights_ampa,
+        synaptic_delays=syn_delays,
+        event_seed=14,
+    )
+
+    # simulation second run
+    simulate_dipole(net, tstop=100.0, n_trials=2, record_vsec="all")
+
```
ntolley:
```suggestion
```
@kmilo9999 I would recommend placing this chunk of code in a fixture that replaces `setup_net` for the other refactored tests

Since most of the refactored functions start with
```py
net = setup_net
dpls = simulate_dipole(net, tstop=100.0, n_trials=2, record_vsec="all")
```
it would be best pass in a fixture so that this simulation is only run once and then re-used by the relevant  tests

hnn_core/tests/test_viz.py
```diff
@@ -257,13 +230,62 @@ def test_dipole_visualization(setup_net):
...
+    net.cell_response.plot_spikes_raster()
     # test cell response plotting
     with pytest.raises(TypeError, match="trial_idx must be an instance of"):
         net.cell_response.plot_spikes_raster(trial_idx="blah", show=False)
     net.cell_response.plot_spikes_raster(trial_idx=0, show=False)
     fig = net.cell_response.plot_spikes_raster(trial_idx=[0, 1], show=False)
     assert len(fig.axes[0].collections) > 0, "No data plotted in raster plot"
 
+
+def test_dipole_viz_cell_response_plot_spikes_hist(setup_net):
```
ntolley:
```suggestion
def test_dipole_viz_cell_response_plot_spikes_hist(setup_net, run_simulation):
```
Is it possible to pass in multiple fixtures to a test?

Here it makes sense to have both a simulation run with an empty network (from `setup_net`), but then another fixture called `run_simulation` could return
```py
net, dpl = run_simulation
```
which corresponds to the code block that I indicated lower down in the test

The `run_simulation` fixture could be used by almost every other test and replace `setup_net`. This should reduce the number of redundant simulations run during testing

ntolley (changes requested):
Great work @kmilo9999 !!

I have one suggestion for creating a new fixture called `run_simulation` that is passed to these tests. It should include the following code:
```py
net = setup_net
weights_ampa = {"L2_pyramidal": 5.4e-5, "L5_pyramidal": 5.4e-5}
syn_delays = {"L2_pyramidal": 0.1, "L5_pyramidal": 1.0}

net.add_bursty_drive(
    "beta_prox",
    tstart=0.0,
    burst_rate=25,
    burst_std=5,
    numspikes=1,
    spike_isi=0,
    n_drive_cells=11,
    location="proximal",
    weights_ampa=weights_ampa,
    synaptic_delays=syn_delays,
    event_seed=14,
)

net.add_bursty_drive(
    "beta_dist",
    tstart=0.0,
    burst_rate=25,
    burst_std=5,
    numspikes=1,
    spike_isi=0,
    n_drive_cells=11,
    location="distal",
    weights_ampa=weights_ampa,
    synaptic_delays=syn_delays,
    event_seed=14,
)

dpl = simulate_dipole(net, tstop=100.0, n_trials=2, record_vsec="all")
return net, dpl
```

hnn_core/tests/test_viz.py
asoplata: Looks like you can from https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#use-fixtures-in-classes-and-modules-with-usefixtures via stuff like

```python
@pytest.mark.usefixtures("cleandir", "anotherfixture")
def test():
    ...
```

hnn_core/tests/test_viz.py
asoplata: You could also use something similar to the `class TestCellResponsePlotters` later in the file in order to group together tests that test plotting of `Dipole` objects.

hnn_core/tests/test_viz.py
asoplata: This test should be merged with and/or included with other `CellResponse` plotting tests, not standalone

hnn_core/tests/test_viz.py
asoplata: In addition to the expansion of use of fixtures, this function ALSO appear at first glance to belong as a method under `TestCellResponsePlotters` since the primary thing being tested is `net.cell_response.plot_spikes_hist`

asoplata:
Done with my review for now; mostly just agreeing with Nick's suggestions, and expanding the class approach used in the same file. After these changes, this will probably be merge-ready

kmilo9999:
@ntolley @asoplata I applied your suggestions to the PR. Please review it.

hnn_core/tests/test_viz.py
asoplata: Resolved by Camilo's latest commit

hnn_core/tests/test_viz.py
asoplata: Resolved in the latest commit by Camilo, where the simulation fixture uses the network setup fixture.

hnn_core/tests/test_viz.py
```diff
@@ -210,85 +210,114 @@ def test_dipole_visualization(setup_net):
...
-    )
-    net.cell_response.plot_spikes_hist(
-        spike_types={"group1": ["beta"]}, color={"group1": "r"}
-    )
+    def test_dipole_viz_cell_response_plot_spikes_hist(self, run_simulation):
+        """Test spike histogram plotting with trial_idx and color argument variations."""
+        net, _ = run_simulation
+        net.cell_response.plot_spikes_hist()
+        # simulation second run. first one it's in the run_simulation definition
+        simulate_dipole(net, tstop=100.0, n_trials=2, record_vsec="all")
```
asoplata: Why was there a need to add a second simulation run for this test? I don't think that running a second simulation should change the behavior of `plot_spikes_hist`, and if there have been two simulations of two trials, I don't think `plot_spikes_hist` is going to aggregate them.

hnn_core/tests/test_viz.py
asoplata: done in new commit.

asoplata:
Applied some new refactors, and now I think this is ready for merge / final review @ntolley

hnn_core/tests/test_viz.py
kmilo9999: Not sure, that's why I left the comment so you could take a look at that part. The test was failing if the simulation is run once.

asoplata: Okay thanks, I've removed it in the last commit

hnn_core/tests/test_viz.py
```diff
@@ -153,142 +153,120 @@ def test_network_visualization(setup_net):
...
-    with pytest.raises(ValueError):
-        net.cell_response.plot_spikes_hist(color={"beta_prox": "z", "beta_dist": "g"})
-    with pytest.raises(TypeError, match="Dictionary values of color must"):
-        net.cell_response.plot_spikes_hist(color={"beta_prox": 123, "beta_dist": "g"})
-    with pytest.raises(ValueError, match="'beta_dist' must be"):
-        net.cell_response.plot_spikes_hist(color={"beta_prox": "r"})
-    plt.close("all")
+        # pytest deprecation warning for tmin and tmax
+        with pytest.warns(FutureWarning, match="tmin and tmax are deprecated"):
+            plot_dipole(dpls[0], show=False, tmin=10, tmax=100)
```
ntolley: This could be moved to `test_decimation_options()`, it's not a perfect fit but it belongs better there because `plot_dipole()` is used more frequently

hnn_core/tests/test_viz.py
```diff
@@ -330,8 +308,6 @@ def test_drive_strength(setup_net):
     any_plot = any(ax.lines or ax.patches or ax.images for ax in figure.axes)
     assert any_plot  # At least one axis contains graphical elements
 
-    plt.close("all")
```
ntolley, line 333: I think this should stay, in fact perhaps there should be a `plt.close("all")` at the end of every viz test?

ntolley: or is this less of a problem now that the tests have been refactored into smaller individual tests?

earlier the close statements were added because we were breaking the runners by adding so many figures, but those were all one big script run sequentially

hnn_core/tests/test_viz.py
```diff
@@ -722,6 +698,104 @@ def test_spikes_from_read_spikes(self, base_simulation_spikes):
             f"{n_plotted_raster_overlay}"
         )
 
+    def test_no_data_in_raster_plt(self, base_simulation_spikes):
+        """Test that the raster plot contains data for various trial_idx inputs."""
+        net, _ = base_simulation_spikes
```
ntolley:
```suggestion
    def test_no_data_in_raster_plt(self, class_setup_net):
        """Test that the raster plot contains data for various trial_idx inputs."""
        net = class_setup_net
        dpls = simulate_dipole(net, tstop=100.0, n_trials=2, record_vsec="all")
```
I think this is a mistake, the current `base_simulation_spikes` would run a simulation that does indeed have spikes

We want to test the situation where no spikes have occurred (e.g. simulating an empty network as suggested). This is because previously trying to make this plot with no spikes threw an error which was a bug

hnn_core/tests/test_viz.py
```diff
@@ -722,6 +698,104 @@ def test_spikes_from_read_spikes(self, base_simulation_spikes):
...
+        net.cell_response.plot_spikes_raster()
+        # test cell response plotting
+        with pytest.raises(TypeError, match="trial_idx must be an instance of"):
+            net.cell_response.plot_spikes_raster(trial_idx="blah", show=False)
+        net.cell_response.plot_spikes_raster(trial_idx=0, show=False)
+        fig = net.cell_response.plot_spikes_raster(trial_idx=[0, 1], show=False)
+        assert len(fig.axes[0].collections) > 0, "No data plotted in raster plot"
```
ntolley:
```suggestion
        assert len(fig.axes[0].collections) == 0
```

hnn_core/tests/test_viz.py
asoplata: There is an automatically run `plt.close("all")` call at the end of every test in this file -- it's done by the first fixture in this file up above at https://github.com/brown-ccv/hnn-core-ccv/blob/f5aaed2c0669f63eb1719ada1e2840e907b65575/hnn_core/tests/test_viz.py#L32-L36 . Notice the `autouse` keyword, which, when combined with `yield` in this case, I believe causes the fixture to run AFTER every test (as opposed to before every test if we did not have the `yield` keyword).

hnn_core/tests/test_viz.py
asoplata: Done in latest commit.

hnn_core/tests/test_viz.py
asoplata: Thanks for catching this Nick, and thanks for addressing this in https://github.com/brown-ccv/hnn-core-ccv/pull/14

hnn_core/tests/test_viz.py
asoplata: Nick addressed this in https://github.com/brown-ccv/hnn-core-ccv/pull/14

asoplata:
Setting to auto-merge now that Nick and I have finished our review. Thanks @kmilo9999 !!

PR #1351, opened 2026-09-10
[MRG] MAINT Ignore pixi files for codespell

ntolley:
As the title states, I found that when using  a pixi environment the codespell during `make test` goes into every single file in the environment and takes wayyyy too long. This just adds the relevant files to the ignore list.


asoplata:
LGTM

PR #1352, opened 2026-09-10
[MRG] FIX Add back and doc param fname handling

asoplata:
This adds back in something that was accidentally deleted in https://github.com/jonescompneurolab/hnn-core/commit/49f210fb41481ab859537e9b3e32a74117ae95fc which was the ability for network models functions (such as `neymotin_2020_model`) to take their argument `params` as a string or Path representing the location of a filename. You can compare the current `master` branch
https://github.com/jonescompneurolab/hnn-core/blob/e117514ec5fc41be5e4761be413fcffe4ba2638a/hnn_core/network_models.py#L219-L221 to stable v0.6.1 here
https://github.com/jonescompneurolab/hnn-core/blob/v0.6.1/hnn_core/network_models.py#L75-L76 to see that accepting `params` as a `str` was accidentally removed.

This was only surfaced by our CircleCI's running of `plot_simulate_somato.py` failing
https://app.circleci.com/pipelines/github/jonescompneurolab/hnn-core/5771/workflows/d9c91460-3004-49e7-bc50-1277ca3d3a6d/jobs/6164 since, surprisingly, we did not have any tests for this extremely important argument case! This PR also adds a test to make sure that the four ways of passing `params` work, and are identical.

### DIFF
--- a/hnn_core/network_models.py
+++ b/hnn_core/network_models.py
@@ -171,15 +171,21 @@ def neymotin_2020_model(
 
     Parameters
     ----------
-    params : str | dict | None
-        The path to the parameter file for constructing the network.
-        If None, parameters loaded from default.json
-        Default: None
-    add_drives_from_params : bool
+    params : str | Path | dict | None, default=None
+        Custom Network parameters to use, if any. If string or Path, it is assumed to be
+        a path to a legacy "flat" JSON file containing the parameters in the style of
+        `hnn_core/param/default.json` (NOT a modern "hierarchical" JSON file like
+        `hnn_core/param/neymotin2020_base.json`). If dict, it is assumed to be a
+        dictionary of parameters in the "flat" JSON style. If None (the default), the
+        default parameters are used from `hnn_core/param/default.json`. Note that if you
+        have drive parameters included, but you ALSO set the deprecated
+        `add_drives_from_params` to True, the drives will be added twice if the drive
+        names are the same.
+    add_drives_from_params : bool, default=False
         If True, add drives as defined in the params-dict. NB this is mainly
         for backward-compatibility with HNN GUI, and will be deprecated in a
         future release. Default: False
-    legacy_mode : bool
+    legacy_mode : bool, default=False
         Set to False by default. Enables matching HNN GUI output when drives
         are added suitably. Will be deprecated in a future release.
     mesh_shape : tuple of int (default: (10, 10))
@@ -219,6 +225,8 @@ def neymotin_2020_model(
     if params is None:
         params_fname = hnn_core_root / "param" / "default.json"
         params = read_params(params_fname)
+    elif isinstance(params, str) or isinstance(params, Path):
+        params = read_params(params)
 
     # Define cell types for Jones 2009 model
     # data is here in metaData format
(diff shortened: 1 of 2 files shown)


hnn_core/network_models.py
```diff
@@ -219,6 +225,8 @@ def neymotin_2020_model(
     if params is None:
         params_fname = hnn_core_root / "param" / "default.json"
         params = read_params(params_fname)
+    elif isinstance(params, str) or isinstance(params, Path):
```
ntolley, line 231: Should there be a typecheck to make sure `params` is one of str, Path, and dict?

ntolley: We don't have an explicit case to check if dict, but if it is dict then nothing needs to be done right?

asoplata: You're right on both accounts. I forgot to add type checking, so I've added it in the new commit. This control flow block is also only used for turning `params` into a proper dictionary, so I've added a comment explaining why we don't need to do anything in that case.

asoplata:
- [x] TODO I forgot to add to the other network models
done in latest commit

hnn_core/network_models.py
asoplata: Marking this as resolved since the type validation has been added to all models at this point.

katduecker (approved):
Looks good to me!

hnn_core/network_models.py
```diff
@@ -171,15 +171,21 @@ def neymotin_2020_model(
...
-    add_drives_from_params : bool
+    params : str | Path | dict | None, default=None
+        Custom Network parameters to use, if any. If string or Path, it is assumed to be
+        a path to a legacy "flat" JSON file containing the parameters in the style of
+        `hnn_core/param/default.json` (NOT a modern "hierarchical" JSON file like
+        `hnn_core/param/neymotin2020_base.json`). If dict, it is assumed to be a
+        dictionary of parameters in the "flat" JSON style. If None (the default), the
+        default parameters are used from `hnn_core/param/default.json`. Note that if you
+        have drive parameters included, but you ALSO set the deprecated
+        `add_drives_from_params` to True, the drives will be added twice if the drive
```
ntolley:
```suggestion
        `add_drives_from_params` to True, the drives will be overwritten if the drive
```
Apologies I missed this, last question to address and we can merge this

If the drives have the same name, I would expect them to be overwritten correct? And if so, the docstring should indicate which values are used in the end

asoplata: Glad you caught this, since while I thought I understood what was going on, there is an error in the docstring; there is no adding-twice or overwriting. Instead, the behavior appears to be like the following: @ntolley what do you think of adding this?

```
    Parameters
    ----------
    params : str | Path | dict | None, default=None
        Custom Network parameters to use, if any. If string or Path, it is assumed to be
        a path to a legacy "flat" JSON file containing the parameters in the style of
        `hnn_core/param/default.json` (NOT a modern "hierarchical" JSON file like
        `hnn_core/param/neymotin2020_base.json`). If dict, it is assumed to be a
        dictionary of parameters in the "flat" JSON style. If None (the default), the
        default parameters are used from `hnn_core/param/default.json`. Note that if you
        want to use any drives defined in the params (either your provided custom params
        or the default), then you must also set `add_drives_from_params` to True. If you
        pass any custom params, then no default params will be used, including for
        drives.
```

PR #1354, opened 2026-09-17
ENH/GUI: Plot drive arrows on dipole plots

### LINKED ISSUE #1130: ENH/GUI: Add ability to plot drive arrows on dipole plots
(This issue, which is still a good one, was copied over from the set of unfinished GUI features at https://docs.google.com/spreadsheets/d/1dveq87P0sLMWqXXbLYFjCHYXjOlcg-ieMphjp-7JtTM/edit?pli=1&gid=0#gid=0 )

This would be a nice feature brought over from the Original HNN, as shown in Figure 4 of the Neymotin 2020 paper here: https://elifesciences.org/articles/51214#fig4

fig4

Souravrajvi0:
## Summary

Closes #1130.

Restores **drive timing markers on dipole plots** in HNN-core, matching the intent of **Original HNN** and [Neymotin et al. 2020, eLife Fig 4](https://elifesciences.org/articles/51214#fig4): colored indicators at external drive onsets so users can relate **input time** to **dipole deflections** on one axis.

## What changed

| Area | Change |
|------|--------|
| `hnn_core.viz` | New `plot_drive_arrows()`; `plot_dipole(..., net=, show_drive_arrows=)` |
| GUI | Analysis control **Overlay drive arrows on dipole** (default on); arrows drawn only for dipole plot types |
| Styling | **Proximal** (red): short arrows **below** the trace, pointing **up**. **Distal** (green): short arrows **above**, pointing **down**. Colors from `default_drive_colors`. **No text labels** by default (`show_labels=False`). **Faint dashed grey verticals** at each drive onset. Defaults **`line_width=2.0`** and **`arrow_height=1.75`** (chunkier, taller markers after label removal—for accessibility). Both kwargs are optional on `plot_drive_arrows` / `plot_dipole`. |
| Tests | `TestDipoleViz::test_plot_drive_arrows` |

**Drive types:** evoked & Gaussian (`mu`), bursty (`tstart`). Poisson/rhythmic omitted (no single onset).

**Not in this PR (follow-ups):** separate timing-strip row like legacy GUI panel D; spike-train drives; GUI controls for arrow width/height; exact pixel parity with deprecated `hnn` GUI.

## How to verify

1. GUI: default ERP simulation → Visualization → **current dipole** with overlay on → proximal/distal markers at default drive times (~27 / ~64 / ~137 ms) with grey guides.
2. Toggle **Overlay drive arrows on dipole** off → arrows and guides disappear.
3. `pytest hnn_core/tests/test_viz.py::TestDipoleViz::test_plot_drive_arrows`
4. Optional: `pytest hnn_core/tests/test_gui.py`

## Screenshots

Latest PNGs at commit `5e903c01` — see [screenshot comment](https://github.com/jonescompneurolab/hnn-core/pull/1354#issuecomment-5764825812).


Souravrajvi0:
### Screenshots

Default ERP aggregate dipole with drive arrows (`neymotin_2020_model`, `add_drives_from_params=True`):

Drive arrows on dipole

Before / after overlay on the same simulation:

With and without drive arrows

ntolley:
Hey @Souravrajvi0 this is really great work! Just chatted with @asoplata and some quick comments:
- Can you place the proximal arrows on the bottom? So that the red arrows are pointing upwards
- I think the text labels on the arrows can be removed

In addition to the arrows, it may be useful to have a faint dashed grey line underneath so that we can see exactly where these drives cross the waveform

Souravrajvi0:
### Screenshots

Default ERP aggregate dipole with drive arrows (`neymotin_2020_model`, `add_drives_from_params=True`):

Drive arrows on dipole

Before / after overlay on the same simulation:

With and without drive arrows

Souravrajvi0:
> Hey @Souravrajvi0 this is really great work! Just chatted with @asoplata and some quick comments:
> - Can you place the proximal arrows on the bottom? So that the red arrows are pointing upwards
> - I think the text labels on the arrows can be removed
> 
> In addition to the arrows, it may be useful to have a faint dashed grey line underneath so that we can see exactly where these drives cross the waveform

Hey @ntolley thanks for the review, all three are in on the latest push if you can check , proximal red arrows below pointing up, distal green above pointing down, no text labels on the overlay as well, and faint dashed grey verticals at each drive time. Updated screenshots are in the PR comment thread is there please check!! Happy to adjust if anything still looks off

Souravrajvi0:
### Updated screenshots (thicker / taller defaults)

After meeting feedback: **`line_width=2.0`**, **`arrow_height=1.75`** 


Drive arrows on dipole

With and without overlay:

With and without drive arrows

@asoplata @ntolley , please confirm these defaults look now that labels are off. Happy to nudge `line_width` or `arrow_height` if needed.

hnn_core/viz.py
```diff
@@ -269,6 +269,208 @@ def plot_laminar_lfp(
     return ax.get_figure()
 
 
+def _drive_arrow_label(drive_name):
+    """Short display label for default ERP drive name patterns."""
+    if "evdist" in drive_name:
+        return "evdist"
+    if "evprox" in drive_name:
+        return "evprox"
+    return drive_name
```
dylansdaniels: We can remove _drive_arrow_label() now since it's not needed

asoplata:
1. Hey Sourav, would you mind deleting the image additions to `.github/`? They are helpful for reference but we do not want to merge them into the final branch.

2. Also, I believe there is an existing ruff format error. If you go to your repo's top-level directory and run `make format-overwrite`, that will apply ruff formatting to all your files. (We cannot see if any tests have failed on CI until ruff passes)

3. When running the tests locally, I find that at least one test is failing, I get:

```
 hnn_core/tests/test_gui.py::test_dipole_data_overlay - AssertionError: assert 8 == 5
```

from an `ax.lines` test at https://github.com/jonescompneurolab/hnn-core/blob/0420385e7ca6940f9342b7c1c39778c83d7b7a9b/hnn_core/tests/test_gui.py#L1139 . As far as I can tell, this is the only line that is failing among all our tests, which is great! I think the solution is that this test needs to be updated since we are probably now drawing more lines than we were before. (If I comment out that one line, the rest of the test passes)

4. Could you please remove `plot_drive_arrows` from the public API? We try to keep our public API as small as possible, and it would be best if the *only* way that a user uses this is in the context of a pre-existing dipole plot, aka as an argument to `plot_dipole`, which you have already done. Would you mind A) removing it from the public API in `doc/api.rst` and **B) renaming it to `_add_arrows_to_dipole`.** We want to avoid prefixing the function with `plot_XXX` since we want functions of that name to be the primary entry point for creating plots.

hnn_core/viz.py
```diff
@@ -269,6 +269,208 @@ def plot_laminar_lfp(
...
+        event_time = marker["time"]
+        label = marker["label"] if show_labels else ""
+        color = marker["color"]
+
+        if marker["location"] == "proximal":
+            offset_idx = time_offsets_bottom.get(event_time, 0)
+            time_offsets_bottom[event_time] = offset_idx + 1
+            time_plot = event_time + offset_idx * 1.5
+            arrow_tip_y = ymin_data + tip_inset
+            label_y = ymin_plot + anchor_inset + offset_idx * stack_step
```
asoplata: Hey Sourav, as @dylansdaniels and I reviewed the code, we noticed that the length of the proximal vs distal arrows are slightly different. However, we had trouble following where this difference was coming from throughout this entire function (not just the highlighted region. Would you mind refactoring the overall logic to be a bit simpler with more self-explanatory variable names (we try to write our code such that names can be understood by a first- or second-year graduate student who may be relatively new to programming)? And would you mind changing the code so that the arrows are the same length?

A related thing is that it appears that `offset_idx` may be intended for dealing with cases where multiple drives start at the same time. However, we are not concerned with this case: if multiple arrows need to be created that overlap at the same point in time, that's okay, and we don't need to show multiple arrows (or any offset arrows). I think not accounting for this will also help make the code much simpler.

hnn_core/viz.py
Souravrajvi0: This is noted @dylansdaniels

Souravrajvi0:
> 1. Hey Sourav, would you mind deleting the image additions to `.github/`? They are helpful for reference but we do not want to merge them into the final branch.
> 2. Also, I believe there is an existing ruff format error. If you go to your repo's top-level directory and run `make format-overwrite`, that will apply ruff formatting to all your files. (We cannot see if any tests have failed on CI until ruff passes)
> 3. When running the tests locally, I find that at least one test is failing, I get:
> 
> ```
>  hnn_core/tests/test_gui.py::test_dipole_data_overlay - AssertionError: assert 8 == 5
> ```
> 
> from an `ax.lines` test at
> 
> https://github.com/jonescompneurolab/hnn-core/blob/0420385e7ca6940f9342b7c1c39778c83d7b7a9b/hnn_core/tests/test_gui.py#L1139
> 
> . As far as I can tell, this is the only line that is failing among all our tests, which is great! I think the solution is that this test needs to be updated since we are probably now drawing more lines than we were before. (If I comment out that one line, the rest of the test passes)
> 4. Could you please remove `plot_drive_arrows` from the public API? We try to keep our public API as small as possible, and it would be best if the _only_ way that a user uses this is in the context of a pre-existing dipole plot, aka as an argument to `plot_dipole`, which you have already done. Would you mind A) removing it from the public API in `doc/api.rst` and **B) renaming it to `_add_arrows_to_dipole`.** We want to avoid prefixing the function with `plot_XXX` since we want functions of that name to be the primary entry point for creating plots.

Hey @asoplata  , thanks again for the detailed review. I went through the comments and pushed the updates.
A few things I changed:

1. Removed plot_drive_arrows from the public API and renamed the helper to _add_arrows_to_dipole. The user-facing path is still plot_dipole(..., show_drive_arrows=True).
2. Removed the three drive-arrow PNGs from .github/images/ and ran formatting/ruff on viz.py.
3. Updated test_dipole_data_overlay to expect 8 lines, accounting for the 5 dipole/data traces plus the 3 drive-onset guide lines. So the test won't fail anymore 
4. Updated the viz tests for the helper rename and line_width changes.
5. Simplified _add_arrows_to_dipole by removing the offset_idx stacking and horizontal offsets. Now we only draw one marker for the same (drive_time, location).
6. Changed the proximal and distal arrows to use the same arrow_length_y so they have the same length.
7. Renamed a few variables to make the arrow-positioning logic easier to follow.
8. Increased the arrow length to 0.127 * y_span * arrow_height after the refactor made them look shorter than before, so they are closer to the previous proximal arrow size.
9. Increased the default line_width from 2.0 to 3.0 so the arrow shaft and head are more visible.
10. Also removed the unused _drive_arrow_label() from Dylan's comments. Labels are off by default, and when enabled they use the full drive names.

Souravrajvi0:
image

PR #1356, opened 2026-09-17
[WIP] Bug fix in plot_spikes_raster

katduecker:
This small PR fixes a bug that I may have introduced when refactoring the plot_spikes_raster function. There is an error when one of the cell types didn't spike. To work around that, I now track the maximum gid in the event plot for loop and use that to scale the raster.


asoplata:
LGTM, thanks for fixing it Katharina. Can you please add a small test to make sure the raster is scaled correctly?

katduecker:
Done! I also added the option to have gid_ranges as an input like we do for other functions, and I now force to ylim to be maximum gid + marker_size instead of letting matplotlib figure it out itself..

I think we should really refactor cell_response to have access to info in network though.. at some point...

asoplata:
When reviewing this, I discovered another more serious bug: **`plot_spikes_raster`'s use of the `cell_types` argument is now broken**. This appears to be related to the mega PR we did for the Duecker model, but strangely our testing did not catch it.

For example, if I take `master` as of  https://github.com/jonescompneurolab/hnn-core/commit/a5014172419379f3cbe96655cda8714bf202f767 (the commit right before the Duecker model merge), then run a simple network with a single drive, then plot it using `net.cell_response.plot_spikes_raster(cell_types=["L2_pyramidal", "alpha_prox"], show=True)`, I get the below, which is correct and what I expect to get.

a50141724

However, when I checkout https://github.com/jonescompneurolab/hnn-core/commit/49f210fb41481ab859537e9b3e32a74117ae95fc and repeat, I get the following, which is incorrect. This below plot seems to show all "cell" spikes (no drives), even if I pass legitimate options to `cell_types`. The below also appears if I don't try to plot the drives, and instead only use `net.cell_response.plot_spikes_raster(cell_types=["L2_pyramidal"], show=True)`

49f210fb4

In other words, there is a regression currently on `master` with respect to the `cell_types` argument of this function. **This bug should also be fixed in this PR, since it's directly related to this code**.

We should also take this regression as a warning: *Just because the tests are passing doesn't mean that our plotting has new bugs related to the recent mega PR*. For example, our tests related to this argument in https://github.com/jonescompneurolab/hnn-core/blob/00e9cbcbf616ed77b0cb885b766fd9a212fbc5ac/hnn_core/tests/test_viz.py#L682-L691 didn't catch this. Hopefully, this is the only regression, but we need to be cautious. I will try to hand check all the plotting arguments that were recently changed.

-------

Unrelated: I discovered the above while I was trying to review the use of the `cell_types` argument. Separately, when trying to review this PR, I think the documentation and variable names need some refactoring for `plot_spikes_raster` and, to a much smaller extent, `plot_spikes_hist`. (These are pre-existing issues prior to this PR #1356 that Katharina opened)

> I think we should really refactor cell_response to have access to info in network though.. at some point...

You're tellin' me. I was reviewing this and VERY confused when comparing `plot_spikes_raster` to `plot_spikes_hist` and trying to understand which functions work on drives vs don't work on drives, until I discovered:

```python
(Pdb++) net.cell_response.cell_types # <--- includes drives, equivalent to the unique names of cell_response.spike_types
['L2_basket', 'L2_pyramidal', 'L5_basket', 'L5_pyramidal', 'beta_dist', 'beta_prox']
(Pdb++) net.cell_response._cell_type_names  # <--- does NOT include drives
['L2_basket', 'L2_pyramidal', 'L5_basket', 'L5_pyramidal']
(Pdb++) net.cell_types.keys() # <-- does NOT include drives
dict_keys(['L2_basket', 'L2_pyramidal', 'L5_basket', 'L5_pyramidal'])
```

Both `plot_spikes_raster` and `plot_spikes_hist` appear to functionally support usage of BOTH proper "cells" and drives (before the regression mentioned above). However, `plot_spikes_hist` only discusses "spike types" without defining what those are, and all the examples are only examples of drives, weakly implying that that function only deals with drives or "drive cells". Conversely, `plot_spikes_raster` sometimes refers only to cells like with its `cell_types` argument, but other times (such as Katharina's new argument docstring and in a comment at https://github.com/jonescompneurolab/hnn-core/blob/00e9cbcbf616ed77b0cb885b766fd9a212fbc5ac/hnn_core/viz.py#L741 ) suggests that it does accommodate drives. However, almost all of its code deals with `cell types`, which for most users / new developers they will think it's referring to cell types as in `net.cell_types.keys()`. This is confusing, because this function handles BOTH "cell types" in the Network sense and drives as well.

@ntolley @dylansdaniels Therefore, I propose we do the following:

- In the documentation of `plot_spikes_hist`, define "spike types" to be that of both "cell types" (such as `Network.cell_types.keys()`) and drive types. Also provide a cell type name in one of the examples, in combination with the existing drive names used.
- Rename `CellResponse.cell_types` to be `CellResponse.spike_type_names`, which is more accurate to how it's defined
- In `plot_spikes_raster`, add a new argument `spike_types` identical to that of `plot_spikes_hist`, so that 1. we define what it means, 2. the API between the two function is similar, and 3. we discriminate spike types from their partitioned subsets of colloquial "cell types" and "drives"
- In `plot_spikes_raster`, we refactor the variable names accordingly, so that it's clear when we're talking spike types vs only cell types vs only drive drives.
- In `plot_spikes_raster`, we add a small `FutureWarning` that `cell_types` argument is deprecated in favor or `spike_types`

I will try to help with this as soon as I can, but I may not get to it before Wednesday.

katduecker:
@asoplata I fixed the bug

PR #1358, opened 2026-09-17
[MRG] Ref/improve duecker model args

asoplata:
This does two things to the new `duecker_ET_model` function

1. Removes the argument `legacy_mode`, since it is unnecessary to be used with the Duecker model
2. Removes the argument `add_drives_from_params` (which is currently broken and causes an error), and replaces it with `add_default_drives`. The new `add_default_drives` boolean argument provides users with an easy argument to add the three drives currently used in our testing script `script_duecker_simulate_save.py`, which itself uses drives taken from     https://github.com/jonescompneurolab/hnn-tuning/blob/duecker_ET_model/new_model_HNN/ERP/handtune_new_syn.ipynb  as of 2026-07-15. 

Note that just like the model itself, the drive parameters don't have to be "written in stone" here. They can be subject to change in the future until the time of version release or preprint publication. However, this approach (similar to the one in #1340) of providing an easy argument to add the appropriate drives is not only convenient for users, but also extremely convenient for testing.


asoplata:
@katduecker Kind of related, in the above notebook there's a line 

```python
net.set_cell_positions(inplane_distance=30.)
```

Should this be the default used for `duecker_ET_model`? Currently I think it's set to `1`, not `30`: https://github.com/jonescompneurolab/hnn-core/blob/d470259a05256f5f571eef2eed09a2d4050b2f8b/hnn_core/network_models.py#L698

asoplata:
Can we also rename these like:

- `prox1 -> evprox1`
- `prox2 -> evprox2`
- `dist1 -> evdist1`

?

katduecker:
@asoplata I have a preference for keeping inplane distance at 1 by default because it makes the plotting of space constant easier, and it's a good baseline to work from when thinking about adjusting the distances.

I remember Steph saying that she doesn't think the drives should be named evprox and evdist, but prox and dist. This was in a meeting a while and I don't think anyone took a note of it. Since in the papers we always refer to them as proximal and distal, I have a preference for not naming them evprox and evdist.

asoplata:
@katduecker Okay, other than the inplane distance (which I didn't change from `master`) and the drive names (which I also didn't change from `master`, they still keep your preferred names), are you happy with this PR otherwise?

